PR #28, opened 2020-05-22
Implement DQN

### LINKED ISSUE #6: Implement Vanilla DQN
Almost ready, performance tests are on-going....

- doc example need to be updated afterward

@Artemis-Skade is working on it

### LINKED ISSUE #37: Memory allocation for buffers
With the current implementation of buffers.py one can request a buffersize which doesn't fit in the memory provided but because of numpys implementation of ``np.zeros()`` the memory is not allocated before it is actually used. But because the buffer is meant to be filled completely (otherwise one could just use a smaller buffer) the computer will finally run out of memory and start to swap heavily. Because there are only smaller parts of the buffer that are accessed at once (minibatches) the system will just swap the necessary pages in and out of memory. At that moment the progress of the run is most likely lost and one has to start a new run with a smaller buffer.

I would recommend using ``np.ones`` instead, as it will allocate the buffer at the beginning and fail if there is not enough memory provided by the system. The only issue is that there is no clear error description in the case where the system memory is exceeded but python gets simply killed by the OS with a SIGKILL. Maybe one could catch that command?

### LINKED ISSUE #46: Potential continuity bug in the replay buffer when calling .learn multiple times
```python
agent.learn(N)
agent.learn(N)
# ER isn't reset but the environment is.
```
This isn't a bug - as is - but the exact behaviour needs to be defined in order to decide what is the correct way to deal with a bug in ER in the DQN branch:

_Originally posted by @PartiallyTyped in https://github.com/DLR-RM/stable-baselines3/pull/28#issuecomment-640253275_

Artemis-Skade:
## Description

Implementation of vanilla dqn

closes #6 
closes #37 
closes #46 

Missing:

- [x] Update examples to include DQN
- [x] Add test for replay buffer truncation

## Motivation and Context

## Types of changes

## Checklist:


stable_baselines3/dqn/dqn.py
```diff
@@ -0,0 +1,210 @@
...
+
+            # Optimize the policy
+            self.policy.optimizer.zero_grad()
+            loss.backward()
+            # Clip grad norm
+            th.nn.utils.clip_grad_norm_(self.policy.parameters(), self.max_grad_norm)
+            self.policy.optimizer.step()
+
+            # Update target networks
+            if self._n_updates % self.target_update_interval == 0:
```
araffin, line 150: you should update `self._n_updates` before, otherwise this condition may never be true

stable_baselines3/dqn/dqn.py
```diff
@@ -0,0 +1,210 @@
...
+from stable_baselines3.common.type_aliases import GymEnv, MaybeCallback
+from stable_baselines3.common.utils import get_linear_fn
+from stable_baselines3.dqn.policies import DQNPolicy
+
+
+class DQN(OffPolicyRLModel):
+    """
+    Deep Q-Network (DQN)
+
+    Paper: https://arxiv.org/abs/1312.5602
```
araffin, line 17: Please add the nature paper here too

stable_baselines3/dqn/dqn.py
```diff
@@ -0,0 +1,210 @@
...
+        used for evaluating the agent periodically. (Only available when passing string for the environment)
+    :param policy_kwargs: (dict) additional arguments to be passed to the policy on creation
+    :param verbose: (int) the verbosity level: 0 no output, 1 info, 2 debug
+    :param seed: (int) Seed for the pseudo random generators
+    :param device: (str or th.device) Device (cpu, cuda, ...) on which the code should be run.
+        Setting it to auto, the code will be run on the GPU if possible.
+    :param _init_setup_model: (bool) Whether or not to build the network at the creation of the instance
+    """
+
+    def __init__(self, policy: Union[str, Type[DQNPolicy]],
```
araffin, line 46: Please use the defaults from the nature paper (see Table "List of hyperparameters and their values")

stable_baselines3/dqn/dqn.py
Miffyli: I would refer specifically to nature paper, to avoid any confusion with parameters. I think they differ a bit between the two papers.

araffin: The problem with nature paper is that it is behind a paywall :/

Miffyli: Ah, right :/. I see the problem, but I think we should refer to both papers, and say that we use parameters (and default architectures) from the nature version, to avoid any confusion.

stable_baselines3/dqn/dqn.py
Artemis-Skade: I changed most of them however the nature paper suggests that there should be a train_freq smaller than 1, e.g. 1/4 so only one update every 4 actions, that is currently not possible should I change that?

stable_baselines3/dqn/dqn.py
Artemis-Skade: I was unable to extract the architecture from the paper but I guess it is already implemented in the CnnPolicy? @araffin

stable_baselines3/dqn/dqn.py
Miffyli: I think it should be included. `train_freq=1` makes things slower and perhaps overuses the recent samples for training (in any case, lets stay true to the original paper).

stable_baselines3/dqn/dqn.py
Miffyli: Yes, [NatureCNN](https://github.com/DLR-RM/stable-baselines3/blob/master/stable_baselines3/common/policies.py#L53) is the correct architecture.

stable_baselines3/dqn/dqn.py
```diff
@@ -0,0 +1,213 @@
...
+                 learning_rate: Union[float, Callable] = 25e-5,
+                 buffer_size: int = 1000000,
+                 learning_starts: int = 50000,
+                 batch_size: Optional[int] = 32,
+                 gamma: float = 0.99,
+                 train_freq: int = 4,
+                 gradient_steps: int = 1,
+                 tau: float = 1.0,
+                 target_update_interval: int = 10000,
+                 exploration_fraction: float = 1,
```
araffin, line 58:
```suggestion
                 exploration_fraction: float = 0.1,
```

docs/modules/dqn.rst
```diff
@@ -0,0 +1,96 @@
...
+-  Gym spaces:
+
+
+============= ====== ===========
+Space         Action Observation
+============= ====== ===========
+Discrete      ✔      ✔
+Box           ❌      ✔
+MultiDiscrete ❌      ❌
+MultiBinary   ❌      ❌
```
araffin, line 44:
```suggestion
MultiDiscrete ❌      ✔
MultiBinary   ❌      ✔
```

docs/modules/dqn.rst
```diff
@@ -0,0 +1,96 @@
...
+Example
+-------
+
+.. code-block:: python
+
+  import gym
+  import numpy as np
+
+  from stable_baselines3 import DQN
+  from stable_baselines3.sac import MlpPolicy
```
araffin, line 57:
```suggestion
  from stable_baselines3.dqn import MlpPolicy
```

docs/modules/dqn.rst
```diff
@@ -0,0 +1,96 @@
...
+  model.learn(total_timesteps=10000, log_interval=4)
+  model.save("dqn_pendulum")
+
+  del model # remove to demonstrate saving and loading
+
+  model = DQN.load("dqn_pendulum")
+
+  obs = env.reset()
+  while True:
+      action, _states = model.predict(obs)
```
araffin, line 71:
```suggestion
      action, _states = model.predict(obs, deterministic=True)
```

docs/misc/changelog.rst
```diff
@@ -15,7 +15,8 @@ New Features:
 - Added env checker (Sync with Stable Baselines)
 - Added ``VecCheckNan`` and ``VecVideoRecorder`` (Sync with Stable Baselines)
 - Added determinism tests
-- Added ``cmd_util`` and ``atari_wrappers``
+- Added ``cmd_utils`` and ``atari_wrappers``
+- Added ``DQN``
```
araffin, line 19: missing: replay buffer dtype + excluded params

stable_baselines3/common/utils.py
```diff
@@ -74,6 +74,25 @@ def get_schedule_fn(value_schedule: Union[Callable, float]) -> Callable:
     return value_schedule
 
 
+def get_linear_fn(start: float, end: float, end_fraction: float):
```
araffin, line 77: missing return type hint

stable_baselines3/dqn/dqn.py
```diff
@@ -0,0 +1,213 @@
...
+    :param verbose: (int) the verbosity level: 0 no output, 1 info, 2 debug
+    :param seed: (int) Seed for the pseudo random generators
+    :param device: (str or th.device) Device (cpu, cuda, ...) on which the code should be run.
+        Setting it to auto, the code will be run on the GPU if possible.
+    :param _init_setup_model: (bool) Whether or not to build the network at the creation of the instance
+    """
+
+    def __init__(self, policy: Union[str, Type[DQNPolicy]],
+                 env: Union[GymEnv, str],
+                 learning_rate: Union[float, Callable] = 25e-5,
```
araffin, line 49:
```suggestion
                 learning_rate: Union[float, Callable] = 2.5e-4,
```

If I remember, the default optimizer of DQN was not Adam (but it did not exist at that time)
so we will see with the performance tests

PartiallyTyped: It was RMSProp.

stable_baselines3/dqn/dqn.py
Artemis-Skade: this is because I used the hyperparameters from the nature paper https://www.nature.com/articles/nature14236/tables/1 see final exploration frame

stable_baselines3/dqn/dqn.py
araffin: yep, this should do the trick:
```yaml
policy_kwargs: "dict(optimizer_class=th.optim.RMSprop,
                                  optimizer_kwargs=dict(alpha=0.95, momentum=0.95, eps=0.01))"
```
although I think we need to do `target_update_interval = target_update_interval // train_freq` to be correct (with `min(target_update_interval) = 1`)

stable_baselines3/dqn/dqn.py
Artemis-Skade: okay, they had 50.000.000 frames total, so I guess a fraction of 0.02 would be appropriate?

araffin: usually, Atari is on 10M steps (40M frames because of frame-skip)
this is a fraction of the total training steps, so should be lower than 1 anyway.
I guess we will have to tune the hyperparameters ourself to make it work in 10M steps...

Unfortunately, rainbow (that compares extensions to DQN) does not reference all of them...
https://arxiv.org/abs/1710.02298

stable_baselines3/dqn/dqn.py
```diff
@@ -0,0 +1,214 @@
...
+                 gradient_steps: int = 1,
+                 tau: float = 1.0,
+                 target_update_interval: int = 10000,
+                 exploration_fraction: float = 0.02,
+                 exploration_initial_eps: float = 1.0,
+                 exploration_final_eps: float = 0.1,
+                 max_grad_norm: float = 10,
+                 tensorboard_log: Optional[str] = None,
+                 create_eval_env: bool = False,
+                 policy_kwargs: Optional[Dict[str, Any]] = {'optimizer_class': th.optim.RMSprop,
```
araffin, line 64: you should not use mutable as default arguments... please look at how it is done for A2C

araffin:
For reference, we should match the results in https://github.com/NervanaSystems/coach/tree/master/benchmarks/dqn
using the hyperparameters from https://github.com/NervanaSystems/coach/blob/master/rl_coach/agents/dqn_agent.py

I see 3 differences:
1. they are using Adam as optimizer
2. they are using Huber loss instead of MSE
3. they are using a different epsilon (different from the one used for training) for evaluation `exploration.evaluation_epsilon = 0.05`

I don't see where is the learning rate defined though :/

araffin:
EDIT: they mentioned "The parameters used for DQN are the same parameters as described in the original paper, except for the optimizer (changed to ADAM) and learning rate (1e-4) used."

Artemis-Skade:
> For reference, we should match the results in https://github.com/NervanaSystems/coach/tree/master/benchmarks/dqn
> using the hyperparameters from https://github.com/NervanaSystems/coach/blob/master/rl_coach/agents/dqn_agent.py
> 
> I see 3 differences:
> 
>     1. they are using Adam as optimizer
> 
>     2. they are using Huber loss instead of MSE
> 
>     3. they are using a different epsilon (different from the one used for training) for evaluation `exploration.evaluation_epsilon = 0.05`
> 
> 
> I don't see where is the learning rate defined though :/

I am confused now, which parameters should I use?

araffin:
>I am confused now, which parameters should I use?

For now, keep them "as-is", maybe just update the loss using "Huber" (aka "SmoothL1Loss") loss (https://pytorch.org/docs/stable/nn.html#torch.nn.SmoothL1Loss) instead of `mse_loss`

Miffyli:
Hmm I would default to original-implementation optimizers, especially if it being referenced to. Why are you not checking performance against stable-baselines results, btw?

stable_baselines3/common/policies.py
```diff
@@ -53,7 +53,7 @@ def forward(self, observations: th.Tensor) -> th.Tensor:
 class NatureCNN(BaseFeaturesExtractor):
     """
     CNN from DQN nature paper: https://arxiv.org/abs/1312.5602
```
Miffyli, line 55: I am removing this reference in #35, to avoid any confusion as to what is the network architecture.

stable_baselines3/dqn/policies.py
```diff
@@ -0,0 +1,267 @@
...
+                 device: Union[th.device, str] = 'auto',
+                 activation_fn: Type[nn.Module] = nn.ReLU,
+                 epsilon: float = 0.05,
+                 normalize_images: bool = True):
+        super(QNetwork, self).__init__(observation_space, action_space,
+                                       features_extractor=features_extractor,
+                                       normalize_images=normalize_images,
+                                       device=device)
+
+        if net_arch is None:
```
Miffyli, line 39: This defaulting would cause headache (e.g. you give CNN policy, and now you suddenly have network with layers NatureCNN - 64 - 64 - output, while users expect NatureCNN - output). This seems to be a common trend around rest of the code that I think should be fixed.

stable_baselines3/dqn/policies.py
```diff
@@ -0,0 +1,267 @@
...
+
+import gym
+import numpy as np
+import torch as th
+import torch.nn as nn
+from stable_baselines3.common.policies import (BasePolicy, register_policy, create_mlp,
+                                               NatureCNN, BaseFeaturesExtractor, FlattenExtractor)
+
+
+class QNetwork(BasePolicy):
```
Miffyli, line 11: Couldn't this processing be done in the Policy code? The "network" part of Q-learning is just the feature_extractor plus linear mapping to outputs size, so I do not think we need separate class for this, since it is not required.

stable_baselines3/dqn/policies.py
```diff
@@ -0,0 +1,267 @@
...
+        excluding the learning rate, to pass to the optimizer
+    """
+
+    def __init__(self, observation_space: gym.spaces.Space,
+                 action_space: gym.spaces.Space,
+                 lr_schedule: Callable,
+                 net_arch: Optional[List[int]] = None,
+                 device: Union[th.device, str] = 'auto',
+                 activation_fn: Type[nn.Module] = nn.ReLU,
+                 epsilon: float = 0.05,
```
Miffyli, line 121: I think `epsilon` should be on algorithm level, not in the policy (along with the scheduling stuff). It is not required to use the policy.

stable_baselines3/dqn/policies.py
araffin: it won't be the case normally (see line 135), what do you propose instead?

araffin:
>you not checking performance against stable-baselines results, btw

Good point, I did not do that because SB DQN is a bit special, I need to deactivate dueling and double dqn. Will try to do the benchmark next week.
I will also use the same pre-processing (as we switched to gym preprocessing in SB3).

stable_baselines3/dqn/policies.py
araffin: true, this was mostly do avoid code duplication when creating the target. It also mimics what is done in SAC/TD3.

stable_baselines3/dqn/policies.py
Miffyli: Oh man, a proper brainfart. For some reason I have missed this in all of the code I have been going over. It looks good as it is, my bad!

araffin:
Result of early benchmark: Pong can be solved in 2M steps but I only managed to reach a mean episodic reward of 10 on Breakout :/
I will look for bug/differences next week

araffin:
I've good and bad news... the good news is that SB2 Vanilla DQN seems to work with paper hyperparameters (but with Adam optimizer and update lr) and that SB3 atari preprocessing works:

```
OrderedDict([('buffer_size', 1000000),
             ('double_q', False),
             ('env_wrapper',
              ['stable_baselines3.common.atari_wrappers.AtariWrapper']),
             ('exploration_final_eps', 0.05),
             ('frame_stack', 4),
             ('learning_rate', 0.0001),
             ('learning_starts', 50000),
             ('n_timesteps', 10000000.0),
             ('policy', 'CnnPolicy'),
             ('policy_kwargs', 'dict(dueling=False)'),
             ('prioritized_replay', False),
             ('target_network_update_freq', 10000),
             ('train_freq', 4)])
```

training reward of SB2 for reference:
```
--------------------------------------
| % time spent exploring  | 41       |
| episodes                | 2500     |
| mean 100 episode reward | 5.9      |
| steps                   | 615557   |
--------------------------------------
--------------------------------------
| % time spent exploring  | 37       |
| episodes                | 2600     |
| mean 100 episode reward | 7.6      |
| steps                   | 656971   |
--------------------------------------
--------------------------------------
| % time spent exploring  | 32       |
| episodes                | 2700     |
| mean 100 episode reward | 10.1     |
| steps                   | 707281   |
--------------------------------------
--------------------------------------
| % time spent exploring  | 27       |
| episodes                | 2800     |
| mean 100 episode reward | 12.3     |
| steps                   | 764229   |
--------------------------------------
--------------------------------------
| % time spent exploring  | 21       |
| episodes                | 2900     |
| mean 100 episode reward | 16       |
| steps                   | 831493   |
--------------------------------------
--------------------------------------
| % time spent exploring  | 13       |
| episodes                | 3000     |
| mean 100 episode reward | 20.5     |
| steps                   | 913075   |
--------------------------------------

```

The bad news is that mean that the current SB3 implementation is buggy, I will try to track down the issue. (Could be CNN initialization, gradient clipping, ... I assume it is subtile)

araffin:
I think I know what is the problem... forgot frame-stacking in the SB3 hyperparams (I'm bad at copy pasting ^^")

araffin:
Ok, looks much better =)
```
| rollout/            |          |
|    ep_len_mean      | 3.48e+03 |
|    ep_rew_mean      | 26       |
|    exploration rate | 0.05     |
| time/               |          |
|    episodes         | 3076     |
|    fps              | 442      |
|    time_elapsed     | 2400     |
|    total timesteps  | 1062959  |
| train/              |          |
|    learning_rate    | 0.0001   |
|    n_updates        | 253239   |
```
other good news, the SB3 buffer implementation seems significantly faster than SB2 (no more slow calls to  `.append()`).

stable_baselines3/common/base_class.py
```diff
@@ -681,6 +681,12 @@ class OffPolicyRLModel(BaseRLModel):
     :param buffer_size: (int) size of the replay buffer
     :param learning_starts: (int) how many steps of the model to collect transitions for before learning starts
     :param batch_size: (int) Minibatch size for each gradient update
+    :param tau: (float) the soft update coefficient ("Polyak update", between 0 and 1)
```
Miffyli, line 684: Could clarify here that "tau=1.0" means hard updates (no soft updating)

araffin: yep, thought the same... but we may change the hyperparameters as a buffer of 1M frames x 4 don't fit on many pc
will see (currently using a buffer size of 5e5 for the tests)

araffin:
>What if the episode terminates here?

Yes, I guess that was @PartiallyTyped concern too...

araffin:
I think the solution is to not write `next_observation` if an episode terminates

PartiallyTyped:
This only works if the episode terminates because the mdp terminates and not because the agent run out of steps.

araffin:
>This only works if the episode terminates because the mdp terminates and not because the agent run out of steps.

We don't make any distinctions in SB3, we have the `TimeFeatureWrapper` in the zoo for that.

Miffyli:
> We don't make any distinctions in SB3, we have the TimeFeatureWrapper in the zoo for that.

This was not supported by SB3 to any degree yet, so I do not think we should bother with it just yet (v3.1 stuff). 

Not storing next observation has its limitations though, and could cause confusion later on if some other algorithm needs to know the observation in the terminal state.

Yh what a can of worms 😅

PartiallyTyped:
Okay nevermind my previous reply. My brain has been fried. I was thinking that termination is not necessarily done signal due to 'end of MDP'.

Regardless.

Here is my suggestion. I think it checks out.  
Keep an `n` sized buffer and store all the observations. At any point in time, there is a limit as to which elements we can sample, the limit can be found through `pos`. Before the buffer is filled, `pos` gives us the size, this is a hard limit on what we can sample. After the buffer is filled, `pos` gives us a logical limit. It tells us exactly transition we can't sample, i.e. the point where the trajectories overlap.

The idea isn't exactly trivial but it seems sound:

- Write `obs` at `pos`, next_obs at `(pos+1) % buffer_size`.
- `pos` takes values in [0, n-1], when pos is at n-1, it writes the nth obs at index n-1, and the nth next obs at index `n%buffer_size`.
- Sample:
  - if the buffer isn't full, sample up to `pos`, obs is the sampled index, next_obs is `(current_obs_index+1)%buffer_size`
  - if the buffer is full, then we can sample {0..n}-{pos}, except the `pos` position, because the pos position blends the trajectories. To implement this, we sample `([1, n-1] + pos) % buffer size` for current obs, and +1 for next_obs.

The whole point of `[1, n-1] + pos` is that we account for the position of pos after the buffer is full. If pos is 0, then we can sample everything but `pos`, if it is at 1, then we sample {2..`buffer_size-1`} + {0}.

PartiallyTyped:
So it works and the solution is just this line inside of sample.

```diff
-         batch_inds = np.random.randint(0, upper_bound, size=batch_size)
+         batch_inds = (np.random.randint(1, upper_bound, size=batch_size) + self.pos)%self.buffer_size
```

stable_baselines3/common/buffers.py
```diff
@@ -188,7 +210,8 @@ def _get_samples(self,
                      ) -> ReplayBufferSamples:
         data = (self._normalize_obs(self.observations[batch_inds, 0, :], env),
                 self.actions[batch_inds, 0, :],
```
PartiallyTyped, line 212:
```suggestion
                     ) -> ReplayBufferSamples:
        batch_inds = (batch_inds + pos) % self.buffer_size
        data = (self._normalize_obs(self.observations[batch_inds, 0, :], env),
                self.actions[batch_inds, 0, :],
```

This corrects the sample alignment issue we discussed when using a single array of observations.

PartiallyTyped: Alternatively, override the sample function with this:

```python
    def sample(self,
               batch_size: int,
               env: Optional[VecNormalize] = None
               ):
        """
        :param batch_size: (int) Number of element to sample
        :param env: (Optional[VecNormalize]) associated gym VecEnv
            to normalize the observations/rewards when sampling
        :return: (Union[RolloutBufferSamples, ReplayBufferSamples])
        """
        upper_bound = self.buffer_size if self.full else self.pos
        batch_inds = (np.random.randint(1, upper_bound, size=batch_size)+self.pos) % self.buffer_size
        return self._get_samples(batch_inds, env=env)
```

araffin: ok, I will see, need to think at it with a clear head (so tomorrow ^^#)

PartiallyTyped: Of course. I will add some tests that verify this specific issue.

araffin: the first solution does not seem to work but the second does ;)

PartiallyTyped: Woups, `self.pos` haha. I think the latter is cleaner ^^'

PartiallyTyped: The only caveat here is that the implementation expects that the trajectories are continuous which, I suppose, is fine for now.

Miffyli: I am rather confused by this function. If we only have samples till `self.pos` index, and take random integers from [`self.pos + 1`, `self.pos + 1 + self.pos`], won't some of these be above `self.pos` and thus lead to invalid observations?

araffin: and it should be 
```python
batch_inds = (np.random.randint(1, upper_bound, size=batch_size) + self.pos) % upper_bound
```
otherwise you are going to sample a lot of zeros...

araffin: actually, still not correct

Miffyli: I think this is still same as just doing `batch_inds = np.random.randint(1, upper_bound, size=batch_size)`. I am not even sure what we are trying to fix here any more.

PartiallyTyped: @Miffyli , yes that only works when the buffer is full.

This works for both cases:

```python
    def sample(self,
               batch_size: int,
               env: Optional[VecNormalize] = None
               ):
        """
        :param batch_size: (int) Number of element to sample
        :param env: (Optional[VecNormalize]) associated gym VecEnv
            to normalize the observations/rewards when sampling
        :return: (Union[RolloutBufferSamples, ReplayBufferSamples])
        """
        
        upper_bound = self.buffer_size if self.full else self.pos
        
        if self.full:
             batch_inds = (np.random.randint(1, self.buffer_size, size=batch_size)+self.pos) % self.buffer_size
        else:
              batch_inds = (np.random.randint(1, self.pos, size=batch_size)+self.pos) % self.buffer_size
        return self._get_samples(batch_inds, env=env)
```

The first part samples from {0.. buffer_size-1} - {pos}, the second is for the case where the buffer isn't full, so sampling {0..pos-1} is fine.

PartiallyTyped: The whole point is to avoid sampling `self.pos` because once the buffer is full, then sampling from the index `self.pos` is wrong, because that will use two unrelated observations.

PartiallyTyped: This notebook illustrates the problems with current implementation. The source is copied and pasted from the repo:

https://colab.research.google.com/drive/1J_9TVXdPTPemFv8sPxzYwgqb6tGfyhCx?usp=sharing

PartiallyTyped: First, with an array of size `buff_size`, we have access to buff_size - 1 observations.
Second, sampling `self.pos` should never be possible, but
`batch_inds = np.random.randint(0, upper_bound, size=batch_size)` from the base class doesn't account for `self.pos` when the buffer is full, therefore, at some point, it will sample `self.pos` if the buffer is full.

araffin: >This notebook illustrates the problems with current implementation. 

I don't have access to it... but implemented your last solution locally (tests are running)

PartiallyTyped: Ah sorry. This should do it:

https://colab.research.google.com/drive/1J_9TVXdPTPemFv8sPxzYwgqb6tGfyhCx?usp=sharing

PartiallyTyped:
I think there is a continuity bug with the ER buffer.

```python
agent.learn(N)
agent.learn(N)
# ER isn't reset but the environment is.
```
The consequence is that we change the transition at `buffer.pos-1`...
When we call `.learn(..., reset_num_timesteps=True)`, we set `self._last_obs = env.reset()`. Thus when we store the transition, we store it at `buffer.pos`, which breaks transition at index `buffer.pos-1`.

araffin:
>The consequence is that we change the transition at

True, but the bug is already present, no? (and maybe in SB2 too)
Please open an issue so we don't lose track of that one ;)

PartiallyTyped:
I think the only reason the bug exists is due to the single array for the observations, so unless SB2  moved on to a single array for it, it shouldn't be an issue.

araffin:
I've done additional performance tests and could train DQN on CartPole, LunarLander, MountainCar and Acrobot (with some hyperparameter tuning for each).

Regarding Atari games, I could finally get results that look like the ones from RL Coach on Breakout, and surprise surprise.... the issue may come from gym pre-processing (we really cannot rely on OpenAI code... @Miffyli  you should read that for your tests). I will push SB2 pre-processing instead soon.

DQN is a bit unstable on Breakout (catastrophic drop of performance from time to time) but I would say that's normal and that should be better with the extensions, no?

```
|    ep_rew_mean      | 103      |
|    total timesteps  | 2620146  |
```
(this is the training reward with \epsilon = 0.1, the real performance of the agent is higher with \epsilon = 0.05)

Miffyli:
> DQN is a bit unstable on Breakout (catastrophic drop of performance from time to time) but I would say that's normal and that should be better with the extensions, no?

It has been long since I last time used vanilla DQN, but yeah this can happen, but it should _eventually_ level out. I agree we could keep the preprocessing code in this repo, just to make it clear what kind of preprocessing is done and it is not messed up by different `gym` versions.

araffin:
Sample results from benchmark:

```

# BreakoutNoFrameskip-v4
|    ep_rew_mean      | 148      |
|    exploration rate    | 0.1      |
|    total timesteps  | 9999607  |


# SpaceInvadersNoFrameskip-v4
|    total timesteps  | 9997681  |
|    exploration rate    | 0.1      |
|    ep_rew_mean      | 931      |
```

I will re-launch experiments with smaller buffer and lower exploration rate because I've got some interesting early results:
```
# BreakoutNoFrameskip
|    ep_rew_mean      | 184      |
|    exploration rate | 0.05     |
```

stable_baselines3/dqn/policies.py
araffin: After thinking more about this code, the main issue is the feature extraction part.
If we remove that class, it makes it harder to properly have a target network...

araffin:
Closing this as it is merged with master but Github is buggy right now...

araffin:
> I suspect the difference may not come from the algorithm but from our pre-processing/feature extraction.

I was unfortunately right.. see #132 
Now DQN on Breakout is much more stable and that would also explain why it did not work so well with `tau != 1`....

PartiallyTyped:
Did you by any chance compare with tf rmsprop, or adam vs rmsprop?

PR #30, opened 2020-05-23
Tensorboard integration

### LINKED ISSUE #9: Tensorboard integration
Maybe do that in the logger...
It should be optional as it requires tensorflow to be installed...

@rolandgvc is working on it

rolandgvc:
## Description

Adding support for logging to tensorboard.

Missing:
- [x] Documentation
- [x] More tests
- [x]  check we don't make the same mistakes as SB2 (https://github.com/hill-a/stable-baselines/issues/855 https://github.com/hill-a/stable-baselines/issues/56 )

## Motivation and Context

closes #9 

## Types of changes

## Checklist:


stable_baselines3/ppo/ppo.py
```diff
@@ -296,6 +305,38 @@ def train(self, n_epochs: int, batch_size: int = 64) -> None:
         if hasattr(self.policy, 'log_std'):
             logger.logkv("std", th.exp(self.policy.log_std).mean().item())
 
+        # TODO extract learning rate from optimizer?
+        if self.tb_writer is not None:
```
araffin, line 309: my idea was to avoid duplicating the logger and do something similar to https://github.com/hill-a/stable-baselines/blob/master/stable_baselines/logger.py#L218 or https://github.com/rlworkgroup/dowel/blob/master/src/dowel/tensor_board_output.py#L31

stable_baselines3/ppo/ppo.py
```diff
@@ -296,6 +305,38 @@ def train(self, n_epochs: int, batch_size: int = 64) -> None:
...
+            self.tb_writer.add_scalar("input_info/discounted_rewards",
+                                      np.mean(self.rollout_buffer.rewards), self.num_timesteps)
+            self.tb_writer.add_scalar("input_info/advantage", np.mean(self.rollout_buffer.advantages), self.num_timesteps)
+            self.tb_writer.add_scalar("input_info/clip_range", clip_range, self.num_timesteps)
+            if self.clip_range_vf is not None:
+                self.tb_writer.add_scalar("input_info/clip_range_vf", clip_range_vf, self.num_timesteps)
+            self.tb_writer.add_scalar("input_info/old_log_probs", np.mean(old_log_probs), self.num_timesteps)
+            self.tb_writer.add_scalar("input_info/old_values", np.mean(old_values), self.num_timesteps)
+
+            if self.full_tensorboard_log:
```
araffin, line 328: I would remove the `full_tensorboard_log` option too.

stable_baselines3/ppo/ppo.py
rolandgvc: Agreed. Didn't know about it :)

rolandgvc: I got it working. So 
```python
logger.logkv("loss/entropy_loss", np.mean(entropy_losses))
```
will log to both stdout and tensorboard. However, I can't figure out how to send logs to only one of the outputs of the `Logger`, e.g. 
```python
logger.logkv("iterations", iteration)
```
Any ideas?

araffin: maybe additional keyword argument to `logkv` ?

rolandgvc: Makes sense 👍

rolandgvc: How does this look?
```python
def logkv(key: Any, val: Any, exclude: Optional[List[str]]) -> None:
    """
    Log a value of some diagnostic
    Call this once for each diagnostic quantity, each iteration
    If called many times, last value will be used.

    :param key: (Any) save to log this key
    :param val: (Any) save to log this value
    :param exclude: (List[str]) list of outputs to be excluded
    """

    Logger.CURRENT.logkv(key, (val, exclude))
```

rolandgvc:
How was tensorboard creating PPO_1, PPO_2 etc. for every experiment in sb2?

araffin:
>How was tensorboard creating PPO_1, PPO_2 etc. for every experiment in sb2?

https://github.com/hill-a/stable-baselines/blob/master/stable_baselines/common/base_class.py#L1138

stable_baselines3/common/logger.py
```diff
@@ -21,7 +27,7 @@ class KVWriter(object):
     Key Value writer
     """
 
-    def writekvs(self, kvs: Dict) -> None:
+    def writekvs(self, kvs: Dict, step: int) -> None:
```
araffin, line 30: please give it a default value, so we keep the previous behavior if not using tb

stable_baselines3/common/logger.py
```diff
@@ -99,6 +110,9 @@ def writekvs(self, kvs: Dict) -> None:
 
     @classmethod
     def _truncate(cls, string: str) -> str:
+        # Remove tensorboard tag from std outputs
+        if string.find('/') > 0:
```
araffin, line 114: why? it more readable with the tag in std out, no?

stable_baselines3/common/logger.py
```diff
@@ -65,10 +71,15 @@ def __init__(self, filename_or_file: Union[str, TextIO]):
...
 
-    def writekvs(self, kvs: Dict) -> None:
+    def writekvs(self, kvs: Dict, step: int) -> None:
         # Create strings for printing
         key2str = {}
         for (key, val) in sorted(kvs.items()):
+            if val[1] is not None:
+                if 'stdout' in val[1]:
+                    continue
+            val = round(val[0], 5)
```
araffin, line 81: ?

rolandgvc: I thought rounding would look nicer as some values can be very long

stable_baselines3/common/logger.py
rolandgvc: Some keys would get quite long and being truncated because of the tags. Maybe we can output the tag as a parent and the keys under as children?

rolandgvc: nevermind, I've enlarged the truncate threshold a bit and it looks nice as it is as well

stable_baselines3/common/logger.py
rolandgvc: I've removed it, not much of a difference anyways

stable_baselines3/common/logger.py
araffin: >Some keys would get quite long and being truncated because of the tags. Maybe we can output the tag as a parent and the keys under as children?

this is a good idea

stable_baselines3/common/logger.py
```diff
@@ -65,10 +71,15 @@ def __init__(self, filename_or_file: Union[str, TextIO]):
             self.file = filename_or_file
             self.own_file = False
 
-    def writekvs(self, kvs: Dict) -> None:
+    def writekvs(self, kvs: Dict, step: int) -> None:
         # Create strings for printing
         key2str = {}
         for (key, val) in sorted(kvs.items()):
+            if val[1] is not None:
```
araffin, line 78: missing comment

stable_baselines3/common/logger.py
```diff
@@ -99,7 +110,7 @@ def writekvs(self, kvs: Dict) -> None:
 
     @classmethod
     def _truncate(cls, string: str) -> str:
-        return string[:20] + '...' if len(string) > 23 else string
+        return string[:20] + '...' if len(string) > 30 else string
```
araffin, line 113: I know this was not the case before, but I think this is a good time to have variables now instead of magic numbers ;)

stable_baselines3/common/logger.py
```diff
@@ -230,43 +266,49 @@ def make_output_format(_format: str, log_dir: str, log_suffix: str = '') -> KVWr
...
     If called many times, last value will be used.
 
     :param key: (Any) save to log this key
     :param val: (Any) save to log this value
+    :param exclude: (str or tuple) outputs to be excluded
     """
-    Logger.CURRENT.logkv(key, val)
+    if isinstance(exclude, str):
+        exclude = exclude,
+    Logger.CURRENT.logkv(key, (val, exclude))
```
araffin, line 281: not sure to get why you pass a tuple and not an extra argument...

stable_baselines3/ppo/ppo.py
```diff
@@ -331,21 +354,19 @@ def learn(self,
...
+                logger.logkv("iterations", iteration, exclude='tensorboard')
                 if len(self.ep_info_buffer) > 0 and len(self.ep_info_buffer[0]) > 0:
-                    logger.logkv('ep_rew_mean', self.safe_mean([ep_info['r'] for ep_info in self.ep_info_buffer]))
-                    logger.logkv('ep_len_mean', self.safe_mean([ep_info['l'] for ep_info in self.ep_info_buffer]))
-                logger.logkv("fps", fps)
-                logger.logkv('time_elapsed', int(time.time() - self.start_time))
-                logger.logkv("total timesteps", self.num_timesteps)
-                logger.dumpkvs()
+                    logger.logkv('ep_rew_mean', self.safe_mean([ep_info['r']
+                                                                for ep_info in self.ep_info_buffer]), exclude='tensorboard')
```
araffin, line 360: why do you exclude that one?
this is one of the most helpful metrics....

stable_baselines3/common/logger.py
rolandgvc: Well this way we have support for multiple exclusions

araffin: I meant:
```python
logkv(key, val, ['one', 'two', 'three'])
```
would work, no?
my remark is more about `(val, exclude)` instead of `val, exclude`

rolandgvc: because the exclusion would still be kept in the dictionary and easier to process

rolandgvc: Would you like a `self.exclusions` along `self.name2val` instead?

stable_baselines3/common/logger.py
rolandgvc: Ah, okay

stable_baselines3/ppo/ppo.py
rolandgvc: Visualizing Iterations in tensorboard? Isn't it just a linear plot?

stable_baselines3/common/logger.py
araffin: yes, please

stable_baselines3/ppo/ppo.py
araffin: `ep_rew_mean` is the mean episode reward

rolandgvc: Ah, by mistake I guess

stable_baselines3/common/logger.py
```diff
@@ -27,7 +27,7 @@ class KVWriter(object):
     Key Value writer
     """
 
-    def writekvs(self, kvs: Dict, step: int) -> None:
+    def writekvs(self, kvs: Dict, exclusions: List, step: int) -> None:
```
araffin, line 30: Please precise the type of list, e.g. `List[str]`

stable_baselines3/common/logger.py
```diff
@@ -444,8 +444,9 @@ def logkv(self, key: Any, val: Any) -> None:
         :param val: (Any) save to log this value
         """
         self.name2val[key] = val
+        self.exclusions.append(exclude)
```
araffin, line 447: why not a dict?

stable_baselines3/ppo/ppo.py
```diff
@@ -282,19 +289,34 @@ def train(self, n_epochs: int, batch_size: int = 64) -> None:
...
+        logger.logkv("session/episode_reward", np.sum(self.rollout_buffer.rewards))
+        logger.logkv("input_info/discounted_rewards", np.sum(self.rollout_buffer.returns))
+        logger.logkv("input_info/advantage", np.mean(self.rollout_buffer.advantages))
+        logger.logkv("input_info/infoclip_range", clip_range)
+        if self.clip_range_vf is not None:
+            logger.logkv("input_info/clip_range_vf", clip_range_vf)
+        logger.logkv("input_info/old_log_probs", np.mean(old_log_probs))
+        logger.logkv("input_info/old_values", np.mean(old_values))
+
+        # Histograms
```
araffin, line 313: please remove those, I don't think we need them, no?
(unless you are doing full debug)

stable_baselines3/common/logger.py
rolandgvc: You're right

rolandgvc:
Forgot to ask, do you still want to have the graph of the model in tensorboard?

araffin:
>Forgot to ask, do you still want to have the graph of the model in tensorboard?

why not, but how easy is it to get?

stable_baselines3/common/logger.py
```diff
@@ -190,25 +213,40 @@ def close(self) -> None:
...
-    """
-    try:
-        float(value)
-        return True
-    except TypeError:
-        return False
+    def writekvs(self, kvs: Dict, excs: Dict, step: int) -> None:
+        for (key, val), (_, exc) in zip(sorted(kvs.items()), sorted(excs.items())):
+            if exc is not None:
+                if 'tensorboard' is exc or 'tensorboard' in exc:
```
araffin, line 227: please use complete name (avoid `exc` and prefer `excluded`)
most of the logger code is from the old baselines... so not the best practices for naming...

araffin: maybe a good time to rename `kvs` to something with more meaning

rolandgvc: Got it, will do a full renaming for logger then :)

rolandgvc:
Graph still requires to have a FileWriter object passed around, I was trying to get away without it. I will look into it

rolandgvc:
Oops, my mistake :) Works now

rolandgvc:
So what I've done was to create the writer in `ppo.py` and pass it through many functions to the tensorboard output. Honestly it looks ugly and I wouldn't keep it, especially that the graph is not very helpful either. I would skip it for now.
Screenshot 2020-05-25 at 19 01 04

rolandgvc:
Excluding the graph, the tb logger now gives the same outputs as in sb2 . However, I'm noticing half-baked support for csv and json outputs. Are these supposed to be "hidden" features for advanced users? I'll wait for your review of the renamings before making the changes to all models.

araffin:
>I would skip it for now.

ok

>However, I'm noticing half-baked support for csv and json outputs.

What do you mean?
The logger is mainly undocumented as it is mostly an internal feature..

> I'll wait for your review of the renamings before making additional changes.

Please fix the issues in the tests first.

stable_baselines3/common/logger.py
```diff
@@ -21,18 +27,21 @@ class KVWriter(object):
     Key Value writer
     """
 
-    def writekvs(self, kvs: Dict) -> None:
+    def write_key_values(self, key_values: Dict, key_excluded: Dict, step: int) -> None:
```
araffin, line 30: Can't we merge `writekvs` and `writeseq`? 
so we can simply name it `write()`

stable_baselines3/common/logger.py
```diff
@@ -230,52 +276,54 @@ def make_output_format(_format: str, log_dir: str, log_suffix: str = '') -> KVWr
 # API
 # ================================================================
 
-def logkv(key: Any, val: Any) -> None:
+def log_key_value(key: Any, value: Any, exclude: Union[str, Tuple[str]] = None) -> None:
```
araffin, line 279: what about `record` instead of `log_key_value` ?

stable_baselines3/common/logger.py
```diff
@@ -230,52 +276,54 @@ def make_output_format(_format: str, log_dir: str, log_suffix: str = '') -> KVWr
...
 
     :param key_values: (dict) the list of keys and values to save to log
     """
     for key, value in key_values.items():
-        logkv(key, value)
+        log_key_value(key, value)
 
 
-def dumpkvs() -> None:
+def dump_key_values(step: int = 0) -> None:
```
araffin, line 313: what about `dump()` or `dump_log()` (as there not other `dump` function, no?)

stable_baselines3/common/logger.py
```diff
@@ -230,52 +276,54 @@ def make_output_format(_format: str, log_dir: str, log_suffix: str = '') -> KVWr
...
     """
     Write all of the diagnostics from the current iteration
     """
-    Logger.CURRENT.dumpkvs()
+    Logger.CURRENT.dump_key_values(step)
 
 
-def getkvs() -> Dict:
+def get_key_values() -> Dict:
```
araffin, line 320: `get_log_dict()` ?

stable_baselines3/common/logger.py
```diff
@@ -466,8 +521,8 @@ def close(self) -> None:
         """
         closes the file
         """
-        for fmt in self.output_formats:
-            fmt.close()
+        for formats in self.output_formats:
```
araffin, line 524: `format`

stable_baselines3/common/utils.py
```diff
@@ -110,3 +112,19 @@ def get_device(device: Union[th.device, str] = 'auto') -> th.device:
...
+
+def get_latest_run_id(log_path, log_name):
+    """
+    Returns the latest run number for the given log name and log path,
+    by finding the greatest number in the directories.
+
+    :return: (int) latest run number
+    """
+    max_run_id = 0
+    for path in glob.glob("{}/{}_[0-9]*".format(log_path, log_name)):
```
araffin, line 125: please use f-string

araffin: (and add typing)

rolandgvc: Regarding typing, I see in the code sometimes `Optional[str]` and sometimes `Optional[str] = None`. What would be the differences between these and `str = None`?

araffin: `str = None` is not correct, `Optional` means that the variable can be `None` (but `None` is not a string)

`Optional[str] = None` is when you give a default value

stable_baselines3/common/logger.py
rolandgvc: You mean eliminating SeqWriter class and have a single function?
```python
def write(self, key_values: Dict, key_excluded: Dict, sequence: List, step: int) -> None:
```

araffin: Imaybe not...
or rename to `write()`, `write_sequence()`
to be honest, I did not really take the time to read and understand what does `ẁrite_seq` exactly do, but the name is quite obscure...

rolandgvc:
I noticed something, `dump()` happens if `verbose>=1`, but we still log outside that if statement nonetheless. Isn't this wasteful?

rolandgvc:
Also, I have a bug with td3. Ideas?
```
  File "/Users/rolandgavrilescu/Github/stable-baselines3/stable_baselines3/td3/td3.py", line 189, in learn
    self.train(gradient_steps, batch_size=self.batch_size, policy_delay=self.policy_delay)
  File "/Users/rolandgavrilescu/Github/stable-baselines3/stable_baselines3/td3/td3.py", line 118, in train
    next_actions = (self.actor_target(replay_data.next_observations) + noise).clamp(-1, 1)
TypeError: 'NoneType' object is not callable
```

araffin:
>Also, I have a bug with td3. Ideas?

Could you give more context? (minimal example to reproduce the error)

> Isn't this wasteful? Or should we dump outside the if statement?

I'm not sure for that one, maybe pass "verbose" to the logger and let the different format choose what to do with it because it would be counter-intuitive to not log to tensorboard if a tensorboard folder is passed but `verbose=0`.

stable_baselines3/sac/sac.py
```diff
@@ -259,6 +267,11 @@ def learn(self,
                                      n_eval_episodes, eval_log_path, reset_num_timesteps)
         callback.on_training_start(locals(), globals())
 
+        if self.tensorboard_log is not None and SummaryWriter is not None:
```
araffin, line 270: Looks like something that should be done in the base class, no? (to avoid code duplication)

rolandgvc:
> Could you give more context? (minimal example to reproduce the error)

I just tried running the example from docs (no other modifications):
```python
# The noise objects for TD3
n_actions = env.action_space.shape[-1]
action_noise = NormalActionNoise(mean=np.zeros(n_actions), sigma=0.1 * np.ones(n_actions))

model = TD3(MlpPolicy, 'Pendulum-v0', action_noise=action_noise, verbose=1)
model.learn(total_timesteps=10000, log_interval=10)
```

araffin:
> > Could you give more context? (minimal example to reproduce the error)
> 
> I just tried running the example from docs (no other modifications):
> 
> ```python
> # The noise objects for TD3
> n_actions = env.action_space.shape[-1]
> action_noise = NormalActionNoise(mean=np.zeros(n_actions), sigma=0.1 * np.ones(n_actions))
> 
> model = TD3(MlpPolicy, 'Pendulum-v0', action_noise=action_noise, verbose=1)
> model.learn(total_timesteps=10000, log_interval=10)
> ```

this should work (the `gym.make` is missing... we should update the doc):
```
import numpy as np
import gym

from stable_baselines3 import TD3
from stable_baselines3.td3.policies import MlpPolicy
from stable_baselines3.common.noise import NormalActionNoise, OrnsteinUhlenbeckActionNoise

env = gym.make('Pendulum-v0')
# The noise objects for TD3
n_actions = env.action_space.shape[-1]
action_noise = NormalActionNoise(mean=np.zeros(n_actions), sigma=0.1 * np.ones(n_actions))

model = TD3(MlpPolicy, 'Pendulum-v0', action_noise=action_noise, verbose=1)
model.learn(total_timesteps=10000, log_interval=10)
```

rolandgvc:
>this should work (the gym.make is missing... we should update the doc):

Tried it as well, even with a fresh clone of the master. Same issue. You should be able to reproduce it as well.

stable_baselines3/ppo/ppo.py
araffin: i don't see the real need for most users

araffin:
i tried locally and in a colab notebook: https://colab.research.google.com/github/Stable-Baselines-Team/rl-colab-notebooks/blob/sb3/stable_baselines_getting_started.ipynb

stable_baselines3/common/base_class.py
```diff
@@ -874,21 +886,22 @@ def collect_rollouts(self,
                     action_noise.reset()
 
                 # Display training infos
-                if self.verbose >= 1 and log_interval is not None and self._episode_num % log_interval == 0:
+                if log_interval is not None and self._episode_num % log_interval == 0:
+                    stdout = "" if self.verbose >= 1 else "stdout"
```
araffin, line 890: missing comment and maybe add an option in the logger to completely exclude one output, no?
in fact, it would be simpler not to create that output

rolandgvc: But some logs are still sent to stdout, no?

araffin: when verbose=0, normally not

rolandgvc: So no logging at all to stdout when verbose=0?

araffin: yes ;)

rolandgvc:
I think this should be it

rolandgvc:
I think this should be it

stable_baselines3/ppo/ppo.py
```diff
@@ -282,19 +276,24 @@ def train(self, n_epochs: int, batch_size: int = 64) -> None:
...
+        logger.record("loss/value_loss", np.mean(value_losses))
+        logger.record("loss/approx_kl", np.mean(approx_kl_divs))
+        logger.record("loss/clip_fraction", np.mean(clip_fraction))
+        logger.record("loss/loss", loss.item())
+        logger.record("loss/explained_variance", explained_var)
+        if hasattr(self.policy, "log_std"):
+            logger.record("loss/std", th.exp(self.policy.log_std).mean().item())
+
+        logger.record("session/n_updates", self._n_updates, exclude="tensorboard")
+        logger.record("session/episode_reward", np.sum(self.rollout_buffer.rewards))
```
araffin, line 291: you can remove that one (and it seems buggy, I tested it with the rl zoo using `python train.py --algo ppo --env CartPole-v1 -tb /tmp/sb3`)

araffin: in fact, this one is not correct

araffin:
@rolandgvc I will try to do the full review this week, for now it looks good ;)
only minor changes needed I think (regarding the sections).

araffin:
@Miffyli @AdamGleave @hill-a @erniejunior  I would need your thought on that one regarding the name of the logger categories.

@rolandgvc had those three:
- session/
- loss/
- input_info/

I replaced them by:
- session/ -> number of timesteps, time elapsed, ..
- train/ -> losses, approx kl, clip fraction, learning rate, explained variance
- rollout/ -> mean episode reward, mean episode length, ...

but I'm not sure about the names... Also, I would like to maybe  use context manager to define those.

araffin:
@rolandgvc I update the description and added some TODOs ;) (mainly quick checks and documentation)
otherwise, I could test your work successfully with the zoo =)

AdamGleave:
> * session/ -> number of timesteps, time elapsed, ..
> * train/ -> losses, approx kl, clip fraction, learning rate, explained variance
> * rollout/ -> mean episode reward, mean episode length, ...
> 
> but I'm not sure about the names... Also, I would like to maybe use context manager to define those.

These names seem good to me. Maybe `time/` rather than `session/` since it seems to just be times + FPS? 

I wonder if `rollout/episodes` should be in `session/episodes` instead? It feels dual to `session/total_timesteps`.

rolandgvc:
> @rolandgvc I updated the description and added some TODOs ;) (mainly quick checks and documentation)
> otherwise, I could test your work successfully with the zoo =)

The tb logger logs tensors (len>1) as histograms. Should we mention this through the callback example in the docs?

araffin:
>The question about histograms remains, should we briefly mention it or exemplify it?

This belong to tensorboard/pytorch documentation, no?

docs/guide/tensorboard.rst
```diff
@@ -0,0 +1,104 @@
...
+        Custom callback for plotting additional values in tensorboard.
+        """
+
+        def __init__(self, verbose=0):
+            super(TensorboardCallback, self).__init__(verbose)
+
+        def _on_step(self) -> bool:
+            # Log scalar value (here a random variable)
+            value = np.random.random()
+            logger.record('random_value', value)
```
araffin, line 99: in fact, I did a commit to have access to it, you can just do `self.logger`

stable_baselines3/common/utils.py
```diff
@@ -110,3 +119,41 @@ def get_device(device: Union[th.device, str] = 'auto') -> th.device:
...
+    max_run_id = 0
+    for path in glob.glob(f"{log_path}/{log_name}_[0-9]*"):
+        file_name = path.split(os.sep)[-1]
+        ext = file_name.split("_")[-1]
+        if log_name == "_".join(file_name.split("_")[:-1]) and ext.isdigit() and int(ext) > max_run_id:
+            max_run_id = int(ext)
+    return max_run_id
+
+
+def configure_logger(verbose, tensorboard_log, tb_log_name, reset_num_timesteps=True):
```
araffin, line 140: type hints missing

stable_baselines3/ppo/ppo.py
```diff
@@ -305,17 +301,20 @@ def learn(self,
...
-              reset_num_timesteps: bool = True) -> 'PPO':
+              reset_num_timesteps: bool = True) -> "PPO":
 
         iteration = 0
         callback = self._setup_learn(eval_env, callback, eval_freq,
                                      n_eval_episodes, eval_log_path, reset_num_timesteps)
 
-        # if self.tensorboard_log is not None and SummaryWriter is not None:
-        #     self.tb_writer = SummaryWriter(log_dir=os.path.join(self.tensorboard_log, tb_log_name))
+        utils.configure_logger(self.verbose, self.tensorboard_log, tb_log_name, reset_num_timesteps)
```
araffin, line 310: why not doing that in `setup_learn` ? This is also where `num_timesteps` is reset or not

tests/test_tensorboard.py
```diff
@@ -0,0 +1,44 @@
+import os
+import shutil
+
+import pytest
+
+from stable_baselines3 import A2C, PPO, SAC, TD3
+
+TENSORBOARD_DIR = '/tmp/tb_dir/'
```
araffin, line 8: please use pytest `tmp_path` (cf pytest doc or logger test for an example)

tests/test_tensorboard.py
```diff
@@ -0,0 +1,44 @@
...
+    'ppo': (PPO, 'CartPole-v1'),
+    'sac': (SAC, 'Pendulum-v0'),
+    'td3': (TD3, 'Pendulum-v0'),
+}
+
+N_STEPS = 1000
+
+
+@pytest.mark.parametrize("model_name", MODEL_DICT.keys())
+def test_tensorboard(model_name):
```
araffin, line 24: can't you merge the two?

stable_baselines3/ppo/ppo.py
rolandgvc: Could you expand? In `configure_logger` I make sure run dir index is kept the same.

rolandgvc:
> > The question about histograms remains, should we briefly mention it or exemplify it?
> 
> This belong to tensorboard/pytorch documentation, no?

What I meant was to let the user know that if they log a tensor they would get the histogram as well.

stable_baselines3/ppo/ppo.py
araffin: I meant that you could put `configure_logger` inside `setup_learn`, that would avoid duplicating code

araffin:
>What I meant was to let the user know that if they log a tensor they would get the histogram as well.

But currently, we don't log any tensor, no?

rolandgvc:
We don't, but I thought it should be mentioned in case users would want to

araffin:
>We don't, but I thought it should be mentioned in case users would want to

Do it if you want, but I don't think it is really part of SB documentation.

stable_baselines3/td3/td3.py
```diff
@@ -167,11 +167,10 @@ def learn(self,
               reset_num_timesteps: bool = True) -> OffPolicyRLModel:
 
         callback = self._setup_learn(eval_env, callback, eval_freq,
-                                     n_eval_episodes, eval_log_path, reset_num_timesteps)
+                                     n_eval_episodes, eval_log_path, reset_num_timesteps,
+                                     self.verbose, self.tensorboard_log, tb_log_name)
```
araffin, line 171: why do you pass `self.verbose`? You have access to it inside the class...

rolandgvc: Ah good catch

PR #34, opened 2020-05-26
Implemented Vectorized Action Noise

### LINKED ISSUE #33: [feature-request] Vectorized/Stacked Action Noise
I am opening an issue to discuss the implementation and leave the PR for the code review.

The motivation behind this addition is that for OU or other noise processes we need the state to exist until the end of the trajectory. We can't use a single process for multiple parallel episodes as they have different lengths. Thus, stacked/vectorized processes allow us to keep the processes for as long as the particular episode goes.

My internal code (from the old issue in SB2).

```python

class VecNoise(ActionNoise):

    def __init__(self, base_class: Type[ActionNoise], num_noise: int, **noise_params):
        super().__init__()
        constr = partial(base_class, **noise_params)
        self._noise: List[ActionNoise] = [constr() for _ in range(num_noise)]

    def reset(self) -> None:
        """
        Call reset in all noise objects
        """
        for noise in self._noise:
            noise.reset()

    def reset_by_index(self, indices: Iterable[int]) -> None:
        """
        Args:
            indices: the indices to reset
        """
        for index in indices:
            self._noise[index].reset()

    def __repr__(self) -> str:
        return "VecNoise({})".format(repr(self._noise[0]))

    def __call__(self) -> np.ndarray:
        return np.stack([noise() for noise in self._noise])

```

An alternative is to provide the noise directly and copy/deepcopy it

```python

class VecNoise(ActionNoise):

    def __init__(self, base_noise:ActionNoise, num_noise: int):
        super().__init__()
        self._noise: List[ActionNoise] = []
        for _ in range(num_noise):
            noise = copy.deepcopy(base_noise)
            noise.reset() 
            self._noise.append(noise)
   ...
```
Or
```python

class VecNoise(ActionNoise):

    def __init__(self, base_noise:ActionNoise, num_noise: int):
        super().__init__()
        self._noise: List[ActionNoise] = tuple(map(copy.deepcopy, [base_noise]*num_noise))
        tuple(map(ActionNoise.reset, self._noise))
   ...
```

PartiallyTyped:
Vectorized Action Noise allows for multiple instances of
ActionNoiseProcesses to run in parallel. This makes it easier to
run TD3/SAC/DDPG with VecEnv.

## Description
The motivation behind this addition is that for OU or other noise processes we need the state to exist until the end of the trajectory. We can't use a single process for multiple parallel episodes as they have different lengths. Thus, stacked/vectorized processes allow us to keep the processes for as long as the particular episode goes.

## Motivation and Context

closes #33 

## Types of changes

## Checklist:

### DIFF
--- a/stable_baselines3/common/noise.py
+++ b/stable_baselines3/common/noise.py
@@ -100,7 +100,7 @@ class VectorizedActionNoise(ActionNoise):
             self.n_envs = int(n_envs)
             assert self.n_envs > 0
         except (TypeError, AssertionError):
-            raise ValueError("Expected n_envs to be positive integer greater than 0")
+            raise ValueError(f"Expected n_envs={n_envs} to be positive integer greater than 0")
 
         self.base_noise = base_noise
         self.noises = [copy.deepcopy(self.base_noise) for _ in range(n_envs)]
--- a/stable_baselines3/common/noise.py
+++ b/stable_baselines3/common/noise.py
@@ -1,5 +1,6 @@
-from typing import Optional
+from typing import Optional, List, Iterable
 from abc import ABC, abstractmethod
+import copy
 
 import numpy as np
 
@@ -45,7 +46,7 @@ class NormalActionNoise(ActionNoise):
 
 class OrnsteinUhlenbeckActionNoise(ActionNoise):
     """
-    A Ornstein Uhlenbeck action noise, this is designed to aproximate brownian motion with friction.
+    An Ornstein Uhlenbeck action noise, this is designed to aproximate brownian motion with friction.
 
     Based on http://math.stackexchange.com/questions/1287634/implementing-ornstein-uhlenbeck-in-matlab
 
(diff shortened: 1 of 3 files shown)


stable_baselines3/common/noise.py
```diff
@@ -84,3 +85,55 @@ def reset(self) -> None:
...
+            self.noises[index].reset()
+
+    def __repr__(self) -> str:
+        base_repr = repr(self.noises[0]) if len(self.noises) else ""
+        return f"VecNoise(BaseNoise={base_repr}), n_envs={len(self.noises)})"
+
+    def __call__(self) -> np.ndarray:
+        """
+        Generate and stack the action noise from each noise object
+        if n_envs == 0, returns a 0x0 matrix.
```
araffin, line 123: why would you allow such thing?

stable_baselines3/common/noise.py
```diff
@@ -84,3 +85,55 @@ def reset(self) -> None:
...
+            noise = np.zeros((0, 0))
+        return noise
+
+    @property
+    def noises(self) -> List[ActionNoise]:
+        return self._noises
+
+    @noises.setter
+    def noises(self, noises: List[ActionNoise]) -> None:
+        self._noises = noises
```
araffin, line 137: you should check the length too, no?

tests/test_utils.py
```diff
@@ -107,3 +110,19 @@ def dummy_callback(locals_, _globals):
 
     episode_rewards, _ = evaluate_policy(model, model.get_env(), n_eval_episodes, return_episode_rewards=True)
     assert len(episode_rewards) == n_eval_episodes
+
+
+def test_VecNoise():
```
araffin, line 115:
```suggestion
def test_vec_noise():
```
to be consistent with the rest of the names

tests/test_utils.py
```diff
@@ -107,3 +110,19 @@ def dummy_callback(locals_, _globals):
...
+    base: ActionNoise = OrnsteinUhlenbeckActionNoise(mu, sigma)
+    vec = VectorizedActionNoise(base, 4)
+    assert vec().shape == (4, 4)
+    assert not (vec() == base()).all()
+    vec = VectorizedActionNoise(base, 0)
+    assert vec().shape == (0, 0)
+    mu = np.ones(4)
+    sigma = np.zeros(4)
+    base = NormalActionNoise(mu, sigma)
+    vec = VectorizedActionNoise(base, 4)
```
araffin, line 127: minor: add a small test for getter/setter

stable_baselines3/common/noise.py
PartiallyTyped: I also need to ensure it is not None.

stable_baselines3/common/noise.py
PartiallyTyped: Because I wanted to discuss this. Ranges alone don't consider the length since `range(negative)` is always valid even if it never gives elements so I wanted to add an assertion but, in general, the parameters are assumed to be valid and sound. So I want to suggest to set a 'standard way to handle these cases', e.g. throwing `AssertionError` or `ValueError`.

araffin: yes, please add an `assert n_envs > 0` in the constructor

stable_baselines3/common/noise.py
araffin: I meant `assert len(noises) == n_envs`

PartiallyTyped: Now that I think of it, if I put a setter, then it's assumed to be public, so it needs sanity checks, yes?

araffin: yes, that's what I meant

stable_baselines3/common/noise.py
```diff
@@ -84,3 +85,82 @@ def reset(self) -> None:
...
+
+    @property
+    def noises(self) -> List[ActionNoise]:
+        return self._noises
+
+    @noises.setter
+    def noises(self, noises: List[ActionNoise]) -> None:
+        noises = list(noises)  # raises TypeError if not iterable
+        if len(noises) != self.n_envs:
+            raise ValueError(f"Found {len(noises)}, expected {self.n_envs}")
```
araffin, line 151: why not an assert? (because here the error message does not say what is found)

stable_baselines3/common/noise.py
```diff
@@ -84,3 +85,82 @@ def reset(self) -> None:
...
+    :param base_noise: ActionNoise The noise generator to use
+    :param n_envs: (int) The number of parallel environments
+    """
+
+    def __init__(self, base_noise: ActionNoise, n_envs: int):
+        try:
+            self.n_envs = int(n_envs)
+            assert self.n_envs > 0
+        except (TypeError, AssertionError):
+            raise ValueError("Expected n_envs to be positive integer greater than 0")
```
araffin, line 103:
```suggestion
            raise ValueError(f"Expected n_envs={n_envs} to be positive integer greater than 0")
```
minor: give the value that was passed

stable_baselines3/common/noise.py
PartiallyTyped: I missed this one. I added an assertion there in the last commit.

Issue #17, opened 2020-05-11
Review of Existing Code


AdamGleave:
@araffin has done a great job creating this version of the library but has done it mostly solo, so much of the code has never been reviewed. 

I suggest the other maintainers each take responsibility for reviewing a portion of the code. Rather than doing a traditional code review, since it's already committed I suggest we just make a PR with any changes we think should take place, or raise an issue for non-trivial proposals.

I think the priority is code that is used in multiple algorithms and/or defines the public API and which is new. This includes:
  - [x] `common/base_class.py`
  - [x] `common/distributions.py`
  - [x] `common/policies.py`
  - [x] `common/type_aliases.py`

Next would be the individual algorithms:
  - [x] A2C
  - [x] PPO
  - [ ] SAC
  - [ ] TD3

Also new parts of the documentation could use re-reading/confirming:
  - [x] Documentation

If this sounds like a good idea to others, then perhaps we can each claim a few entries on the above list and then start a PR for it? Also feel free to edit the post to break it up differently or to add files I've missed.

AdamGleave:
Some self-review from @araffin (lightly edited, mistakes my own):
  - A2C:
     + Imports policy from PPO, could move it to common e.g. as `OnlineActorCriticPolicy`.
     + Also derives from PPO, so again introducing a base class e.g. `OnlineAlgorithm` could help.
  - TD3 and SAC should share part of their policy.
  - TD3: policy does not handle action noise so if saved independently of the RL algorithm it will only be a deterministic policy.

Miffyli:
I can do A2C and PPO review (core algorithms + files they use), along with updating them to share the code more elegantly as pointed above (`OnlineActorCriticPolicy` etc).

araffin:
@Miffyli nice =) (apparently, we can add Windows and mac os thanks to github ci (to be tested))
i think some `os.path.join` are missing, no?

i will ping @hill-a privately ;) (he needs some motivation but he can do things and his input is usually valuable)
maybe an additional "auto-review": i did not use "_" prefix for all private attributes.

AdamGleave:
I'm happy to review `common/base_class.py`, `common/distributions.py`, `common/policies.py` and `common/type_aliases.py` although may not have time until after NeurIPS deadline.

araffin:
Mentioned somewhere: `tmp_path` should be used everywhere in the tests instead of creating custom path

AdamGleave:
[...]

PR #35, opened 2020-05-26
Review of code (A2C, PPO and refactoring)

Miffyli:
A review of current code, especially on A2C and PPO. Also includes refactoring similar to what was done in stable-baselines.

This PR lives on as questions are answered and things are discussed out.

Related issue: #17 

## Description (what is currently included. Feel free to comment)
- Created new file common/torch_layers.py, similar to SB refactoring
  - Contains all PyTorch network layer definitions and feature extractors:
    `MlpExtractor`, `create_mlp`, `NatureCNN`
- Fixed reference in NatureCNN (pointed to older version with different network architecture)
- Fix comments saying "CxWxH" instead of "CxHxW" (same style as in torch docs / commonly used)
- Added bit further comments on register/getting policies.
- Moved PPOPolicy to "ActorCriticPolicy" in common/policies.py
- Split PPO (agent class) into OnPolicyAlgorithm, to be shared with A2C
  - Shares rollout collection and `learn` function. `train` is still defined per algorithm.
  - Removed bunch of default parameters from OnPolicyAlgorithm(does not make sense as stand-alone parameters)
- Rename `progress` (value from 1 in start of training to 0 in end) to `progress_remaining`.
- Rename `BaseRLModel` to `BaseAlgorithm` (along with offpolicy and onpolicy variants)
- Moved following functions from `BaseAlgorithm`: 
  - `_load_from_file` to `load_from_zip_file` (save_util.py)
  - `_save_to_file_zip` to `save_to_zip_file` (save_util.py)
  - `safe_mean` to `safe_mean` (utils.py)
  - `check_env` to `check_for_correct_spaces` (utils.py. Renamed to avoid confusion with environment checker tools)
- Move on-policy and off-policy base algorithms to `common/on_policy_algorithm.py` and `common/off_policy_algorithm.py`, respectively.
- Add `policies.py` files for A2C/PPO, which define MlpPolicy/CnnPolicy (renamed ActorCriticPolicies).
- Fix `register_policy` to allow re-registering same policy for same sub-class (i.e. assign same value to same key).
- Add bit of clarification Mlp/CnnPolicies on developer.rts
- Moved static function `_is_vectorized_observation` from common/policies.py to common/utils.py under name `is_vectorized_observation`.

## Open questions (please reply in comments)
- ~~Rename `OnlineActorCriticPolicy` to just `ActorCriticPolicy`? I do not see reason to call it "Online", unless I am missing something. It could be `OnPolicyActorCriticPolicy` if we can not drop the "on policy" part.~~
- ~~`BaseRLModel` seems to include functions that could be outside it (e.g. `safe_mean`). Should these be moved under `common/utils.py`?~~
- ~~Regarding image-transposing stuff: Should we just require environments to output CxHxW, and throw an error otherwise (for Conv2Ds it does not matter if it is CxWxH)? This would save space with the whole image-transposing stuff. I am not aware of Gym posing limitations on the order of image dimensions. I am not super-fan of the library modifying things like this quietly.~~
- ~~There seems to be overlapping use of cnn extractor and mlp (i.e. CNN followed by a small MLP, by default). I think this behaviour should be changed to "no extra layers after CNN extractor, if one is provided".~~
- ~~Relating `progression` in schedulers: Now the notion is that `progress=1.0` means we are at the beginning of the training, but this sounds counter-intuitive. `progress=0.0` means we would have zero progress, i.e. that would be the beginning of training. Should we change this?~~

## TODOs and personal checklist:
- ~~[ ] Think about VecTranspose~~
- [x] Go through documentation for e.g. missing mentions of OnPolicyAlgorithm.
- [x] Check that documentation is through-out for the "MlpPolicy/CnnPolicy" naming concept (for both users and developers). Also how we should have `[algorithm]/policies.py` files.
- [x] Double-check docstrings are not missing parameters / do not include other parameters
- ~~[ ] Check A2C/PPO performance matches stable-baselines in basic benchmarks and Atari.~~

## Types of changes

## Checklist:
- [ ] I have checked the codestyle using `make lint`


stable_baselines3/common/torch_layers.py
```diff
@@ -0,0 +1,218 @@
...
+        # Save dim, used to create the distributions
+        self.latent_dim_pi = last_layer_dim_pi
+        self.latent_dim_vf = last_layer_dim_vf
+
+        # Create networks
+        # If the list of layers is empty, the network will just act as an Identity module
+        self.shared_net = nn.Sequential(*shared_net).to(device)
+        self.policy_net = nn.Sequential(*policy_net).to(device)
+        self.value_net = nn.Sequential(*value_net).to(device)
+
```
PartiallyTyped, line 211: Weight initialization should be performed here. This could be done by passing the initialization function as a contructor parameter. This is necessary because according to https://openreview.net/forum?id=r1etN1rtPB , different initialization schemes result in different results. e.g. Xavier uniform (pytorch default) vs diagonal (openai's PPO and some other algorithms use diagonal).

araffin:
>Description (what is currently included. Feel free to comment)

Overall, I agree and I'm happy by all your changes =)

>Split PPO (agent class) into OnPolicyRLModel, to be shared with A2C 

Minor: I'm wondering if you should put everything in `base_class.py` or have one file per base class?

>Fix comments saying "CxWxH" instead of "CxHxW" (same style as in torch docs / commonly used)

Nice catch, I think I would have never noticed ^^

>Rename OnlineActorCriticPolicy to just ActorCriticPolicy? 

Ok, I agree the on/off-policy is more for the algorithm, but the docstring should be accurate (so people won't think you can use it with SAC for instance).

>eems to include functions that could be outside it (e.g. safe_mean). Should these be moved under common/utils.py?

yes if you think so. I don't have strong feeling on those helpers.

>Regarding image-transposing stuff: Should we just require environments to output CxHxW, and throw an error otherwise 
>his would save space with the whole image-transposing stuff. I am not aware of Gym posing limitations on the order of image dimensions. I am not super-fan of the library modifying things like this quietly.

Maybe the name `VecTranspose` is wrong. The main issue is that gym env are using tf convention: HxWxC (or WxHxC), the issue is that the channel dimension is last.

>CNN followed by a small MLP, by default

As mentioned in the DQN issue, I don't think this is the case (I do a check normally)

>means we are at the beginning of the training, but this sounds counter-intuitive. progress=0.0 means we would have zero progress, i.e. that would be the beginning of training. Should we change this?

This is more a legacy issue... That was how it was done in SB and would be quite a breaking change. Maybe renaming progress to `remaining_time` or something similar would be the solution?

docs/modules/td3.rst
```diff
@@ -14,7 +14,7 @@ We recommend reading `OpenAI Spinning guide on TD3 <https://spinningup.openai.co
 
 .. warning::
 
-  The TD3 model does not support ``stable_baselines3.ppo.policies`` because it uses double q-values
+  The TD3 model does not support ``stable_baselines3.common.policies`` because it uses double q-values
```
araffin, line 17: about that, I'm not sure if we should still keep those "common" policies as the default (because they don't work with 50% of the algorithm now). That's also why I try to use strings in the example to refer to a policy.

stable_baselines3/a2c/a2c.py
```diff
@@ -4,13 +4,13 @@
...
+from stable_baselines3.common.base_class import OnPolicyRLModel
 from stable_baselines3.common.type_aliases import GymEnv, MaybeCallback
 from stable_baselines3.common.utils import explained_variance
-from stable_baselines3.ppo.policies import PPOPolicy
-from stable_baselines3.ppo.ppo import PPO
+from stable_baselines3.common.policies import OnlineActorCriticPolicy
 
 
-class A2C(PPO):
+class A2C(OnPolicyRLModel):
```
araffin, line 13: Didn't we say we wanted to rename it to `BaseAlgorithm` and `BasePolicy` to `BaseModel` ?

stable_baselines3/a2c/a2c.py
```diff
@@ -72,16 +73,17 @@ def __init__(self, policy: Union[str, Type[PPOPolicy]],
...
                                   use_sde=use_sde, sde_sample_freq=sde_sample_freq,
                                   tensorboard_log=tensorboard_log, policy_kwargs=policy_kwargs,
                                   verbose=verbose, device=device, create_eval_env=create_eval_env,
                                   seed=seed, _init_setup_model=False)
 
         self.normalize_advantage = normalize_advantage
-        # Override PPO optimizer to match original implementation
+
+        # Update optimizer inside the policy if we want to use RMSProp rather
+        # than Adam
```
araffin, line 86:
```suggestion
        # Update optimizer inside the policy if we want to use RMSProp (original implementation)
        # rather than Adam
```

stable_baselines3/a2c/a2c.py
```diff
@@ -152,15 +154,15 @@ def train(self, gradient_steps: int, batch_size: Optional[int] = None) -> None:
     def learn(self,
               total_timesteps: int,
               callback: MaybeCallback = None,
-              log_interval: int = 100,
+              log_interval: int = 1,
```
araffin, line 157: I know you did that for consistency but the thing is that A2C uses usually small `n_steps` and as a results, it floods the terminal (compared to PPO for instance)

stable_baselines3/common/policies.py
```diff
@@ -373,42 +303,360 @@ def parameters_to_vector(self) -> np.ndarray:
...
+                                                      device,
+                                                      features_extractor_class,
+                                                      features_extractor_kwargs,
+                                                      optimizer_class=optimizer_class,
+                                                      optimizer_kwargs=optimizer_kwargs,
+                                                      squash_output=squash_output)
+
+        # Default network architecture, from stable-baselines
+        if net_arch is None:
+            if features_extractor_class == FlattenExtractor:
```
araffin, line 378: That's what I meant for the default network architecture with CNN (but I may be wrong)

stable_baselines3/common/policies.py
```diff
@@ -373,42 +303,360 @@ def parameters_to_vector(self) -> np.ndarray:
...
+            # TODO: check for features_extractor
+            for module in [self.features_extractor, self.mlp_extractor,
+                           self.action_net, self.value_net]:
+                # Values from stable-baselines, TODO: check why
+                gain = {
+                    self.features_extractor: np.sqrt(2),
+                    self.mlp_extractor: np.sqrt(2),
+                    self.action_net: 0.01,
+                    self.value_net: 1
+                }[module]
```
PartiallyTyped, line 490:
```suggestion
            modules = {
                    self.features_extractor: np.sqrt(2),
                    self.mlp_extractor: np.sqrt(2),
                    self.action_net: 0.01,
                    self.value_net: 1
            }
            for module, gain in modules.items():
```

Miffyli:
> Minor: I'm wondering if you should put everything in base_class.py or have one file per base class?
Actually I had the same idea. `base_class.py` is too large right now. I am going to split these into three files (base, offpolicy and onpolicy).

> Maybe the name VecTranspose is wrong. The main issue is that gym env are using tf convention: HxWxC (or WxHxC), the issue is that the channel dimension is last.

Hmm that's true (also the same convention for images in general). I am going to give this bit more thought how it could be cleaner to include.

> This is more a legacy issue... That was how it was done in SB and would be quite a breaking change. Maybe renaming progress to remaining_time or something similar would be the solution?

Fair point, but I agree we should at least rename the variable. I will include this.

docs/modules/td3.rst
Miffyli: I am not sure if I understand your point here.

stable_baselines3/a2c/a2c.py
Miffyli: Hmm I might have missed this one, but indeed "Algorithm" is better name for the algorithms. Perhaps the acting parts (class handling the network) should still be called a "Policy", as it does the acting/stepping? "Model" sounds more like just the inner-network part, that outputs some stuff and does not care what it does outside that.

stable_baselines3/a2c/a2c.py
Miffyli: Oh wups, I am not sure how this variable got changed. I admit I did it for consistency, but I somehow missed the semantic meaning of this variable. I will update this to "100" again!

docs/modules/td3.rst
araffin: in SB2, it was called `common.policies.MlpPolicy`because it was used by most algorithms (A2C, ACER, ACKTR, PPO, TRPO).
However, now, this is not the case anymore

Miffyli: Ah, right, I understand now. Perhaps if we move ActorCriticPolicies to a new file, e.g. `common/actor_critic_policy.py`? I would like to avoid the cross-algorithm imports so they do not interfere with each other and it is clear what is used by what :)

araffin: I understand your point. Maybe, keep it in `policies.py` but we should always use `from stable_baselines3.ppo import MlpPolicy` in the doc
(and anyway, in the doc, I try to use strings as it is the best practice not to fuck up things)

Miffyli: Now I feel your point: Importing algorithm-specific policies from "[algorithm_name].policies" translates well to any algorithm, rather than knowing that A2C/PPO need "common.policies", TD3 needs "td3.policies" etc. How about having "policies.py" for each algorithm, but they can then import a shared policy from common, i.e. only contain lines like

```python
# ppo/policies.py file
from stable_baselines3.common.policies import OnlineActorCriticPolicy, OnlineActorCriticCnnPolicy
```

araffin: it was my point ;) (just afraid we have to tweak the auto doc)

araffin:
> Check A2C/PPO performance matches stable-baselines in basic benchmarks and Atari.

I would do that in an issue, as it may take more time. This PR is also really valuable for the code structure so I would like to not delay it too much (because other PR will depend on it).

Miffyli:
> I would do that in an issue, as it may take more time. This PR is also really valuable for the code structure so I would like to not delay it too much (because other PR will depend on it).

Noted, I will leave this for another issue. I will continue on this PR later today when (hopefully) my Linux machine is free so I can do things properly and not wrestle with Windows.

Miffyli:
@araffin 

Things should be good to go now. Should all the changes be listed in changelog (I assume so), or are they part of "pre-release" tuning?

araffin:
>Things should be good to go now. Should all the changes be listed in changelog (I assume so), or are they part of "pre-release" tuning?

Yes, it is always good to document structure changes like that. Even for pre-release.

docs/modules/sac.rst
```diff
@@ -14,7 +14,7 @@ A key feature of SAC, and a major difference with common RL algorithms, is that
 
 .. warning::
 
-  The SAC model does not support ``stable_baselines3.ppo.policies`` because it uses double q-values
+  The SAC model does not support ``stable_baselines3.common.policies`` because it uses double q-values
```
araffin, line 17: I thought we decided to remove those mentions?

stable_baselines3/common/on_policy_algorithm.py
```diff
@@ -0,0 +1,225 @@
...
+                                              create_eval_env=create_eval_env, support_multi_env=True, seed=seed)
+
+        self.n_steps = n_steps
+        self.gamma = gamma
+        self.gae_lambda = gae_lambda
+        self.ent_coef = ent_coef
+        self.vf_coef = vf_coef
+        self.max_grad_norm = max_grad_norm
+        self.rollout_buffer = None
+        self.tensorboard_log = tensorboard_log
```
araffin, line 80: tensorboard things are defined in the base algorithm class normally

stable_baselines3/common/policies.py
```diff
@@ -183,12 +115,14 @@ def _predict(self, observation: th.Tensor, deterministic: bool = False) -> th.Te
...
 
-    def predict(self, observation: np.ndarray,
+    def predict(self,
+                observation: np.ndarray,
                 state: Optional[np.ndarray] = None,
                 mask: Optional[np.ndarray] = None,
                 deterministic: bool = False) -> Tuple[np.ndarray, Optional[np.ndarray]]:
         """
         Get the policy action and state from an observation (and optional state).
+        Includes sugar-coating to handle different observations (e.g. normalizing images).
```
araffin, line 125: I like the term ;)

stable_baselines3/ppo/ppo.py
```diff
@@ -91,41 +86,29 @@ def __init__(self, policy: Union[str, Type[PPOPolicy]],
...
-        self.gae_lambda = gae_lambda
         self.clip_range = clip_range
         self.clip_range_vf = clip_range_vf
-        self.ent_coef = ent_coef
-        self.vf_coef = vf_coef
-        self.max_grad_norm = max_grad_norm
-        self.rollout_buffer = None
         self.target_kl = target_kl
+
         self.tb_writer = None
```
araffin, line 103: this one can be removed, no?

PR #74, opened 2020-06-26
Automatic folder creation when saving models

### LINKED ISSUE #71: [Feature-Request] Automatically create path to save location
## Motivation
It's very unfortunate to lose a few hours of work simply because the destination folder doesn't exist ( or the path was mistyped ^^')

## Suggestion(s)
Either of:
- Automatically create the path to the destination folder and raise a warning if the path didn't exist.
- Save to /tmp/ if the path doesn't exist and raise a warning.

PartiallyTyped:
Automatically creates folders when saving models.

## Description

I created an `open_path(path, verbosity, extension) -> io.BaseIO` function that takes a path in the form of `str, pathlib.Path, io.BaseIO`. If the path is either of `str, pathlib.Path`, it ensures that the path:
 - exists
 - has the correct extension (.zip, .pkl)  

and opens it. If the path does not exist, it is created. If there is a conflict with a folder, a warning is raised and the path is changed to `{path}_2`. In both cases, a warning is raised. If verbosity is set to 2, and the path exists, another warning is raised.

ps. Added some type annotations.

## Motivation and Context

Closes #71 

## Types of changes

## Checklist:

### DIFF
--- a/stable_baselines3/common/base_class.py
+++ b/stable_baselines3/common/base_class.py
@@ -16,7 +16,7 @@ from stable_baselines3.common.utils import (set_random_seed, get_schedule_fn, up
 from stable_baselines3.common.vec_env import DummyVecEnv, VecEnv, unwrap_vec_normalize, VecNormalize, VecTransposeImage
 from stable_baselines3.common.preprocessing import is_image_space
 from stable_baselines3.common.save_util import (recursive_getattr, recursive_setattr, save_to_zip_file,
-                                                load_from_zip_file, open_path)
+                                                load_from_zip_file)
 from stable_baselines3.common.type_aliases import GymEnv, MaybeCallback
 from stable_baselines3.common.callbacks import BaseCallback, CallbackList, ConvertCallback, EvalCallback
 from stable_baselines3.common.monitor import Monitor
@@ -531,5 +531,4 @@ class BaseAlgorithm(ABC):
             # Retrieve state dict
             params_to_save[name] = attr.state_dict()
 
-        path = open_path(path, verbose=self.verbose)
         save_to_zip_file(path, data=data, params=params_to_save, tensors=tensors)
--- a/stable_baselines3/common/base_class.py
+++ b/stable_baselines3/common/base_class.py
@@ -2,6 +2,7 @@ import time
 from typing import Union, Type, Optional, Dict, Any, List, Tuple, Callable
 from abc import ABC, abstractmethod
 from collections import deque
+import pathlib
 
 import gym
 import torch as th
@@ -291,7 +292,7 @@ class BaseAlgorithm(ABC):
         return self.policy.predict(observation, state, mask, deterministic)
 
     @classmethod
-    def load(cls, load_path: str, env: Optional[GymEnv] = None, **kwargs):
+    def load(cls, load_path: str, env: Optional[GymEnv] = None, **kwargs) -> 'BaseAlgorithm':
         """
         Load the model from a zip-file
 
@@ -475,11 +476,11 @@ class BaseAlgorithm(ABC):
         """
         return ["policy", "device", "env", "eval_env", "replay_buffer", "rollout_buffer", "_vec_normalize_env"]
 
-    def save(self, path: str, exclude: Optional[List[str]] = None, include: Optional[List[str]] = None) -> None:
+    def save(self, path: Union[str, pathlib.Path], exclude: Optional[List[str]] = None, include: Optional[List[str]] = None) -> None:
         """
         Save all the attributes of the object and the model parameters in a zip-file.
 
-        :param path: path to the file where the rl agent should be saved
+        :param (Union[str,pathlib.Path]): path to the file where the rl agent should be saved
         :param exclude: name of parameters that should be excluded in addition to the default one
         :param include: name of parameters that might be excluded but should be included anyway
         """
--- a/stable_baselines3/common/base_class.py
+++ b/stable_baselines3/common/base_class.py
@@ -479,14 +479,14 @@ class BaseAlgorithm(ABC):
 
     def save(
         self,
-        path: Union[str, pathlib.Path, io.IOBase],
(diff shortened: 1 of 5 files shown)


Miffyli:
I see the core functionality is written twice, once in `save_utils` and second time in `off_policy_algorithm`. Would it be possible to have the same checks only in one place, e.g. `check_file_path_for_writing` in `save_utils`, which only checks if given filepath would be good for storing files, and possibly creates the necessary parent directories?

PartiallyTyped:
What about a function `open_path(str|Path)->io.BaseIO` that opens the file (and creates the necessary folders/warnings) that returns an FP or throws an error? Then add the option to save to `/tmp/` if opening fails in the save functions?

Also, I would like to add a verbose parameter for logging when `verbose==2`.

PartiallyTyped:
This is rather weird, I am unable to reproduce the test error on my machine.

stable_baselines3/common/save_util.py
```diff
@@ -174,32 +176,61 @@ def json_to_data(json_string: str,
     return return_data
 
 
-def save_to_zip_file(save_path: str, data: Dict[str, Any] = None,
-                     params: Dict[str, Any] = None, tensors: Dict[str, Any] = None) -> None:
+@functools.singledispatch
+def open_path(path, verbose=0, suffix=None) -> io.IOBase:
```
Miffyli, line 180: These functions could use proper docstrings. This function specifically only "works" as intended if `path` is `io.IOBase`, no? I think that should be checked here.

stable_baselines3/common/save_util.py
```diff
@@ -174,32 +176,61 @@ def json_to_data(json_string: str,
...
+    return open_path(path)
+
+
+def save_to_zip_file(save_path, data: Dict[str, Any] = None,
+                     params: Dict[str, Any] = None, tensors: Dict[str, Any] = None, verbose=0) -> None:
     """
     Save a model to a zip archive.
 
-    :param save_path: Where to store the model.
+    :param save_path: (Union[str, pathlib.Path, io.IOBase]) Where to store the model.
```
Miffyli, line 216: I think we could replace io.IOBase with io.BytesIO all around the code. I think all saving is done in bytes. This way we would not need casting :)

tests/test_save_load.py
```diff
@@ -1,5 +1,6 @@
 import os
```
Miffyli, line 1: If saving to `io.BytesIO` (or any IOBase) is supported, that should be tested. This is probably easy copy paste from [stable-baselines tests](https://github.com/hill-a/stable-baselines/blob/master/tests/test_save.py#L40).

stable_baselines3/common/save_util.py
```diff
@@ -174,32 +176,61 @@ def json_to_data(json_string: str,
...
+    return path
+
+
+@open_path.register(str)
+def open_path_str(path: str, verbose=0, suffix=None) -> io.IOBase:
+    return open_path(pathlib.Path(path), verbose=verbose, suffix=suffix)
+
+
+@open_path.register(pathlib.Path)
+def open_path_pathlib(path: pathlib.Path, verbose=0, suffix=None) -> io.IOBase:
```
Miffyli, line 191: Tests for these would be appreciated.

stable_baselines3/common/off_policy_algorithm.py
```diff
@@ -102,23 +104,26 @@ def _setup_model(self):
...
-    def load_replay_buffer(self, path: str):
+    def load_replay_buffer(self, path: Union[str, pathlib.Path]) -> None:
         """
         Load a replay buffer from a pickle file.
 
-        :param path: (str) Path to the pickled replay buffer.
+        :param path: (Union[str, pathlib.Path]) Path to the pickled replay buffer.
         """
-        with open(path, 'rb') as file_handler:
+        path = pathlib.Path(path)
```
Miffyli, line 123: I would prefer allowing any suffix here, if the file exists, [similar to how stable-baselines did it](https://github.com/hill-a/stable-baselines/blob/master/stable_baselines/common/base_class.py#L614). If file does not exist, then try with the default .pkl suffix.

stable_baselines3/common/base_class.py
```diff
@@ -472,11 +474,16 @@ def excluded_save_params(self) -> List[str]:
...
 
-    def save(self, path: str, exclude: Optional[List[str]] = None, include: Optional[List[str]] = None) -> None:
+    def save(
+        self,
+        path: Union[str, pathlib.Path, io.IOBase],
+        exclude: Optional[List[str]] = None,
+        include: Optional[List[str]] = None,
+    ) -> None:
         """
         Save all the attributes of the object and the model parameters in a zip-file.
```
Miffyli, line 484: Update docs to reflect the new behaviour ("If directory does not exist, create one" etc). Same for other "save" functions.

Miffyli:
Overall things look good, just bit of testing and docs needed :).

stable_baselines3/common/save_util.py
PartiallyTyped: The function works with any object that supports the `io.IOBase` api regardless of whether it inherits from it so, due to duck-typing, I left it as generic as possible and checking only the `str` and `Path` cases.

stable_baselines3/common/off_policy_algorithm.py
PartiallyTyped: I see, yes I will add it.

stable_baselines3/common/save_util.py
Miffyli: I see there is also "StringIO" which is subclass of "IOBase", but I do not think the stuff being stored here can be stored in text buffers. I was pondering if we should use BytesIO for that reason to make it clear it can not be _any_ IOBase, but something that handles bytes :)

PartiallyTyped:
My branch has been messed up ... I will open a new PR.

PR #81, opened 2020-06-30
N-step updates for off-policy methods

### LINKED ISSUE #47: [feature-request] N-step returns for TD methods
Originally posted by @PartiallyTyped in https://github.com/hill-a/stable-baselines/issues/821
"
N-step returns allow for much better stability, and improve performance when training DQN, DDPG etc, so it will be quite useful to have this feature.

A simple implementation of this would be as a wrapper around ReplayBuffer so it would work with both Prioritized and Uniform sampling. The wrapper keeps a queue of observed experiences compute the returns and add the experience to the buffer.
"

Roadmap: v1.1+ (see #1 )

partiallytyped:
#### OffPolicyAlgorithm

- Now accepts buffer class as argument type
  - Rationale: allows for other types of replay buffers besides the current implementations
- Now accepts arguments for the replay buffer
   - Rationale: makes it easier to pass arguments because we don't need to check every scenario.
   - I haven't moved `buffer_size` in it

#### NStepReplayBuffer

- Derives ReplayBuffer and adds two more arguments, `n_step` and `gamma`.
- Overrides `_get_samples(self, batch_inds, env) -> ReplayBufferSamples` to compute NStepReturns.
  - Computation is done purely through numpy without python loops. This makes the implementation easier and faster.

__Computing Nstep returns__

1. Find correct transition indices
2. Create a 2d mask for the rewards using done
3. Filter the mask so that `self.pos` is ignored because it contains older experience.
4. Multiply the filter with the reward, the discounting, and sum across.

---

## Motivation and Context

Closes #47 

## Types of changes

## Checklist:


araffin:
Funny, I also recently gave it a try here: https://github.com/DLR-RM/stable-baselines3/tree/feat/n-steps

PartiallyTyped:
The implementation appears mostly the same. I would like to add the class and the arguments as parameters to the `OffPolicyAlgorithm` along with a variable named `replay_args`, in a similar fashion to how policies are passed. This will make it easy to add other methods as well. IIRC energy based replay buffer was also considered at some point.

PartiallyTyped:
@araffin 
So I was working on this and there's a bug with both your and my implementations.
The bug is due to `% self.buffer_size`

https://github.com/DLR-RM/stable-baselines3/blob/41eb0cb37e8ac35413c0b11d0d5bd5cfbd2c92e2/stable_baselines3/common/buffers.py#L441-L454

The bug is basically the same as with memory optimization. To be exact, going around the buffer (% self.buffer_size) gives no quarantine that the transition is part of the same trajectory.

araffin:
>The bug is basically the same as with memory optimization.

:see_no_evil: 
Yep, I did only quick test with it and could not see any improvement yet.

>Completed it. Removed loops, added tests, using just numpy arrays, no loops. Should be quite fast.

Sounds good =)

Ping me when you want a review

PartiallyTyped:
I merged master and removed the generic from `offpolicyalgorithm` since it wasn't discussed. I think it's ready for a review.

araffin:
>Ps. sorry for autoformatting >.< ...

I will do a PR soon for that ;)
Apparently will be with black.

stable_baselines3/common/buffers.py
```diff
@@ -387,15 +380,81 @@ def get(self, batch_size: Optional[int] = None) -> Generator[RolloutBufferSample
...
+        # two dim matrix of rewards for the indices
+        rewards = self.rewards[indices].reshape(not_dones.shape)
+        rewards = self._normalize_reward(rewards, env)
+
+        # filter to ignore loops around the buffer, see:
+        # https://github.com/DLR-RM/stable-baselines3/pull/81#issuecomment-653741592
+        # basically we need to ignore indices that are equal to self.pos
+        # because the indice at self.pos is older
+        # not_dones that are from other transitions are filtered out.
+        filt = np.hstack([np.ones((len(batch_inds), 1)), not_dones[:, 1:]]) * (indices != self.pos)
```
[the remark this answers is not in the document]
araffin: oh... you also included the not dones in there

"not_dones that are from other transitions are filtered out." -> not super clear to me....

and still not sure what the filter should look like...

stable_baselines3/common/buffers.py
```diff
@@ -387,15 +380,81 @@ def get(self, batch_size: Optional[int] = None) -> Generator[RolloutBufferSample
...
+        filt = np.multiply.accumulate(filt, 1).reshape(filt.shape)
+
+        # weighted rewards
+        rewards = filt * gammas * rewards
+        rewards = np.add.reduce(rewards, 1).reshape(len(batch_inds), 1).astype(np.float32)
+
+        increments = np.add.reduce(filt, 1).astype(np.int).reshape(batch_inds.shape)
```
araffin, line 449: missing comment

stable_baselines3/common/buffers.py
```diff
@@ -387,15 +380,81 @@ def get(self, batch_size: Optional[int] = None) -> Generator[RolloutBufferSample
...
+        next_obs_indices = (increments + batch_inds - 1) % self.buffer_size
```
araffin, line 450: I will stop here today. Overall looks ok with lot of nice numpy tricks ;)
I will try to do the rest next week.

my checking code:

```python
import numpy as np

buffer_size = 5
batch_inds = np.arange(5)
batch_inds[0] = 1
n_step = 2
gamma = 0.99
dones = np.zeros(5)
dones[3] = 1
rewards = np.ones(5) * 2
rewards[1] = 1
# Buffer pointer
pos = 0

print("rewards\n", rewards)
print("dones\n", dones)
print("buffer_pos=", pos)
print("batch_inds\n", batch_inds)

indices = np.arange(n_step) + batch_inds.reshape(-1, 1) 
indices = indices % buffer_size
print("indices\n", indices)

not_dones = np.multiply.accumulate(1 - dones[indices], 1).reshape(-1, n_step)
print("not_dones\n", not_dones)

gammas = gamma ** np.arange(n_step)
print("gammas\n", gammas)


rewards = rewards[indices].reshape(not_dones.shape)

print("rewards\n", rewards)

filt = np.hstack([np.ones((len(batch_inds), 1)), not_dones[:, 1:]]) * (indices != pos)

print("filter\n", filt)

filt = np.multiply.accumulate(filt, 1).reshape(filt.shape)

print("filter2\n", filt)

rewards = filt * gammas * rewards

print("weighted reward\n", rewards)

rewards = np.add.reduce(rewards, 1).reshape(len(batch_inds), 1).astype(np.float32)

print("reduced rewards\n", rewards)

increments = np.add.reduce(filt, 1).astype(np.int).reshape(batch_inds.shape)

print("increments\n", increments)

next_obs_indices = (increments + batch_inds - 1) % buffer_size

print("next_obs_indices\n", next_obs_indices)
```

Miffyli:
On a quick glimpse it looks good, like arrafin already mentioned, but I would be even more exhaustive with the comments. The numpy tricks are compact and can be performant, but the code becomes harder to understand. I would also include the shape of arrays in the comments with a definition of contents, e.g. `# (batch_size, num_actions) of {0,1}`.

stable_baselines3/common/buffers.py
```diff
@@ -387,15 +380,81 @@ def get(self, batch_size: Optional[int] = None) -> Generator[RolloutBufferSample
...
+        n_envs: int = 1,
+        optimize_memory_usage: bool = False,
+        n_step: int = 1,
+        gamma: float = 0.99,
+    ):
+        super().__init__(buffer_size, observation_space, action_space, device, n_envs, optimize_memory_usage)
+        self.n_step = int(n_step)
+        assert self.n_step <= buffer_size
+        self.gamma = float(gamma)
+        assert 0 <= gamma <= 1
```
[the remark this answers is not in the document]
PartiallyTyped: I forgot those, that and the above were for testing because I didn't integrate them at the algorithms at the time. I will remove them ^^'.

stable_baselines3/common/buffers.py
```diff
@@ -387,15 +380,81 @@ def get(self, batch_size: Optional[int] = None) -> Generator[RolloutBufferSample
...
+
+    def _get_samples(self, batch_inds: np.ndarray, env: Optional[VecNormalize] = None) -> ReplayBufferSamples:
+        assert not np.any(batch_inds == self.pos)
+        actions = self.actions[batch_inds, 0, :]
+
+        gamma = self.gamma
+
+        # Broadcasting turns 1dim arange matrix to 2 dimensional matrix that contains all
+        # the indices, % buffersize keeps us in buffer range
+        indices = (np.arange(self.n_step) + batch_inds.reshape(-1, 1)) % self.buffer_size
```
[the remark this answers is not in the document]
PartiallyTyped: Thanks! I there's a minor bug here. It should be `self.size()`.

stable_baselines3/common/buffers.py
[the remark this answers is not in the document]
PartiallyTyped: The idea is that the filter selects the relevant rewards. The filter always has col 0 to be 1, because it is the immediate reward and arange at col 0 is 0, so that is okay.

The not dones does the same, except for steps 1, 2, ... n-1, so it chooses which rewards to take into consideration. When done =1, `multiply.accumulate` zeros out the future rewards.

> "not_dones that are from other transitions are filtered out." -> not super clear to me....

What I meant to say was that `self.pos` points to either older or empty transitions. We need to account for this through the filter. So using `indices != pos` ensures that we ignore the empty or older transition. 

Did I clear things a bit?

stable_baselines3/common/buffers.py
PartiallyTyped:
```
>>> import numpy as np
>>>
>>> buffer_size = 5
>>> batch_inds = np.arange(5)
>>> batch_inds[0] = 1
>>> n_step = 2
>>> gamma = 0.99
>>> dones = np.zeros(5)
>>> dones[3] = 1
>>> rewards = np.ones(5) * 2
>>> rewards[1] = 1
>>> # Buffer pointer
... pos = 0
>>>
>>> print("rewards\n", rewards)
rewards
 [2. 1. 2. 2. 2.]
>>> print("dones\n", dones)
dones
 [0. 0. 0. 1. 0.]
>>> print("buffer_pos=", pos)
buffer_pos= 0
>>> print("batch_inds\n", batch_inds)
batch_inds
 [1 1 2 3 4]
>>> indices = np.arange(n_step) + batch_inds.reshape(-1, 1)
>>> indices = indices % buffer_size
>>> print("indices\n", indices)
indices
 [[1 2]
 [1 2]
 [2 3]
 [3 4]
 [4 0]]
```
We can agree that all of these are correct :)
So `not_dones` performs a forward filtering on the future states to consider.
It is expected that when the current transition is terminal, the whole row will be 0.
This is accounted for later.
```
>>> not_dones = np.multiply.accumulate(1 - dones[indices], 1).reshape(-1, n_step)
>>> print("not_dones\n", not_dones)
not_dones
 [[1. 1.]
 [1. 1.]
 [1. 0.]
 [0. 0.]
 [1. 1.]]
```
As  expected the 4th row is 0.

```
>>> gammas = gamma ** np.arange(n_step)
>>> print("gammas\n", gammas)
gammas
 [1.   0.99]
>>> rewards = rewards[indices].reshape(not_dones.shape)
>>> print("rewards\n", rewards)
rewards
 [[1. 2.]
 [1. 2.]
 [2. 2.]
 [2. 2.]
 [2. 2.]]
```
Nothing to add above.

This is the interesting part.
The first thing we do here is to correct the `not_dones` in the first column to be equal to 1, because this is the immediate reward. Because of `multiply.accumulate`, `not_dones` will have 0s for transitions from other episodes.
```
>>> filt = np.hstack([np.ones((len(batch_inds), 1)), not_dones[:, 1:]]) * (indices != pos)
>>> print("filter\n", filt)
filter
 [[1. 1.]
 [1. 1.]
 [1. 0.]
 [1. 0.]
 [1. 0.]]
>>> filt = np.multiply.accumulate(filt, 1).reshape(filt.shape)
>>>
>>> print("filter2\n", filt)
filter2
 [[1. 1.]
 [1. 1.]
 [1. 0.]
 [1. 0.]
 [1. 0.]]
>>>
```
There is a bug here, third row, second column should be one. I will fix it and add it to the test cases.
Last row is correct, `pos==0`, so we shouldn't use that transition.

```
>>> rewards = filt * gammas * rewards
>>>
>>> print("weighted reward\n", rewards)
weighted reward
 [[1.   1.98]
 [1.   1.98]
 [2.   0.  ]
 [2.   0.  ]
 [2.   0.  ]]
>>>
>>> rewards = np.add.reduce(rewards, 1).reshape(len(batch_inds), 1).astype(np.float32)
>>>
>>> print("reduced rewards\n", rewards)
reduced rewards
 [[2.98]
 [2.98]
 [2.  ]
 [2.  ]
 [2.  ]]
```
Yup, third row should have had a reward of 3.98.
```
>>>
>>> increments = np.add.reduce(filt, 1).astype(np.int).reshape(batch_inds.shape)
>>>
>>> print("increments\n", increments)
increments
 [2 2 1 1 1]
```
Same bug as above, increments[2] should have been 2.
```
>>>
>>> next_obs_indices = (increments + batch_inds - 1) % buffer_size
>>>
>>> print("next_obs_indices\n", next_obs_indices)
next_obs_indices
 [2 2 2 3 4]
```
Same bug as above.

stable_baselines3/common/buffers.py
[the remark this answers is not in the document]
PartiallyTyped: Nvm, it was correct ... ^^'.

stable_baselines3/common/buffers.py
PartiallyTyped: This is solved now.

stable_baselines3/common/buffers.py
```diff
@@ -381,3 +381,100 @@ def _get_samples(self, batch_inds: np.ndarray, env: Optional[VecNormalize] = Non
...
+        gamma: float = 0.99,
+    ):
+        super().__init__(buffer_size, observation_space, action_space, device, n_envs, optimize_memory_usage)
+        self.n_step = int(n_step)
+        if not 0 < n_step <= buffer_size:
+            raise ValueError("n_step needs to be strictly smaller than buffer_size, and strictly larger than 0")
+        self.gamma = gamma
+
+    def _get_samples(self, batch_inds: np.ndarray, env: Optional[VecNormalize] = None) -> ReplayBufferSamples:
+        assert not np.any(batch_inds == self.pos)
```
araffin, line 423: why?
Only in the optimize memory case, no?

PartiallyTyped: I wanted to make this function accessible without using `self.sample()`.

In `self.sample`, the following prevents `self.pos` from being accessible:

```Python
  if self.full:
            batch_inds = (np.random.randint(1, self.buffer_size, size=batch_size) + self.pos) % self.buffer_size
        else:
            batch_inds = np.random.randint(0, self.pos, size=batch_size)
```

This was done to ensure we can use the same sampling mechanism for both optimized and unoptimized buffers.

araffin:
I'm currently given a quick try to that one (`feat/n-steps` in the zoo), and it already yields some interesting results with DQN on CartPole ;)
And I couldn't notice any slow down.

The command:
```
python train.py --algo dqn --env CartPole-v1 -params replay_buffer_class:NstepReplayBuffer replay_buffer_kwargs:"dict(n_step=4)" --num-threads 2 -tb /tmp/sb3
```

The two first curves are with `n_step=4` (much more stable learning) and the two others are with `n_step=1`

dqn_n_step


I will try with SAC soon.

stable_baselines3/common/buffers.py
araffin: Hmm,
I had to comment that line because it was giving me an error at some point in the training.

>I wanted to make this function accessible without using self.sample().

Why?

PartiallyTyped: It seemed right at the time, but obviously it isn't necessary.

PartiallyTyped:
Looks good! What about the FPS? It should have very small impact, but, there are still some optimizations that can be done.

araffin:
>Looks good! What about the FPS? It should have very small impact, but, there are still some optimizations that can be made.

On DQN with CartPole, as mentioned, I couldn't notice the difference (all runs where around 1000FPS).
For SAC, experiments are currently running but there is no difference for now =)

PartiallyTyped:
For sac, I am not sure if n-steps can be applied directly as I am under the impression that the backup requires the entropy for the intermediate states as well, not just the last one.

araffin:
>For sac, I am not sure if n-steps can be applied directly as I am under the impression that the backup requires the entropy for the intermediate states as well, not just the last one.

Good point, for now I'm just checking quickly assuming the entropy term is neglectable compared to the reward term.
I could see a positive impact on LunarLanderContinuous but not on PyBullet envs yet (probably because the reward is quite shaped). 

The 3 runs with mean reward > 200 are the one with the n-step replay buffer:

sac_lunar

araffin:
I added a sketch of how it would look like for SAC, it fits in ~10 lines. We would need to allocate one more array `log_prob` of size `buffer_size` and pass the actor to the buffer.
I also did a small pre-processing step to remove duplicated indices when predicting the `log_prob`.

PartiallyTyped:
If you have plans for other n_step methods, (I'd love to have retrace), then, we can sketch out how those would work (most of which require action probabilities) and figure out how to incorporate sac there as well, I was thinking of something along the lines of adding an entropy coefficient like so:

```python
# compute rewards, entropy and other weighting fractions onto rewards,
rewards = (rewards + entr_coef * entropy) @ weighting
```

araffin:
>'d love to have retrace

I would need to look at that again. but yes, it would be nice to have an eligibility trace estimate style (i.e. TD(lambda) style).

The weighting is already there with the gammas, no?

>we would have the entropy for (almost) free

?

For me the main issue is how to compute action log proba in a clean way? (passing the actor to the sample method, as we do when we normalize the rewards seems the cleanest for now)

araffin:
Btw, n-step does not help for now on PyBullet envs (I think to think why, maybe some tuning is needed) but it improves results and stability on env like LunarLander or BipedalWalkerHardcore (I could finally get the maximum performance with SAC).

PartiallyTyped:
> The weighting is already there with the gammas, no?

In retrace(lambda), treebackup(lambda), Q(lambda), etc, the weighting also uses some other factor besides gamma, thus I added it in a single term.

> > we would have the entropy for (almost) free
> 
> ?

To compute Retrace, TreeBackup and Importance Sampling, we need the action probabilities. If we have the action probabilities, we also have the entropy (for free), thus if we implement any of the above, we will also have a clear path towards implementing n-steps with entropy.

> For me the main issue is how to compute action log proba in a clean way? (passing the actor to the sample method, as we do when we normalize the rewards seems the cleanest for now)

I was also thinking of passing a function instead of the actor directly but the two are more or less the same.

araffin:
>In retrace(lambda), treebackup(lambda), Q(lambda), etc, the weighting also uses some other factor besides gamma, thus I added it in a single term.

yes, I meant that gamma played already the weighting role.

>I was also thinking of passing a function instead of the actor 

sounds better, yes ;)

PartiallyTyped:
Yeah makes sense, I assume the function will be of the form `np.ndarray -> np.ndarray`, so as to make it agnostic of the device.

araffin:
>Yeah makes sense, I assume the function will be of the form np.ndarray -> np.ndarray, so as to make it agnostic of the device.

Good idea

stable_baselines3/common/buffers.py
```diff
@@ -381,3 +381,118 @@ def _get_samples(self, batch_inds: np.ndarray, env: Optional[VecNormalize] = Non
             self.returns[batch_inds].flatten(),
         )
         return RolloutBufferSamples(*tuple(map(self.to_torch, data)))
+
+
+class NstepReplayBuffer(ReplayBuffer):
```
araffin, line 386: Minor: why not `NStepReplayBuffer`?

stable_baselines3/common/buffers.py
```diff
@@ -381,3 +381,118 @@ def _get_samples(self, batch_inds: np.ndarray, env: Optional[VecNormalize] = Non
...
+        # using accumulate. This ensures that we don't use invalid transitions
+        # not_dones is a [B x n_step] matrix
+        not_dones = np.squeeze(np.multiply.accumulate(1 - self.dones[indices], 1), -1)
+        # vector of the discount factors
+        # [n_step] vector
+        gammas = gamma ** np.arange(self.n_step)
+
+        # two dim matrix of rewards for the indices
+        # using indices we select the current transition, plus the next n_step ones
+        rewards = np.squeeze(self.rewards[indices], -1)
```
araffin, line 445: minor: please always specify explicitly `axis=` so we know what is that parameter

stable_baselines3/common/off_policy_algorithm.py
```diff
@@ -67,6 +67,7 @@ class OffPolicyAlgorithm(BaseAlgorithm):
     :param use_sde_at_warmup: (bool) Whether to use gSDE instead of uniform sampling
         during the warm up phase (before learning starts)
     :param sde_support: (bool) Whether the model support gSDE or not
+    :param n_step: (int) Number of steps to consider when computing the target Q-value
```
araffin, line 70: minor (mostly my fault): docstring missing for the replay buffer class and kwargs

stable_baselines3/common/buffers.py
```diff
@@ -381,3 +381,118 @@ def _get_samples(self, batch_inds: np.ndarray, env: Optional[VecNormalize] = Non
...
+
+    def __init__(
+        self,
+        buffer_size: int,
+        observation_space: spaces.Space,
+        action_space: spaces.Space,
+        device: Union[th.device, str] = "cpu",
+        n_envs: int = 1,
+        optimize_memory_usage: bool = False,
+        n_step: int = 1,
```
araffin, line 413: why not `n_steps` as for A2C/PPO?

stable_baselines3/common/buffers.py
PartiallyTyped: I used it as a single word. I will update it.

PartiallyTyped:
So how should we continue from here? Since n_steps was planned for v1.1, I can take the time to add extensions and draft a more generic version of this that also includes entropy.

araffin:
Yes, sounds good ;)
I would wait for support of all algorithms anyway to avoid inconsistencies.
But if it is ready and fully tested, we can merge it with master.

PartiallyTyped:
I don't have anything to add to NStepReplay, or any other tests. So feel free to merge this and I can start drafting the generic version.

araffin:
>I don't have anything to add to NStepReplay, or any other tests.

So the entropy is not part of it?
Maybe add an example in the doc.

PartiallyTyped:
I’d rather have a separate implementation that includes entropy but it’s up to you.

Yes I will add an example in the docs.

> I don't have anything to add to NStepReplay, or any other tests.
> 
> So the entropy is not part of it?
> Maybe add an example in the doc.
>

araffin:
>I’d rather have a separate implementation that includes entropy but it’s up to you.

Hmm, the issue is that if we merge it now, it will be half-working only (working for DQN, TD3 but not for SAC).

>Yes I will add an example in the docs.

yes, please =)

PartiallyTyped:
Fair point! I have a couple of weeks before grad classes pick up the pace so hopefully I will have everything tied up before then.

PartiallyTyped:
I am considering the following changes:

1. Use a component called say "batch generator" that given the indices produces the batch.
2. Instead of subclassing the buffer class for the different buffers, just pass `buffer_kwargs`, the constructor of the buffer will then create the correct component.

This removes the need to create all buffer combinations, e.g. if we were to have prioritized replay, we wouldn't need to create `n_step_prioritized_replay`.

What do you think?

PartiallyTyped:
@araffin 
I am considering the following changes:

1. Use a component called say "batch generator" that given the indices produces the batch.
2. Instead of subclassing the buffer class for the different buffers, just pass `buffer_kwargs`, the constructor of the buffer will then create the correct component.

This removes the need to create all buffer combinations, e.g. if we were to have prioritized replay, we wouldn't need to create `n_step_prioritized_replay`.

What do you think?

araffin:
>Use a component called say "batch generator" that given the indices produces the batch.

How is that different from `_get_samples()` ?

>Instead of subclassing the buffer class for the different buffers, just pass buffer_kwargs, the constructor of the buffer will then create the correct component.

Not sure what you mean, it sounds like a `_get_samples()`  with a lot of branching?

PartiallyTyped:
> How is that different from `_get_samples()` ?

`_get_samples()` is bound to the class of the buffer. I would like to use composition instead of sub-classing the buffer to create new `_get_samples()`.

very rough example

```python

class ReplayBuffer(BaseBuffer):
     def __init__(self, ..., batch_method, **batch_kwargs):
          self._sampler = Sampler.make(batch_method, **batch_kwargs) # Will create the correct sampler depending on args
          # make is a class method.
```

The consequence here is that we can do:

```python
env = ...
agent_classes = [TD3, SAC] # SAC will add `use_entropy=True` in `buffer_kwargs` unless specified otherwise
for cls in agent_classes:
    agent = cls(env, buffer_kwargs={'batch_method':'nsteps', 'nsteps':5}) # all these can be trivially logged because they are strings and ints
```

The pros are:

 - We remove the need to provide the buffer class
 - The sampler can be swapped on the fly
 - We don't need to create multiple classes for buffers that just combine other classes, e.g. for Nstep+PER, we will not have to add `NStepPer`, and `EntropyNStepPer`, for Retrace(lambda) we will not have to add `RetracePer`, `EntropicRetracePer`
 - We can make it trivial to create new sampler classes and register them.
 - SAC will not actually require a new buffer class for itself that would need to be specified in its arguments to ensure correctness.
 - The number of classes is linear to the number of sampler methods.

The cons are:
 - More types
 - Need to write a function to construct the correct sampler (fairly easy)

araffin:
>I would like to use composition instead of sub-classing the buffer to create new _get_samples().

I see, the idea would be to create a class in charge of sampling instead of putting everything in the `get_sample()` method. (sorry for the delay, was on holidays)

sounds reasonable, @Miffyli or @AdamGleave do you have any comment on that?

Miffyli:
Just to reiterate: The storage would just store the samples and possibly priorities, and sampler reads samples from the storage and computes the desired return (i.e. single step, n-step, retrace, etc etc)? This sounds good to me if they can be separated cleanly, as then other samplers can be added without changing the storage code (as mentioned by PartiallyTyped). I am mainly worried if there could be some sort of memory/performance improvement if the two are combined, but I still prefer separating the two if possible. 

One thing: It should be called something else than "sampler" if the storage defines how samples are sampled, but this "sampler" just computes the returns and returns correct successor states.

PartiallyTyped:
To address the issue of overhead, it would be minimal, perhaps an additional function call or attribute lookup.

AdamGleave:
It seems reasonable to separate these. I second @Miffyli's confusion regarding the name "sampler" for something that computes returns. Perhaps "buffer" and "postprocessor" (instead of "sampler").

geyang:
For n-step, testing it on DMC Quadruped would be good, because n-step is expected to show some gain there.

> >  I’d rather have a separate implementation that includes entropy but it’s up to you.
> Hmm, the issue is that if we merge it now, it will be half-working only (working for DQN, TD3 but not for SAC).

Much better to merge first and break things, rather than stalling.

PartiallyTyped:
Hey @araffin, I have been quite busy with grad school so the branch's become stale.
What would be the minimum requirements to merge this?

araffin:
> Hey @araffin, I have been quite busy with grad school so the branch's become stale. What would be the minimum requirements to merge this?

Hey,

> I have been quite busy with grad school so the branch's become stale.

no worry.

let me try to go back to you this week if I have time.
In the meantime, I tested a more general version in contrib: https://github.com/Stable-Baselines-Team/stable-baselines3-contrib/tree/feat/peng-q-lambda
but haven't had time to benchmark it properly.

>  What would be the minimum requirements to merge this?

now, I'm thinking the place should be better in SB3-contrib (what do you think @Miffyli ?)
and anyway, some benchmarking would be needed + enough testing before merging it.
The first thing you need to do is solve the merge conflicts I think... the biggest difference is the handling of timelimit now.

emrul:
Just out of curiosity I tried this buffer implementation and I think it does not handle multiple environments correctly. Specifically the line:
```
not_dones = np.squeeze(1 - self.dones[indices], axis=-1)
```
causes a `ValueError: cannot select an axis to squeeze out which has size not equal to one` error.

As an aside, any please don't take this as criticism of this project which I enjoy using but its quite sad to see potentially really great contributions/PRs sitting open for nearing 3 years. In that time perhaps a lot of people could have benefitted - even if this was rushed into the contrib project it would have been more useful than as-is.

emrul:
I'm still a bit new to numpy and pytorch manipulation but keen to help with little things if you need (e.g. docs, etc.)

richardjozsa:
what is missing from this PR to be complete?

emrul:
I'm not sure to be honest, I couldn't get it to work (with vectorised environments as previously mentioned) and wasn't sure how to go about fixing it.

PR #89, opened 2020-07-03
Refactor and clean-up of common code

AdamGleave:
## Description
Overall the code looked good -- I was only tempted to make a few fairly minor changes amongst the couple of thousand lines of code.

Changes from reviewing common code as per #17. Specific changes:
  - Add module docstrings
  - `base_class.py`:
    + Refactor some environment creation logic into `maybe_make_env`
    + Remove `raise NotImplementedError()` on abstract methods. This should not be needed as Python will refuse to instantiate  classes with abstract methods.
    + Removed check for "_setup_model" existing: `hasattr(base_class.BaseAlgorithm, '_setup_model')` is `True` so I think this check cannot ever fire.
  - `distributions.py`:
    + Make `Distributions` an ABC, use `@abstractmethod` instead of `NotImplementedError`
    + Have subclasses methods defined in consistent order

Issues I noticed but wasn't sure how to resolve (tick indicates resolved subsequently):
  - [x] In `BaseAlgorithm.load`, we accept `data` missing observation and action space, provided it contains the `"env"` key. But if the user provides an `env` argument, then we check against the (possibly non-existent) observation/action spaces. Can we just always require that data contains observation & action spaces?
  - [x] In `BaseAlgorithm._setup_learn` I think we may have a similar issue to that reported in https://github.com/hill-a/stable-baselines/issues/619 and (partially) fixed in https://github.com/hill-a/stable-baselines/pull/649 In brief, repeated calls to `learn()` each call `_setup_learn()` which resets the `ep_info_buffer` and `ep_success_buffer`. Perhaps we should only reset these if `reset_num_timesteps` is True?
  - [x] In `Distribution`, `proba_distribution` and `proba_distribution_net` are not included, but every subclass defines these methods. Should we add them to the ABC?
  - [x] In `BasePolicy.predict`, the code for `state` and `mask` are commented out. Not sure what's going on here -- do we just not support RNN policies yet? If so, should warn of this in the docstring, and probably open an enhancement issue to resolve this.
  - [x] In `BasePolicy._get_data`, two lines are commented out for `squash_output` and `features_extractor` -- I'm not sure if this is desired or not?
  - [x] In `ActorCriticPolicy._build`, `self.action_net` may not get constructed if we don't recognize the distribution. Is this intentional (e.g. responsibility of subclass to build it) or should we add an `else: raise NotImplementedError()`? I note we do raise `NotImplementedError` in `ActorCriticPolicy._get_action_dist_from_latent`.
  - [x] The `Critic` in SAC and TD3 subclassing `BasePolicy` feels a little odd, e.g. they do not implement `_predict`. However, I'm not sure it's actually worth refactoring.

## Motivation and Context

## Types of changes

## Checklist:

- [X] I have checked the codestyle using `make lint`
- [X] I have ensured `make pytest` and `make type` both pass.


stable_baselines3/common/base_class.py
```diff
@@ -23,12 +25,30 @@
 from stable_baselines3.common.noise import ActionNoise
 
 
+def maybe_make_env(env: Union[GymEnv, str, None], monitor_wrapper: bool, verbose: int) -> Optional[GymEnv]:
```
araffin, line 28: I don't know if it should be there, as a static method or in `utils.py`...
because this is only used by the `BaseAlgorithm` class.

stable_baselines3/common/base_class.py
```diff
@@ -366,14 +370,18 @@ def set_random_seed(self, seed: Optional[int] = None) -> None:
             self.eval_env.seed(seed)
 
     def _init_callback(self,
-                       callback: Union[None, Callable, List[BaseCallback], BaseCallback],
+                       callback: MaybeCallback,
```
araffin, line 373: nice catch, I forgot that I defined that alias ^^

stable_baselines3/common/base_class.py
```diff
@@ -492,29 +500,27 @@ def save(
...
-        # use standard list of excluded parameters if none given
+
+        # Exclude is union of specified parameters (if any) and standard exclusions
         if exclude is None:
-            exclude = self.excluded_save_params()
-        else:
-            # append standard exclude params to the given params
-            exclude.extend([param for param in self.excluded_save_params() if param not in exclude])
+            exclude = []
+        exclude = set(exclude).union(self.excluded_save_params())
```
araffin, line 507: nice trick to use sets instead of lists =)

stable_baselines3/common/base_class.py
AdamGleave: If only used by `BaseAlgorithm` then probably belongs somewhere in this file.

I'd be OK with module-level function or static method. I've got in the habit of avoiding `@staticmethod` after being [brainwashed by Google](https://google.github.io/styleguide/pyguide.html#2174-decision) but I don't think their argument is that strong.

araffin:
>Can we just always require that data contains observation & action spaces?

Sounds reasonable ;) (should be already the case in fact)

>Perhaps we should only reset these if reset_num_timesteps is True?

yep. I tried not to reproduced the errors from SB2 but I may have missed that one.

>In Distribution, proba_distribution and proba_distribution_net are not included, but every subclass defines these methods. Should we add them to the ABC?

Good point. I forgot to add them after refactoring the distributions.

> Not sure what's going on here -- do we just not support RNN policies yet? If so, should warn of this in the docstring, and probably open an enhancement issue to resolve this.

No, we do not support RNN yet, this is in the roadmap for v1.1+ though. I agree with opening an issue to keep track of it ;)

>two lines are commented out for squash_output and features_extractor -- I'm not sure if this is desired or not?

yes, this is desired (see comment above). I had issue in the past because of inheritance.

>self.action_net may not get constructed if we don't recognize the distribution. Is this intentional (e.g. responsibility of subclass to build it) or should we add an else: raise NotImplementedError()?

An error will be thrown earlier in fact: `self.action_dist = make_proba_distribution(action_space, use_sde=use_sde, dist_kwargs=dist_kwargs)`.

>he Critic in SAC and TD3 subclassing BasePolicy feels a little odd, e.g. they do not implement _predict. However, I'm not sure it's actually worth refactoring.

I agree with that feeling... Maybe a new `BaseModel` and `BasePolicy` would derive from it?

Issue #87, opened 2020-07-02
[Enhancement] Improve Typing Support / Ensure every function has return type


PartiallyTyped:
One of the goals for v1 release was to improve typing support. This issue is only concerned return types.
I wanted to improve the type support so I wrote a script to parse python code and filter out function definitions and class method definitions to find out what needs typing. In the process, I figured that this would be quite useful to add as a test.

## Suggestion

Add a test to ensure that every function has a return type.

This isn't to be confused with type verification, e.g. [bad return types](https://google.github.io/pytype/errors.html#bad-return-type).

## Special Cases

In some cases we may want to avoid adding a return type, a decorator could be added onto the function to make it ignored by the test.

## An exhaustive list of functions that are missing return types:

```
Missing types:
stable_baselines3/common/vec_env/dummy_vec_env.py::DummyVecEnv::step_async
stable_baselines3/common/vec_env/dummy_vec_env.py::DummyVecEnv::step_wait
stable_baselines3/common/vec_env/dummy_vec_env.py::DummyVecEnv::seed
stable_baselines3/common/vec_env/dummy_vec_env.py::DummyVecEnv::reset
stable_baselines3/common/vec_env/dummy_vec_env.py::DummyVecEnv::close
stable_baselines3/common/vec_env/dummy_vec_env.py::DummyVecEnv::render
stable_baselines3/common/vec_env/dummy_vec_env.py::DummyVecEnv::_save_obs
stable_baselines3/common/vec_env/dummy_vec_env.py::DummyVecEnv::_obs_from_buf
stable_baselines3/common/vec_env/dummy_vec_env.py::DummyVecEnv::get_attr
stable_baselines3/common/vec_env/dummy_vec_env.py::DummyVecEnv::set_attr
stable_baselines3/common/vec_env/dummy_vec_env.py::DummyVecEnv::env_method
stable_baselines3/common/vec_env/dummy_vec_env.py::DummyVecEnv::_get_target_envs
stable_baselines3/common/logger.py::SeqWriter::write_sequence
stable_baselines3/common/vec_env/vec_normalize.py::VecNormalize::__getstate__
stable_baselines3/common/vec_env/vec_normalize.py::VecNormalize::__setstate__
stable_baselines3/common/vec_env/vec_normalize.py::VecNormalize::set_venv
stable_baselines3/common/vec_env/vec_normalize.py::VecNormalize::step_wait
[...]

PR #136, opened 2020-08-04
Fix double reset and improve typing coverage

### LINKED ISSUE #60: Environment is reset twice per episode when evaluating policy on DummyVecEnv
The `evaluate_policy` helper function reset the environment at the start of each episode:

https://github.com/DLR-RM/stable-baselines3/blob/494ebfd20abe90acc136fdaf215c76ec566acd2c/stable_baselines3/common/evaluation.py#L33-L34

But `DummyVecEnv` automatically resets the environment when `step` returns `done = true`: 

https://github.com/DLR-RM/stable-baselines3/blob/494ebfd20abe90acc136fdaf215c76ec566acd2c/stable_baselines3/common/vec_env/dummy_vec_env.py#L45-L48

This causes the environment to reset twice per episode when evaluating the policy.

araffin:
## Description

closes #60 

addressed part of #87 

## Motivation and Context

## Types of changes

## Checklist:


stable_baselines3/common/vec_env/base_vec_env.py
```diff
@@ -82,12 +66,12 @@ def reset(self):
...
         until step_async() is invoked again.
 
-        :return: ([int] or [float]) observation
+        :return: (VecEnvObs) observation
         """
         raise NotImplementedError()
 
     @abstractmethod
-    def step_async(self, actions):
+    def step_async(self, actions: np.ndarray):
```
Miffyli, line 74: Hmm this might be reverted when we start implementing support for dict/tuple action-spaces, where we can not stack actions into a numpy array. This feature is considered to be added right after v1.0, being so well requested. Same applies to observations.

stable_baselines3/common/vec_env/base_vec_env.py
```diff
@@ -1,12 +1,18 @@
...
+from typing import Any, Dict, Iterable, List, Optional, Sequence, Tuple, Union
 
 import cloudpickle
+import gym
 import numpy as np
 
 from stable_baselines3.common import logger
 
+# Avoid circular import
+VecEnvIndices = Union[None, int, Iterable[int]]
```
Miffyli, line 12: These could use bit of docs on what they will contain (_e.g._ VecEnvStepReturn contains obs, rewards, dones and infos).

stable_baselines3/common/vec_env/base_vec_env.py
araffin: I'm not so confident about dict/tuple support for action spaces, sounds like an open research question to me (mixing proba distributions)

for observations, this is already included in the type ;)

Miffyli: Actions in those cases are usually treated independent, as they are independent in the environment specifications, but indeed it is not super clear how it should be done. Lets go with the current modifications and deal with that later.

PR #138, opened 2020-08-10
Get/set parameters and review of saving and loading

### LINKED ISSUE #70: transferrable models
Hi, is there a way to save a model only with the internal layers of the NN? Or is it possible to modify a trained model (the zip file), to let it retrain with different action dimension?

I am thinking the following example, say we have env_1 with action dimension 10, and env_2, which is more complicated version of env_1, with  action dimension 20. One model is trained with env_1. Can we modify the model (zip file) and use it as initial value to train on env_2 afterwards?

Thank you very much!

### LINKED ISSUE #116: [Feature Request] Allow to pass `strict=False` when loading model
In SB2, for `load_parameters`, we had a `exact_match` parameter to allow key mismatch when loading.
It would be nice to have that, we have to exclude `optimizers` state dict though.
This feature allows for instance to load models with same architecture but different names (recently there was a rename with #92 that broke previously saved policy, with `strict=False` we can still load the old models that are still valid). 

We also had a `custom_objects` parameter to replace parameters on the fly, not sure if it is still needed... @Miffyli could you comment on that?

Miffyli:
Closes #116

closes #70 

Review over saving and loading of models, as well as (possibly) implementing `get_parameters` and `set_parameters` akin to stable-baselines2.

## Changelog
- Rename BaseClass `get_torch_variables` -> `_get_torch_save_params`, and include docstring only in original implementation.
- Rename BaseClass `excluded_save_params` -> `_excluded_save_params`, and include docstring only in original implementation.
- Reorganize functions for clarity (save/load functions are closer to each other, private functions first in classes)
- Clarify documentation on different objects stored (data, params and tensors)
- Rename saved items `tensors` to `pytorch_variables` for clarity.
- Simplify `save_to_zip_file` by combining duplicate code.
- Add `get/set_parameters`, which use `_get_torch_save_params` to gather/set parameters of different objects.

## Motivation and Context

## Types of changes

- [X] New feature (non-breaking change which adds functionality)
- [X] Documentation (update in the documentation)

## Suggestions and TODOs
  - [x] Document the use of recursive_set/getattr in e.g. `get_torch_variables` 
  - [x] Clarify documentation of "excluded_save_params". Main purpose is to avoid storing pytorch stuff with standard pickle.  
  - [x] Merge (if possible) "tensors" with "data" or "params", or at least document well the differences between the three. Merging would simplify the code all-around.
  - [x] Consider renaming "tensors" to e.g. "torch_variables" or "extra_data". Now it sounds like these could be the PyTorch parameters generally. 
  - [x] Remove "cf base class" docstrings and replace with the full docstring (or remove completely).
  - [x] Rename "get_torch_variables" function to something more private. It sounds like it returns the agent parameters, but in reality it returns library-specific stuff.
  - [x] Add `get/set_parameters`.

## Things to think about

- Since PyTorch stores variables with pickle, and since v1.6 they store them as zip-file with pickles, the benefits of zip-files here disappears. It gets even bit silly, since we have a zip-file of zip-files of pickles for parameters. Maybe consider storing the parameters with `np.savez` as with TF?
  - Update: Leave this for later.

## Checklist:


araffin:
>Document the use of recursive_set/getattr in e.g. get_torch_variables

>Clarify documentation of "excluded_save_params". Main purpose is to avoid storing pytorch stuff with standard pickle.

Main purpose is to avoid saving things that cannot or should not be pickled (e.g. the environment, the replay buffer, pytorch variables, ...).

>Merge (if possible) "tensors" with "data" or "params", or at least document well the differences between the three. Merging would simplify the code all-around.

would be nice to reduce the number of variables but not sure if possible. For info, that's mainly @Artemis-Skade who worked on the saving/loading part. We also had some trouble with saving on GPU, loading on cpu...

>Consider renaming "tensors" to e.g. "torch_variables" or "extra_data". Now it sounds like these could be the PyTorch parameters generally.

Yes, anything that can be saved with `th.save` which is not in the policy. I think for now, this is only the entropy temperature in SAC.

>Remove "cf base class" docstrings and replace with the full docstring (or remove completely).

Yep, not sure about that, you choose ;)

> Rename "get_torch_variables" function to something more private. It sounds like it returns the agent parameters, but in reality it returns library-specific stuff.

agree ;)

>Since PyTorch stores variables with pickle, and since v1.6 they store them as zip-file with pickles, the benefits of zip-files here disappears. It gets even bit silly, since we have a zip-file of zip-files of pickles for parameters. Maybe consider storing the parameters with np.savez as with TF?

I did not know that... but that would mean breaking changes for pytorch users of versions 1.4.x and 1.5.x
Also, I don't mind to have zip of zip.

araffin:
For get/set parameters, I think we maybe don't need them as we can access `model.policy` easily, and therefore have access to `model.policy.named_parameters()`, `model.policy.state_dict()` and `model.policy.load_state_dict()`.

stable_baselines3/common/base_class.py
```diff
@@ -366,10 +383,10 @@ def load(cls, load_path: str, env: Optional[GymEnv] = None, **kwargs) -> "BaseAl
             attr = recursive_getattr(model, name)
             attr.load_state_dict(params[name])
 
-        # put tensors back in place
-        if tensors is not None:
-            for name in tensors:
-                recursive_setattr(model, name, tensors[name])
+        # py other pytorch variables back in place
```
araffin, line 386: typo?

Miffyli:
> For get/set parameters, I think we maybe don't need them as we can access model.policy easily, and therefore have access to model.policy.named_parameters(), model.policy.state_dict() and model.policy.load_state_dict().

Same thoughts occurred to me: state_dict is already very convenient to use. Two things worry me though: Should we allow loading parameters from numpy arrays directly (`load_state_dict` does not support this, as far as I am aware), and how to work with agents with multiple networks (e.g. value functions)? I think we should store and load policy and associated q/v-functions together, which could be wrapped under get/set-parameters.

araffin:
>: Should we allow loading parameters from numpy arrays directly 

Doing `params = th.as_tensor(params)` for each parameter does not seem to be a big deal. So I would keep it as is.

>and how to work with agents with multiple networks (e.g. value functions)? I think we should store and load policy and associated q/v-functions together, 

Hmm, `model.policy.state_dict()` already contain all networks. Or you meant something else?

Miffyli:
> Doing params = th.as_tensor(params) for each parameter does not seem to be a big deal. So I would keep it as is.

I will be _that guy_ and argue it would be equally easy to include this in `get_parameters` :). This way the behaviour matches older stable_baselines and users do not have to play with torch to use sb3.

> Hmm, model.policy.state_dict() already contain all networks. Or you meant something else?

I think the get/set_params should follow the state dicts of objects specified by `get_torch_variables`, because e.g. SAC has more stuff than just `model.policy` there:

https://github.com/DLR-RM/stable-baselines3/blob/4fd408bec2a25ba3543520317b753d908078fe5e/stable_baselines3/sac/sac.py#L310

araffin:
>I will be that guy and argue it would be equally easy to include this in get_parameters :). This way the behaviour matches older stable_baselines and users do not have to play with torch to use sb3.

Fair enough ;)

>I think the get/set_params should follow the state dicts of objects specified by get_torch_variables, because e.g. SAC has more stuff than just model.policy there:

Oh, true. But as for now, this is only valid for SAC and it corresponds to a very special variable (entropy temperature).

Miffyli:
> Oh, true. But as for now, this is only valid for SAC and it corresponds to a very special variable (entropy temperature).

Hmm how about the other nn.Modules included in this list? In above SAC example it has policy, actor and critic, all with potentially different parameters. OnPolicyAlgorithms have policy and its optimizer (granted, including all parameters of policy would likely include optimizer too):

https://github.com/DLR-RM/stable-baselines3/blob/f5104a5efca7d1aefbbbb86a41a19c156f6abb2d/stable_baselines3/common/on_policy_algorithm.py#L243

Miffyli:
Added the (long delayed...) functions. If it looks ok, I will check over docs (something seems to be failing there) and update where necessary.

stable_baselines3/common/base_class.py
```diff
@@ -315,41 +355,116 @@ def predict(
...
+
+        for name in params:
+            attr = None
+            try:
+                attr = recursive_getattr(self, name)
+            except Exception:
+                # What errors recursive_getattr could throw? KeyError, but
+                # possible something else too (e.g. if key is an int?).
+                # Catch anything for now.
+                raise ValueError("Key {} is an invalid object name.".format(name))
```
araffin, line 396:
```suggestion
                raise ValueError(f"Key {name} is an invalid object name.")
```

stable_baselines3/common/base_class.py
```diff
@@ -365,14 +480,12 @@ def load(
         model._setup_model()
 
         # put state_dicts back in place
-        for name in params:
-            attr = recursive_getattr(model, name)
-            attr.load_state_dict(params[name])
+        model.set_parameters(params, exact_match=True, device=device)
```
araffin, line 483: why `exact_match` is hardcoded?

stable_baselines3/common/base_class.py
```diff
@@ -315,41 +355,116 @@ def predict(
...
+                # the user has provided a sensible state dictionary.
+                attr.load_state_dict(params[name])
+            else:
+                # Assume attr is th.nn.Module
+                attr.load_state_dict(params[name], strict=exact_match)
+            updated_objects.add(name)
+
+        if exact_match and updated_objects != objects_needing_update:
+            raise ValueError(
+                "Names of parameters do not match agents' parameters: expected {}, got {}".format(
```
araffin, line 422: please use f-strings ;)

araffin:
LGTM ;) 

Maybe one thing to add to the saved model: the SB3 version (could be checked later if needed, but it is good to have it anyway)

stable_baselines3/common/base_class.py
Miffyli: Using exact_match=False would mean some parameters were missing in the saved model file which should not happen unless someone modifies the file. Laying out the hardcoded parameter like this is to signal that we want to make sure every parameter is updated as it was saved, and that nothing is missing.

araffin: ok, fair enough.
What can happen is also that parameters are renamed between versions (it happened to me after refactoring the continuous critic, the name of the parameters were not the same)

araffin:
Is it ready to review now?

stable_baselines3/common/base_class.py
```diff
@@ -513,14 +375,254 @@ def _update_info_buffer(self, infos: List[Dict[str, Any]], dones: Optional[np.nd
...
+    def set_random_seed(self, seed: Optional[int] = None) -> None:
+        """
+        Set the seed of the pseudo-random generators
+        (python, numpy, pytorch, gym, action_space)
+
+        :param seed: (int)
+        """
+        if seed is None:
+            return
+        set_random_seed(seed, using_cuda=self.device == th.device("cuda"))
```
araffin, line 468: since #154 I realized that we should check `self.device.type`, should we fix it here or in a separate PR?

araffin (approved):
LGTM =) (apart from minor remark)

stable_baselines3/common/base_class.py
Miffyli: Sounds like something that could be included here, although would you be able to add it if it is a quick thing? I am reading through the PR but only slowly digesting how it works.

Also any ideas what could be causing the flake8 linting error exactly? My flake8 is not catching it :/

araffin:
>I think the get/set_params should follow the state dicts of objects specified by get_torch_variables, because e.g. SAC has more stuff than just model.policy there:

So at the end, `get/set_parameters` take care of everything but the custom pytorch variables (here only the `log_ent_coef`) which is set/saved separately?

Miffyli:
> So at the end, get/set_parameters take care of everything but the custom pytorch variables (here only the log_ent_coef) which is set/saved separately?

Yeap, with the reasoning that these are rarely used, although if we keep using `torch_variables` it should also be included. This is because tensors lack any `state_dict` operations, but of course could be added with a check for standard tensors. I could do it now, too.

Also flake8 thing pointed out a error I still need to fix, so hold the brakes on merging :)

araffin:
@Miffyli a bit late... this introduces breaking change for previously saved policy, unless `or file_path == "tensors.pth"` is added line 407 of save utils.
I think I will do a quick PR for that.

PR #147, opened 2020-08-23
Add StopTrainingOnMaxEpisodes to callback collection

### LINKED ISSUE #140: [feature request] Add total episodes parameter to model learn method
Hi,

**TL;DR:** I would like the option to pass either a `total_episodes` parameter or a `total_timesteps` to the `model.learn()` method.

Now, for my reasoning. Currently, we can only define the `total_timesteps` when training an agent, as follows.

```
model = A2C('MlpLstmPolicy', env, verbose=1, policy_kwargs=policy_kwargs)
model.learn(total_timesteps=1000)
```

However, for some scenarios (e.g., stock trading), it is quite common to have a fixed number of timesteps per episode, given by the available time-series data points. Also, it can be quite valuable to scan all data points an equal amount of time thoroughly and to determine the number of passes, which is defined by the number of episodes.

Thus, to train for a given number of episodes for a fixed number of timesteps, I have to get the `total_timesteps` value, before passing it to method `model.learn()` as follows:

```
desired_total_episodes = 100
n_points = train_df.shape[0]) # get the number of data points
total_timesteps = desired_total_episodes * n_points
```

Even so, [this answer on StackOverflow](https://stackoverflow.com/a/56766128/2464671) says that

> Where the episode length is known, set it to the desired number of episode you would like to train. However, it might be less because **the agent might not (probably wont) reach max steps every time**.

I must admit I do not know how accurate this answer, but this worries me that my model may not scan all the data equally.

Another option, as discussed in this [issue from previous stable baseline repo](https://github.com/hill-a/stable-baselines/issues/571), is to use a callback function. Still, for this callback approach, I would have to pass a `total_timesteps` variable that is high enough so that I can have the desired number of episodes. Hence, this callback approach seems like an out of the way workaround.

In conclusion, I believe that including the option to pass a `total_episodes` could be a simple and effective approach that would broaden the number of use cases attended by this project.

Thank you for your attention!

xicocaio:
## Description

Added ``StopTrainingOnMaxEpisodes`` callback class to keep track of episodes the current number of episodes ran so far and stop training once the maximum number of episodes ``max_episodes`` is reached.

## Motivation and Context
Closes #140. Also, further details about motivation and implementation discussions at #140. 

## Types of changes

## Checklist:

### DIFF
--- a/docs/guide/callbacks.rst
+++ b/docs/guide/callbacks.rst
@@ -310,20 +310,16 @@ and in total for ``max_episodes * n_envs`` episodes.
 
 .. code-block:: python
 
-    import gym
-
     from stable_baselines3 import A2C
     from stable_baselines3.common.callbacks import StopTrainingOnMaxEpisodes
 
-    # Separate evaluation env
-    eval_env = gym.make('Pendulum-v0')
     # Stops training when the model reaches the maximum number of episodes
-    callback_max_eps = StopTrainingOnMaxEpisodes(max_episodes=5, verbose=1)
+    callback_max_episodes = StopTrainingOnMaxEpisodes(max_episodes=5, verbose=1)
 
     model = A2C('MlpPolicy', 'Pendulum-v0', verbose=1)
     # Almost infinite number of timesteps, but the training will stop
     # early as soon as the max number of episodes is reached
-    model.learn(int(1e10), callback=callback_max_eps)
+    model.learn(int(1e10), callback=callback_max_episodes)
 
 
 .. automodule:: stable_baselines3.common.callbacks
--- a/docs/guide/callbacks.rst
+++ b/docs/guide/callbacks.rst
@@ -292,5 +292,39 @@ An :ref:`EventCallback` that will trigger its child callback every ``n_steps`` t
   model.learn(int(2e4), callback=event_callback)
 
 
+.. _StopTrainingOnMaxEpisodes:
+
+StopTrainingOnMaxEpisodes
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Stop the training upon reaching the maximum number of episodes, regardless of the model's ``total_timesteps`` value.
+Also, presumes that, for multiple environments, the desired behavior is that the agent trains on each env for ``max_episodes``
+and in total for ``max_episodes * n_envs`` episodes.
+
+
+.. note::
+    For multiple environments, the agent will train for a total of ``max_episodes * n_envs`` episodes.
+    However, it can't be guaranteed that this training will occur for an exact number of ``max_episodes`` per environment.
+    Thus, there is an assumption that, on average, each environment ran for ``max_episodes``.
+
+
+.. code-block:: python
+
+    import gym
+
+    from stable_baselines3 import A2C
+    from stable_baselines3.common.callbacks import StopTrainingOnMaxEpisodes
+
+    # Separate evaluation env
+    eval_env = gym.make('Pendulum-v0')
+    # Stops training when the model reaches the maximum number of episodes
+    callback_max_eps = StopTrainingOnMaxEpisodes(max_episodes=5, verbose=1)
+
+    model = A2C('MlpPolicy', 'Pendulum-v0', verbose=1)
+    # Almost infinite number of timesteps, but the training will stop
+    # early as soon as the max number of episodes is reached
+    model.learn(int(1e10), callback=callback_max_eps)
+
+
 .. automodule:: stable_baselines3.common.callbacks
   :members:
--- a/docs/misc/changelog.rst
+++ b/docs/misc/changelog.rst
@@ -13,7 +13,7 @@ Breaking Changes:
 New Features:
 ^^^^^^^^^^^^^
 - Added ``unwrap_vec_wrapper()`` to ``common.vec_env`` to extract ``VecEnvWrapper`` if needed
-- Added ``StopTrainingOnMaxEpisodes`` class to pre-made callback collection - @xicocaio
+- Added ``StopTrainingOnMaxEpisodes`` to callback collection - @xicocaio
 
 Bug Fixes:
 ^^^^^^^^^^
--- a/docs/misc/changelog.rst
+++ b/docs/misc/changelog.rst
@@ -13,7 +13,7 @@ Breaking Changes:
 New Features:
 ^^^^^^^^^^^^^
 - Added ``unwrap_vec_wrapper()`` to ``common.vec_env`` to extract ``VecEnvWrapper`` if needed
-- Added ``StopTrainingOnMaxEpisodes`` to callback collection - @xicocaio
+- Added ``StopTrainingOnMaxEpisodes`` to callback collection (@xicocaio)
 
 Bug Fixes:
 ^^^^^^^^^^
@@ -30,7 +30,7 @@ Others:
 
 Documentation:
 ^^^^^^^^^^^^^^
-- Added ``StopTrainingOnMaxEpisodes`` details and example - @xicocaio
+- Added ``StopTrainingOnMaxEpisodes`` details and example (@xicocaio)
 
 
 
--- a/docs/misc/changelog.rst
+++ b/docs/misc/changelog.rst
@@ -13,6 +13,7 @@ Breaking Changes:
 New Features:
 ^^^^^^^^^^^^^
 - Added ``unwrap_vec_wrapper()`` to ``common.vec_env`` to extract ``VecEnvWrapper`` if needed
+- Added ``StopTrainingOnMaxEpisodes`` class to pre-made callback collection - @xicocaio
 
 Bug Fixes:
 ^^^^^^^^^^
@@ -28,6 +29,9 @@ Others:
 
 Documentation:
 ^^^^^^^^^^^^^^
+- Added ``StopTrainingOnMaxEpisodes`` details and example - @xicocaio
+
+
 
 Pre-Release 0.8.0 (2020-08-03)
 ------------------------------
@@ -392,4 +396,4 @@ And all the contributors:
 @Miffyli @dwiel @miguelrass @qxcv @jaberkow @eavelardev @ruifeng96150 @pedrohbtp @srivatsankrishnan @evilsocket
 @MarvineGothic @jdossgollin @SyllogismRXS @rusu24edward @jbulow @Antymon @seheevic @justinkterry @edbeeching
 @flodorner @KuKuXia @NeoExtended @PartiallyTyped @mmcenta @richardwu @kinalmehta @rolandgvc @tkelestemur @mloo3
-@tirafesi @blurLake @koulakis @joeljosephjin @shwang @rk37 @andyshih12 @RaphaelWag
+@tirafesi @blurLake @koulakis @joeljosephjin @shwang @rk37 @andyshih12 @RaphaelWag @xicocaio
--- a/stable_baselines3/common/callbacks.py
+++ b/stable_baselines3/common/callbacks.py
@@ -453,7 +453,10 @@ class StopTrainingOnMaxEpisodes(BaseCallback):
         self.max_episodes = max_episodes
         self._total_max_episodes = max_episodes
         self.n_episodes = 0
-        self._first_step = True
+
+    def _init_callback(self):
+        # At start set total max according to number of envirnments
+        self._total_max_episodes = self.max_episodes * self.training_env.num_envs
 
     def _on_step(self) -> bool:
         # Checking for both 'done' and 'dones' keywords because:
@@ -461,11 +464,6 @@ class StopTrainingOnMaxEpisodes(BaseCallback):
         # While some models use keyword 'dones' (e.g.,: A2C, PPO)
         done_array = np.array(self.locals.get("done") if self.locals.get("done") is not None else self.locals.get("dones"))
         self.n_episodes += np.sum(done_array).item()
-
-        # Set total max episodes in first step to avoid doing this operation multiple times
-        if self._first_step:
-            self._total_max_episodes = self.max_episodes * self.training_env.num_envs
-            self._first_step = False
 
         continue_training = self.n_episodes < self._total_max_episodes
 
--- a/stable_baselines3/common/callbacks.py
+++ b/stable_baselines3/common/callbacks.py
@@ -469,7 +469,9 @@ class StopTrainingOnMaxEpisodes(BaseCallback):
 
         if self.verbose > 0 and not continue_training:
             mean_episodes_per_env = self.n_episodes / self.training_env.num_envs
-            mean_ep_str = "with an average of {mean_episodes_per_env:.2f} episodes per env" if self.training_env.num_envs > 1 else ""
+            mean_ep_str = (
+                "with an average of {mean_episodes_per_env:.2f} episodes per env" if self.training_env.num_envs > 1 else ""
+            )
 
             print(
                 f"Stopping training with a total of {self.num_timesteps} steps because the "
--- a/stable_baselines3/common/callbacks.py
+++ b/stable_baselines3/common/callbacks.py
@@ -87,7 +87,7 @@ class BaseCallback(ABC):
         """
         self.n_calls += 1
         # timesteps start at zero
-        self.num_timesteps = self.model.num_timesteps + 1
+        self.num_timesteps = self.model.num_timesteps
 
         return self._on_step()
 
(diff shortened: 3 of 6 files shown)


tests/test_callbacks.py
```diff
@@ -10,8 +10,11 @@
     CheckpointCallback,
     EvalCallback,
     EveryNTimesteps,
+    StopTrainingOnMaxEpisodes,
     StopTrainingOnRewardThreshold,
 )
+from stable_baselines3.common.cmd_util import make_vec_env
+from stable_baselines3.common.vec_env import SubprocVecEnv
```
Miffyli, line 17: This could be DummyVecEnv (API is same but no need to hassle with processes).

tests/test_callbacks.py
```diff
@@ -57,6 +63,19 @@ def test_callbacks(tmp_path, model_class):
...
 
+    # Testing for models that support multiple envs
+    if model_class in [A2C, PPO]:
+        envs = make_vec_env(env_name, n_envs=2, seed=0, vec_env_cls=SubprocVecEnv)
+
+        model = model_class("MlpPolicy", envs, policy_kwargs=dict(net_arch=[32]))
+
+        callback = CallbackList([callback_max_eps])
+
+        model.learn(1000, callback=callback)
```
Miffyli, line 74: Looks like should end with maximum episodes reached, being Pendulum-v0 envs with maximum 200 steps. Add check that this indeed happens.

stable_baselines3/common/callbacks.py
```diff
@@ -435,3 +435,47 @@ def _on_step(self) -> bool:
...
+        # While some models use keyword 'dones' (e.g.,: A2C, PPO)
+        done_array = np.array(self.locals.get("done") if self.locals.get("done") is not None else self.locals.get("dones"))
+        self.episodes_played += np.sum(done_array).item()
+
+        # Sets max episodes per env in first step to avoid doing this operation multiple times
+        if self._first_step:
+            self._max_episodes_per_env = self.max_episodes * done_array.size
+            self._first_step = False
+
+        continue_training = bool(self.episodes_played < self._max_episodes_per_env)
```
Miffyli, line 470: No need for `bool`, can do just `self.episodes_played < self._max_episodes_per_env`.

stable_baselines3/common/callbacks.py
```diff
@@ -435,3 +435,47 @@ def _on_step(self) -> bool:
...
+
+class StopTrainingOnMaxEpisodes(BaseCallback):
+    """
+    Stop the training once a maximum number of episodes are played.
+
+    For multiple environments presumes that, the desired behavior is that the agent trains on each env for ``max_episodes``
+    and in total for ``max_episodes * n_envs`` episodes.
+
+    :param max_episodes: (int) Maximum number of episodes to stop training.
+    :param verbose: (int)
```
Miffyli, line 448: Add description for `verbose=1` (prints out when training was killed).

tests/test_callbacks.py
```diff
@@ -57,6 +63,19 @@ def test_callbacks(tmp_path, model_class):
     if os.path.exists(log_folder):
```
Miffyli, line 63: I think this one can be removed.

Miffyli:
Overall looks good on a quick glimpse, dropped in some comments. Will check this more in depth later :)

tests/test_callbacks.py
xicocaio: This was required to make testes for models that support multiple envs (A2C and PPO).

tests/test_callbacks.py
xicocaio: Saw no examples on how to check this automatically, and assumed that was by design, thus just verified behavior manually.

stable_baselines3/common/callbacks.py
xicocaio: For code consistency, followed other code examples that made explicit type declaration even for booleans. E.g,.: In `StopTrainingOnRewardThreshold` we have `continue_training = bool(self.parent.best_mean_reward < self.reward_threshold)`.

stable_baselines3/common/callbacks.py
xicocaio: Aiming at code consistency followed other examples that did not add a description for verbose parameters.

tests/test_callbacks.py
xicocaio: This was a mishap, will remove with a forced push.

docs/guide/callbacks.rst
```diff
@@ -292,5 +292,39 @@ An :ref:`EventCallback` that will trigger its child callback every ``n_steps`` t
...
+
+    # Separate evaluation env
+    eval_env = gym.make('Pendulum-v0')
+    # Stops training when the model reaches the maximum number of episodes
+    callback_max_eps = StopTrainingOnMaxEpisodes(max_episodes=5, verbose=1)
+
+    model = A2C('MlpPolicy', 'Pendulum-v0', verbose=1)
+    # Almost infinite number of timesteps, but the training will stop
+    # early as soon as the max number of episodes is reached
+    model.learn(int(1e10), callback=callback_max_eps)
```
araffin, line 326:
```suggestion
    callback_max_episodes = StopTrainingOnMaxEpisodes(max_episodes=5, verbose=1)

    model = A2C('MlpPolicy', 'Pendulum-v0', verbose=1)
    # Almost infinite number of timesteps, but the training will stop
    # early as soon as the max number of episodes is reached
    model.learn(int(1e10), callback=callback_max_episodes)
```

docs/guide/callbacks.rst
```diff
@@ -292,5 +292,39 @@ An :ref:`EventCallback` that will trigger its child callback every ``n_steps`` t
...
+
+.. code-block:: python
+
+    import gym
+
+    from stable_baselines3 import A2C
+    from stable_baselines3.common.callbacks import StopTrainingOnMaxEpisodes
+
+    # Separate evaluation env
+    eval_env = gym.make('Pendulum-v0')
```
araffin, line 319: not needed, no? (you can also remove the gym import then)

docs/misc/changelog.rst
```diff
@@ -13,6 +13,7 @@ Breaking Changes:
 New Features:
 ^^^^^^^^^^^^^
 - Added ``unwrap_vec_wrapper()`` to ``common.vec_env`` to extract ``VecEnvWrapper`` if needed
+- Added ``StopTrainingOnMaxEpisodes`` class to pre-made callback collection - @xicocaio
```
araffin, line 16:
```suggestion
- Added ``StopTrainingOnMaxEpisodes`` to callback collection (@xicocaio)
```

stable_baselines3/common/callbacks.py
araffin: this was needed because it returned a `np.bool` otherwise. Here, I think it is not needed.

stable_baselines3/common/callbacks.py
```diff
@@ -435,3 +435,47 @@ def _on_step(self) -> bool:
...
+    """
+
+    def __init__(self, max_episodes: int, verbose: int = 0):
+        super(StopTrainingOnMaxEpisodes, self).__init__(verbose=verbose)
+        self.max_episodes = max_episodes
+        self._max_episodes_per_env = max_episodes
+        self.episodes_played = 0
```
araffin, line 455: I would use `n_episodes` so it is not specific to games ;)

stable_baselines3/common/callbacks.py
```diff
@@ -435,3 +435,47 @@ def _on_step(self) -> bool:
...
+
+    def _on_step(self) -> bool:
+        # Checking for both 'done' and 'dones' keywords because:
+        # Some models use keyword 'done' (e.g.,: SAC, TD3, DQN, DDPG)
+        # While some models use keyword 'dones' (e.g.,: A2C, PPO)
+        done_array = np.array(self.locals.get("done") if self.locals.get("done") is not None else self.locals.get("dones"))
+        self.episodes_played += np.sum(done_array).item()
+
+        # Sets max episodes per env in first step to avoid doing this operation multiple times
+        if self._first_step:
```
araffin, line 466: there is a `_init_callback` method for that, and you can use `self.training_env.num_envs` instead of `done_array.size`

stable_baselines3/common/callbacks.py
```diff
@@ -435,3 +435,47 @@ def _on_step(self) -> bool:
...
+            self._max_episodes_per_env = self.max_episodes * done_array.size
```
araffin, line 467: the variable name is not the right one, no?
this is not the number of episodes per env but the total number of episodes, no?

stable_baselines3/common/callbacks.py
```diff
@@ -435,3 +435,47 @@ def _on_step(self) -> bool:
...
+            self._first_step = False
+
+        continue_training = bool(self.episodes_played < self._max_episodes_per_env)
+
+        if self.verbose > 0 and not continue_training:
+            avg_eps_per_env = self.episodes_played // done_array.size
```
araffin, line 473: why not a float division? (and then use `"{:.2f}"` for formatting)

stable_baselines3/common/callbacks.py
xicocaio: You are right, my mistake.

stable_baselines3/common/callbacks.py
xicocaio: My rationale was that the user would expect the training to run for `max_episodes` for each env, and that a few episodes more would not be a problem. Hence, I find that it may be confusing for the user to expect an exact number of episodes during training, and end up with a float number of let's say 10.1 episodes. Also, I believe an exact division will be more common than one that results in a real value, and there is not much information gain in knowing the exact average. What do you think?

tests/test_callbacks.py
xicocaio: @araffin Any comment?

tests/test_callbacks.py
xicocaio: @araffin Any comment?

stable_baselines3/common/callbacks.py
xicocaio: @araffin What do you think?

tests/test_callbacks.py
araffin: https://stable-baselines3.readthedocs.io/en/master/guide/vec_envs.html

I agree with @Miffyli

stable_baselines3/common/callbacks.py
araffin: examples are not always perfect... and more doc is always welcomed ;)

tests/test_callbacks.py
xicocaio: But then I would have to drop the code for testing multiprocessing in A2C and PPO cases.

stable_baselines3/common/callbacks.py
araffin: ok, we just need to make sure that `self.episodes_played // done_array.size` would be greater than one, otherwise the message will be misleading

tests/test_callbacks.py
araffin: You can test with multiple environments but without processes (usually faster), this is what the `DummyVecEnv` is for

see our tutorial https://github.com/araffin/rl-tutorial-jnrr19/tree/sb3 for more details

tests/test_callbacks.py
araffin: as @Miffyli suggests, please use `model.num_timesteps` to check how many episodes were run (because there is 200 timesteps per episode)

stable_baselines3/common/callbacks.py
xicocaio: As I see, this could only happen if the argument `max_episodes` passed by the user is lower than `num_envs`. So, do you prefer that I implicitly set `max_episodes == num_envs` in these cases, and log a warning? Or throw an exception if `max_episodes < num_envs`? Or do you have any other suggestions?

araffin: I thought it should run at least `max_episodes * num_envs`, no? (at least that what is written)

xicocaio: Yes, that is right and the division `n_episodes // num_envs` occurs only if n_episodes >= max_episodes * num_envs. Also, this division will only be less than 1 if `max_episodes < num_envs`.

I am not sure I understand the issue here, so I can just do as you first suggested and change the division to a float one, no problem.

tests/test_callbacks.py
xicocaio: Oh sorry, I did not fully understand that. I thought that `DummyVecEnv` could only be used for single envs. I will change the code accordingly, thank you! =)

tests/test_callbacks.py
xicocaio: Sure, the issue is that I am getting `model.num_timesteps= 398` when I ran multiple envs, and I am not sure why this is happening. I read the documentation about vecenvs and multiple environments, but still do not why this happens, is this an index issue? It is also odd that when I print the callback output it says `Stopping training with a total of 399 steps because`.

araffin: this is normal as `self.num_timesteps += env.num_envs` is incremented after the `callback.on_step()`, I think you can those lines up, same for the off policy case

xicocaio: Sorry, I do not understand. The suggestion was to add a code for checking `num_timesteps`, which I would propose something like this `assert (model.num_timesteps // n_envs) == 200`. Is this right? Where should I place it?

xicocaio: @araffin I think I get it now. Do you want me to move up the line `self.num_timesteps += env.num_envs`, inside the `on_policy_algorithm.py` and `off_policy_algorithm.py` files?

Miffyli: Yes that's how I understood. Personally I think best place is just after doing the calls to `step` functions.

xicocaio: Ok, I will do it. However, the issue is that moving up the `self.num_timesteps += env.num_envs` might break some code. For example, the `on_step()` method in `BaseCallback` has a internal counter `self.num_timesteps = self.model.num_timesteps + 1`. Thus, I am not sure if this was a workaround given how `num_timesteps` is updated inside `on_policy_algorithm.py` and `off_policy_algorithm.py` base classes, or if there is another reason for having this internal counter.

In resume, do you want me to also change this line `self.num_timesteps = self.model.num_timesteps + 1` inside `callbacks.py` file?

araffin: yes, please do (and check that the behavior is the same, for instance using a checkpoint callback)

xicocaio: Sure, I will do it. But, looking closer at both policy base classes, why is this verification `if callback.on_step() is False:`happening so early on? Instead of moving up `self.num_timesteps += env.num_envs` shouldn't the callback check be moved down?

PS: Sorry for all the questions, but it is your project, and I really don't want to do undesired changes. =)

xicocaio:
@Miffyli @araffin Made adjustments based on your feedback, waiting for your review. =)

stable_baselines3/common/callbacks.py
araffin: @xicocaio still valid ;)

stable_baselines3/common/callbacks.py
xicocaio: Sorry, did not fully understand at first. Moved code to `_init_callback`.

xicocaio:
@araffin `make check-codestyle` check is failing on remote, for unchanged files such as `buffers.py`. It works on local though:

```
$ make check-codestyle
# Sort imports
isort --check stable_baselines3/ tests/ docs/conf.py setup.py
# Reformat using black
black --check -l 127 stable_baselines3/ tests/ docs/conf.py setup.py
All done! ✨ 🍰 ✨
77 files would be left unchanged.
```

araffin:
> @araffin `make check-codestyle` check is failed twice on remote, for 7 unchanged files such as `buffers.py`. It works on local though:
> 
> ```
> $ make check-codestyle
> # Sort imports
> isort --check stable_baselines3/ tests/ docs/conf.py setup.py
> # Reformat using black
> black --check -l 127 stable_baselines3/ tests/ docs/conf.py setup.py
> All done! ✨ 🍰 ✨
> 77 files would be left unchanged.
> ```

yes, that's the same issue as in #150 , you need to upgrade your black version.
I will also need to update the docker image.

araffin (approved):
LGTM, thanks =)

PR #154, opened 2020-09-01
Allow to set a device when loading a model

### LINKED ISSUE #153: [feature request] Allow to set the device when loading a model
In common/base_class.py in the 'load' method (lines 319-378), I'd really appreciate it if you could add a 'device="auto" argument to the load method definition (line 319), and set the device using this argument on line 355 instead of the hard-coded parameter.

I can change it and submit a PR on a different branch if you prefer, though it's just a 2 lines change.

Thanks a lot in advance and great work on this repo I really like it so far!

liorcohen5:
Added a 'device' keyword argument to BaseAlgorithm.load(), to enable users to set the model to their device of choice.
Edited test_save_load to also test the load method with all possible devices.
Added the changes to the changelog (I'm not completely confident about the choice of words though).

## Description
Added a 'device' keyword argument with default value 'auto', similarly to the hard-coded value that was used before my change to BaseAlgorithm.load(), and forwarded it to the C'tor call inside the load method (instead of the hard-coded string parameter).

## Motivation and Context
I'm using this repo for my research and when I work with simpler models (tiny MLP for example) it is actually faster to use the CPU even though my machine has a powerful GPU since the overhead of the calls is higher than the benefits.
Thus, when I load the models I'm training, I need to be able to force them to load to the CPU easily.
Since I've seen that the code is already there but currently the device is chosen using a hard-coded string inside the load method, I suggested to make this little but significant change (:

Closes #153 

## Types of changes

## Checklist:

### DIFF
--- a/stable_baselines3/common/save_util.py
+++ b/stable_baselines3/common/save_util.py
@@ -352,6 +352,7 @@ def load_from_pkl(path: Union[str, pathlib.Path, io.BufferedIOBase], verbose=0) 
 def load_from_zip_file(
     load_path: Union[str, pathlib.Path, io.BufferedIOBase],
     load_data: bool = True,
+    device: Union[th.device, str] = "auto",
     verbose=0,
 ) -> (Tuple[Optional[Dict[str, Any]], Optional[TensorDict], Optional[TensorDict]]):
     """
@@ -360,13 +361,14 @@ def load_from_zip_file(
     :param load_path: (str, pathlib.Path, io.BufferedIOBase) Where to load the model from
     :param load_data: Whether we should load and return data
         (class parameters). Mainly used by 'load_parameters' to only load model parameters (weights)
+    :param device: (Union[th.device, str]) Device on which the code should run.
     :return: (dict),(dict),(dict) Class parameters, model state_dicts (dict of state_dict)
         and dict of extra tensors
     """
     load_path = open_path(load_path, "r", verbose=verbose, suffix="zip")
 
     # set device to cpu if cuda is not available
-    device = get_device()
+    device = get_device(device=device)
 
     # Open the zip archive and load data
     try:
--- a/tests/test_save_load.py
+++ b/tests/test_save_load.py
@@ -77,15 +77,18 @@ def test_save_load(tmp_path, model_class):
         model = model_class.load(str(tmp_path / "test_save.zip"), env=env, device=device)
 
         # check if the model was loaded to the correct device
-        assert model.device == get_device(device)
-        assert model.policy.device == get_device(device)
+        assert model.device.type == get_device(device).type
+        assert model.policy.device.type == get_device(device).type
 
         # check if params are still the same after load
         new_params = model.policy.state_dict()
 
         # Check that all params are the same as before save load procedure now
         for key in params:
-            assert th.allclose(params[key], new_params[key]), "Model parameters not the same after save and load."
+            assert new_params[key].device.type == get_device(device).type
+            assert th.allclose(
+                params[key].to("cpu"), new_params[key].to("cpu")
+            ), "Model parameters not the same after save and load."
 
         # check if model still selects the same actions
         new_selected_actions, _ = model.predict(observations, deterministic=True)
--- a/tests/test_save_load.py
+++ b/tests/test_save_load.py
@@ -77,7 +77,8 @@ def test_save_load(tmp_path, model_class):
         model = model_class.load(str(tmp_path / "test_save.zip"), env=env, device=device)
 
         # check if the model was loaded to the correct device
-        assert model.device.type == get_device(device).type
+        assert model.device == get_device(device)
+        assert model.policy.device == get_device(device)
 
         # check if params are still the same after load
         new_params = model.policy.state_dict()
--- a/tests/test_save_load.py
+++ b/tests/test_save_load.py
@@ -13,6 +13,7 @@ from stable_baselines3 import A2C, DDPG, DQN, PPO, SAC, TD3
 from stable_baselines3.common.base_class import BaseAlgorithm
 from stable_baselines3.common.identity_env import FakeImageEnv, IdentityEnv, IdentityEnvBox
 from stable_baselines3.common.save_util import load_from_pkl, open_path, save_to_pkl
+from stable_baselines3.common.utils import get_device
 from stable_baselines3.common.vec_env import DummyVecEnv
 
 MODEL_LIST = [PPO, A2C, TD3, SAC, DQN, DDPG]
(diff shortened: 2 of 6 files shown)


tests/test_save_load.py
```diff
@@ -70,21 +70,26 @@ def test_save_load(tmp_path, model_class):
     # Check
     model.save(tmp_path / "test_save.zip")
     del model
-    model = model_class.load(str(tmp_path / "test_save.zip"), env=env)
 
-    # check if params are still the same after load
-    new_params = model.policy.state_dict()
+    # Check if the model loads as expected for every possible choice of device:
+    for device in ["auto", "cpu", "cuda"]:
```
liorcohen5, line 75: I noticed that the git code comparison looks quite messy. I'm elaborating about the changes I've made here to ease the review process for you:
The actual change that I made here is the added 'for' loop that goes over all possible devices, and at each iteration the device parameter is passed to the call of 'load' (line 76). At the end of each iteration I delete the model (line 92) so it can be loaded cleanly at the next iteration.
Everything else is the same as before, i.e., I've used the same test to ensure proper loading and tested with all possible values of the new argument 'device'.

araffin: it seems that you are actually not testing that the device parameter was successfully used.
Also, you should skip the cuda device if no GPU is available

liorcohen5: 1) You're right. I will work to improve the test.
2) What should be the expected behavior when a user uses "device='cuda'" on a machine with no GPU?
I noticed that the c'tor defaults to using the CPU in that case without notifying the user.
Anyway, I think the test should include all possible inputs while verifying that the outcome matches your expectations. Do you agree?

liorcohen5: I've used in my test the utils.get_device() function (which is used inside the constructor as well) to determine the device. This way, if for example, the behavior of get_device will change so that 'auto' defaults to 'cpu', the test won't break.

tests/test_save_load.py
```diff
@@ -70,21 +71,29 @@ def test_save_load(tmp_path, model_class):
...
-    new_params = model.policy.state_dict()
+    # Check if the model loads as expected for every possible choice of device:
+    for device in ["auto", "cpu", "cuda", th.device("cpu"), th.device("cuda")]:
+        model = model_class.load(str(tmp_path / "test_save.zip"), env=env, device=device)
 
-    # Check that all params are the same as before save load procedure now
-    for key in params:
-        assert th.allclose(params[key], new_params[key]), "Model parameters not the same after save and load."
+        # check if the model was loaded to the correct device
+        assert model.device.type == get_device(device).type
```
araffin, line 80:
```suggestion
        assert model.device == get_device(device)
        assert model.policy.device == get_device(device)
```

tests/test_save_load.py
```diff
@@ -70,21 +71,29 @@ def test_save_load(tmp_path, model_class):
     # Check
     model.save(tmp_path / "test_save.zip")
     del model
-    model = model_class.load(str(tmp_path / "test_save.zip"), env=env)
 
-    # check if params are still the same after load
-    new_params = model.policy.state_dict()
+    # Check if the model loads as expected for every possible choice of device:
+    for device in ["auto", "cpu", "cuda", th.device("cpu"), th.device("cuda")]:
```
araffin, line 76:
```suggestion
    for device in ["auto", "cpu", "cuda"]:
```
Although I agree the test would be more complete when using `th.device` variant,
the conversion from str to `th.device` is handled by `get_device()` and should not be tested here.
The test also run almost 2x faster that way.

araffin:
running on GPU, it yields this error:
```
            # Check that all params are the same as before save load procedure now
            for key in params:
>               assert th.allclose(params[key], new_params[key]), "Model parameters not the same after save and load."
E               RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!

tests/test_save_load.py:87: RuntimeError
```

liorcohen5:
Thanks for all your help! I'll look into it.

liorcohen5:
When comparing the devices in the test, I restored the comparison of types only, since the "get_device()" function doesn't fill the device index, which causes problems.

stable_baselines3/common/save_util.py
```diff
@@ -352,6 +352,7 @@ def load_from_pkl(path: Union[str, pathlib.Path, io.BufferedIOBase], verbose=0)
 def load_from_zip_file(
     load_path: Union[str, pathlib.Path, io.BufferedIOBase],
     load_data: bool = True,
+    device: Union[th.device, str] = "auto",
```
liorcohen5, line 355: does the order of the arguments here makes sense? I'm not sure if I should have added the new argument last, for cases where users didn't use explicit keyword arguments.
On the other hand, I think it makes more sense to be in front of 'verbose'...

liorcohen5:
> When comparing the devices in the test, I restored the comparison of types only, since the "get_device()" function doesn't fill the device index, which causes problems.

Now I'm observing another concerning issue related to this: on my GPU capable machine "test_predict.test_predict" fails on the same assertion (`assert get_device(device) == model.policy.device`) due to the same fact - the `get_device` function doesn't fill the index of the device. This behavior appears on the master branch as well...

araffin:
>doesn't fill the index of the device. This behavior appears on the master branch as well...

I see... but you could easily fix that by passing `device="cuda:0"` in the tests, no?

liorcohen5:
It doesn't fix the issue, unfortunately.  `get_device()` still doesn't fill the index argument. If you look at the implementation of this function, you'll see that in some cases like "auto", `get_device()` uses a hard-coded string with no index.
Also, it's a different test case ("cuda:0" vs "cuda"). I think that the behavior of `get_device()` should be tested in both cases, but it is not directly related to my issue... 
The test that fails on the master branch has nothing to do with my changes (obviously...), so I'm not sure it's right to fix it in this branch. I'd open a bug for this issue and dive into this, and probably fix `get_device()` as well.

araffin:
> It doesn't fix the issue, unfortunately. `get_device()` still doesn't fill the index argument. If you look at the implementation of this function, you'll see that in some cases like "auto", `get_device()` uses a hard-coded string with no index.
> Also, it's a different test case ("cuda:0" vs "cuda"). I think that the behavior of `get_device()` should be tested in both cases, but it is not directly related to my issue...
> The test that fails on the master branch has nothing to do with my changes (obviously...), so I'm not sure it's right to fix it in this branch. I'd open a bug for this issue and dive into this, and probably fix `get_device()` as well.

`get_device("cuda:0")` does fill the index but `get_device("auto")` don't (which is normal), however L148 of `utils.py` needs to be updated (using `.type`).

araffin:
I think as it is fast and easy to fix, please update the tests to use `.type` in the assertion too (even though not directly related to the code you added) and also update L148 of `utils.py` ;).

araffin (approved):
LGTM, thanks =)

Issue #135, opened 2020-08-04
[Bug] Separate feature extractor for target network (SAC/TD3)


araffin:
Same issue as in https://github.com/DLR-RM/stable-baselines3/pull/132
but for SAC/TD3.
It would need special treatment as it would be beneficial to still share the feature extractor between actor and critic to speedup things.

Miffyli:
I _think_ the fix here is just remove the feature-extractor parameters from the list of update-able parameters. I can recall reading somewhere in docs that only the policy updates feature extractor, which is then shared with rest of the networks.

PartiallyTyped:
Not sure about the original SAC paper, but iirc in RAD and CURL the gradients are propagated from the critics.

PR #182, opened 2020-10-11
Add custom arch for off-policy actor/critic networks

### LINKED ISSUE #113: [Feature request] Allow different network architectures for off-policy actor/critic
Currently, for on-policy algorithm we can specify `net_arch=[dict(pi=[64], vf=[64])]`

The idea would be to allow the same (expect the shared part which adds too much complexity) for off-policy algorithms:
`net_arch=[dict(pi=[64], qf=[64])]`.

This should be fairly simple to implement.

araffin:
## Description

Feature described in #113, the main difference with the on-policy equivalent is that we don't allow shared layers (only in the feature extractor) and therefore, `net_arch` can be a dict.

## Motivation and Context

closes #113 

## Types of changes

## Checklist:

### DIFF
--- a/docs/guide/custom_policy.rst
+++ b/docs/guide/custom_policy.rst
@@ -262,7 +262,7 @@ Off-Policy Algorithms
 ^^^^^^^^^^^^^^^^^^^^^
 
 If you need a network architecture that is different for the actor and the critic when using ``SAC``, ``DDPG`` or ``TD3``,
-you can pass a dictionnary of the following structure: ``dict(qf=[<critic network architecture>], pi=[<actor network architecture>])``.
+you can pass a dictionary of the following structure: ``dict(qf=[<critic network architecture>], pi=[<actor network architecture>])``.
 
 For example, if you want a different architecture for the actor (aka ``pi``) and the critic (Q-function aka ``qf``) networks,
 then you can specify ``net_arch=dict(qf=[400, 300], pi=[64, 64])``.
@@ -273,7 +273,7 @@ you only need to specify ``net_arch=[256, 256]`` (here, two hidden layers of 256
 
 .. note::
     Compared to their on-policy counterparts, no shared layers (other than the feature extractor)
-    between the actor and the critic are allowed
+    between the actor and the critic are allowed (to prevent issues with target networks).
 
 
 .. code-block:: python
--- a/docs/guide/custom_policy.rst
+++ b/docs/guide/custom_policy.rst
@@ -258,9 +258,31 @@ If your task requires even more granular control over the policy/value architect
 
 
 
-.. TODO (see https://github.com/DLR-RM/stable-baselines3/issues/113)
-.. Off-Policy Algorithms
-.. ^^^^^^^^^^^^^^^^^^^^^
-..
-.. If you need a network architecture that is different for the actor and the critic when using ``SAC``, ``DDPG`` or ``TD3``,
-.. you can easily redefine the actor class for instance.
+Off-Policy Algorithms
+^^^^^^^^^^^^^^^^^^^^^
+
+If you need a network architecture that is different for the actor and the critic when using ``SAC``, ``DDPG`` or ``TD3``,
+you can pass a dictionnary of the following structure: ``dict(qf=[<critic network architecture>], pi=[<actor network architecture>])``.
+
+For example, if you want a different architecture for the actor (aka ``pi``) and the critic (Q-function aka ``qf``) networks,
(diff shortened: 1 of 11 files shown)


docs/guide/custom_policy.rst
```diff
@@ -258,9 +258,31 @@ If your task requires even more granular control over the policy/value architect
...
-.. Off-Policy Algorithms
-.. ^^^^^^^^^^^^^^^^^^^^^
-..
-.. If you need a network architecture that is different for the actor and the critic when using ``SAC``, ``DDPG`` or ``TD3``,
-.. you can easily redefine the actor class for instance.
+Off-Policy Algorithms
+^^^^^^^^^^^^^^^^^^^^^
+
+If you need a network architecture that is different for the actor and the critic when using ``SAC``, ``DDPG`` or ``TD3``,
+you can pass a dictionnary of the following structure: ``dict(qf=[<critic network architecture>], pi=[<actor network architecture>])``.
```
Miffyli, line 265: Typo "dictionnary"

stable_baselines3/common/torch_layers.py
```diff
@@ -219,3 +219,43 @@ def forward(self, features: th.Tensor) -> Tuple[th.Tensor, th.Tensor]:
...
+
+    For example, to have actor and critic that share the same network architecture,
+    you only need to specify ``net_arch=[256, 256]`` (here, two hidden layers of 256 units each).
+
+    If you want a different architecture for the actor and the critic,
+    then you can specify ``net_arch=dict(qf=[400, 300], pi=[64, 64])``.
+
+    .. note::
+        Compared to their on-policy counterparts, no shared layers (other than the feature extractor)
+        between the actor and the critic are allowed
```
Miffyli, line 248: A word on why would be nice (seems to be difficulties with the parameter updates with target networks?) Same in docs.

docs/guide/custom_policy.rst
```diff
@@ -258,9 +258,31 @@ If your task requires even more granular control over the policy/value architect
...
+For example, if you want a different architecture for the actor (aka ``pi``) and the critic (Q-function aka ``qf``) networks,
+then you can specify ``net_arch=dict(qf=[400, 300], pi=[64, 64])``.
+
+Otherwise, to have actor and critic that share the same network architecture,
+you only need to specify ``net_arch=[256, 256]`` (here, two hidden layers of 256 units each).
+
+
+.. note::
+    Compared to their on-policy counterparts, no shared layers (other than the feature extractor)
+    between the actor and the critic are allowed
```
Miffyli, line 276: Same as later in code: A word on why this is not allowed would be nice.

stable_baselines3/common/torch_layers.py
araffin: yes, mostly for simplicity (and usually not needed)

Miffyli:
Left some small comments. This might be bit of work, but as the issue lies in the correct updating of parameters (a crucial thing for these algorithms), it could make sense to add a test to check that these updates happen correctly. Such code would also be helpful in the future and also for any other off-policy algorithms using similar ideas ^^.

araffin:
> it could make sense to add a test to check that these updates happen correctly.

What do you mean exactly?

Miffyli:
> What do you mean exactly?

That all main network parameters are updated to what is in target network (check it both ways: that all parameters of main network were updated, and that there was no extra parameters in target network).

araffin:
>That all main network parameters are updated to what is in target network (check it both ways: that all parameters of main network were updated, and that there was no extra parameters in target network).

I think this should go in https://github.com/DLR-RM/stable-baselines3/issues/135
regarding that issue, we also need to check that the main network was untouched. I'm not sure how easy it is to test that.

Miffyli:
Hmm ok, I am just mainly worried that the parameter updates do not match and this causes some issues, hence tests would be nice. I think just checking that state_dicts make sense before and after some training could do the trick.

Miffyli:
I went through things and realized polyak-update would raise an exception if parameters were not updated (and updating of parameters is already tested). I added one missing piece of the puzzle which should now confirm that updating of parameters checks out. One final thing would be to make sure `model.parameters()` includes all parameters correctly, but that is unlikely to fail and can be added later. Other than that LGTM!

araffin:
>I went through things and realized polyak-update would raise an exception if parameters were not updated (and updating of parameters is already tested). 

`for param, target_param in zip(params, target_params):` won't raise any error if the length is different?
interesting...

>includes all parameters correctly, but that is unlikely to fail and can be added later.

yes, I would rather do that in #135 if we do it

PR #185, opened 2020-10-12
Fix off-by-one GAE computation

### LINKED ISSUE #183: Better advantage estimation for OnPolicyAlgorithm
Hi.
Your reimplementation helped me to figure out how a2c works. But while I was studying it, I found some places which can be improved in the reimplementation.
1.	There is a bug in https://github.com/DLR-RM/stable-baselines3/blob/a1e055695c3638f9f15de0cb805b8fcbb5c02764/stable_baselines3/common/on_policy_algorithm.py#L182 `compute_returns_and_advantage()` should take values which correspond to the very last observations as it takes the last `dones`. But here the function takes values corresponding to previous observations. It should take values corresponding to `new_obs`, which you need to infer first.
2.	https://github.com/DLR-RM/stable-baselines3/blob/a1e055695c3638f9f15de0cb805b8fcbb5c02764/stable_baselines3/common/buffers.py#L322 The `if` handles only first iteration of the loop. It could be just moved out of the loop.

araffin:
## Description

Fix off-by-one computation (the last value was wrong), move a check out of the for loop and rename `value` to `values` to better reflect what is expected.

Note: unfortunately, I don't think there is a good way to add a test for that...

## Motivation and Context

closes #183 
(similar to a previous PR but without the laziness ;) )

## Types of changes

## Checklist:

- [x] I've written the [CONTRIBUTION](https://github.com/DLR-RM/stable-baselines3/blob/master/CONTRIBUTING.md) guide (**required**)

### DIFF
--- a/stable_baselines3/common/buffers.py
+++ b/stable_baselines3/common/buffers.py
@@ -299,7 +299,7 @@ class RolloutBuffer(BaseBuffer):
         self.generator_ready = False
         super(RolloutBuffer, self).reset()
 
-    def compute_returns_and_advantage(self, last_value: th.Tensor, dones: np.ndarray) -> None:
+    def compute_returns_and_advantage(self, last_values: th.Tensor, dones: np.ndarray) -> None:
         """
         Post-processing step: compute the returns (sum of discounted rewards)
         and GAE advantage.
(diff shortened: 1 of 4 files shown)


stable_baselines3/common/buffers.py
```diff
@@ -310,24 +310,27 @@ def compute_returns_and_advantage(self, last_value: th.Tensor, dones: np.ndarray
...
         # convert to numpy
-        last_value = last_value.clone().cpu().numpy().flatten()
+        last_values = last_values.clone().cpu().numpy().flatten()
 
         last_gae_lam = 0
+        # Last timestep: use the last predicted value for bootstrap
+        # step == self.buffer_size - 1
+        next_non_terminal = 1.0 - dones
+        next_values = last_values
         for step in reversed(range(self.buffer_size)):
```
Miffyli, line 325: I think the "optimization" sought here is to move the first iteration completely outside the loop (i.e. do delta, last_gae_lam and self.advantages steps). This would remove if clause completely and potentially simplify code, but also seem messy with copy-pasted parts, so up to you if you want to include.

araffin: I will probably revert back then

araffin: I don't know, what do you think? (I don't like duplicating code and as you mentioned I don't think we will get any speedup, I did it mostly for clarity)

Miffyli: Fair enough, I was on the edge which would be the way go (clarity/cleanity over last bit of performance).

Miffyli (approved):
LGTM! Now with this we should check how results changed (need to dust off my previous benchmark codes...).

PR #141, opened 2020-08-14
Remove "device" argument from policies

qxcv:
## Description
This fixes issue #111 by removing the `device` keyword argument from policies. It replaces the `device` attribute with a policy that infers the location of the policy from its parameters.

## Motivation and Context

Fixes #111. See that issue for context.

- [x] I have raised an issue to propose this change ([required](https://github.com/DLR-RM/stable-baselines3/blob/master/CONTRIBUTING.md) for new features and bug fixes)

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation (update in the documentation)

## Checklist:

- [x] I've read the [CONTRIBUTION](https://github.com/DLR-RM/stable-baselines3/blob/master/CONTRIBUTING.md) guide (**required**)
- [x] I have updated the changelog accordingly (**required**).
- [ ] My change requires a change to the documentation.
- [x] I have updated the tests accordingly (*required for a bug fix or a new feature*).
- [ ] I have updated the documentation accordingly.
- [x] I have reformatted the code using `make format` (**required**)
- [x] I have checked the codestyle using `make check-codestyle` and `make lint` (**required**)
- [x] I have ensured `make pytest` and `make type` both pass. (**required**)

The `device` argument was not previously mentioned in the Sphinx docs, so I have not updated them (I did update docstrings, though).

PR #194, opened 2020-10-18
Fix initializing CUDA even when `device="cpu"` is used.

### LINKED ISSUE #193: [bug] Agent uses GPU even with `device="cpu"` argument.
Creating agents with `device="cpu"` still loads CUDA libraries and takes up GPU memory when looking at `nvidia-smi`:

```python
from stable_baselines3 import PPO # Happens with A2C too
agent = PPO("MlpPolicy", "CartPole-v1", device="cpu")
```

v0.8.0 is the latest version that does not have this bug.

Versions:
 * Python 3.6
 * Pytorch 1.6

Miffyli:
## Description
`MlpExtractor` defaults to using "auto" device when none is provided (i.e. GPU if available), and this also happened when `device="cpu"` was given to the agent (result of #141). These parameters are then moved to desired device by training algorithm after creation.

This update makes "cpu" the fallback device if none is provided. We might want to formalize that by default everything is put on "cpu" in policies and let learning algorithms do all the moving to right devices.

I could not come up for a reliable this as there is no way to release the memory used from CUDA-initialization (only for clearing any tensors), so if any previous testing code uses any CUDA parts then this can not be tested.

## Motivation and Context
Closes #193

## Types of changes

## Checklist:

### DIFF
--- a/docs/misc/changelog.rst
+++ b/docs/misc/changelog.rst
@@ -19,6 +19,7 @@ Bug Fixes:
 - Fix GAE computation for on-policy algorithms (off-by one for the last value) (thanks @Wovchena)
 - Fix ignoring the exclude parameter when recording logs using json, csv or log as logging format (@SwamyDev)
 - Make ``make_vec_env`` support the ``env_kwargs`` argument when using an env ID str (@ManifoldFR)
+- Fix model creation initializing CUDA even when `device="cpu"` is provided
 
 Deprecations:
 ^^^^^^^^^^^^^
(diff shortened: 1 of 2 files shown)


araffin (approved):
LGTM, if I get it right, the `get_device()` and `device` parameters were working (tensors were moved to cpu after initialization) but not the init?
Unfortunately, threre is also no good way to test that... in the worst case, it does not allocate much on the GPU, no?

Miffyli:
> if I get it right, the get_device() and device parameters were working (tensors were moved to cpu after initialization) but not the init?
Yup. The tensors were initially moved to "auto" device (triggers loading cuda libraries), but learning algorithm then moved them to desired device later.

> Unfortunately, threre is also no good way to test that... in the worst case, it does not allocate much on the GPU, no?
Yeah I digged into this and could not find a good solution :/. The amount of memory allocated seems to vary quite a bit, from 200MB on my Windows machine to 500-800MB on my linux machines. For a single run this is not a problem but becomes a bottleneck if you try to parallelize runs.

PR #196, opened 2020-10-20
Add support to log videos via tensorboard

### LINKED ISSUE #186: [feature request] Support Tensorboard video logging (convenience)
When investigating RL algorithms I've found it quite useful to report videos of a trajectory to Tensorboard, to directly relate it to other performance measures and look at the agent at different stages. 

The implementation would be quite simple I think and if you decide this to be a useful feature I'd be happy to implement it. As a matter of fact I've already done so in my own fork of this repo. Here is how I envisioned its usage:

```python
def render_trajectory(env, model):
    obs = env.reset()
    rewards = []
    done = False
    screens = []
    while not done:
        actions, _ = model.predict(obs, deterministic=True)
        obs, _, done, _ = env.step(actions)
        screen = env.render(mode='rgb_array')
        screens.append(screen.transpose(2, 0, 1))

    return Video(th.ByteTensor([screens]), fps=40)

class ReportTrajectoryCallback(BaseCallback):
    def __init__(self, eval_env, check_freq):
        super().__init__()
        self._eval_env = eval_env
        self._check_freq = check_freq

    def _on_step(self):
        if self.n_calls % self._check_freq == 0:
            video = render_trajectory(self._eval_env, self.model)
            self.logger.record("trajectory/video", video)
        return True

env = gym.make('CartPole-v1')
model = A2C('MlpPolicy', env)
model.learn(total_timesteps=int(1e4), callback=ReportTrajectoryCallback(env, 1000))
```

I'd suggest introducing a `Video` object and a simple extension of the logger to record the video. The `TensorBoardOutputFormat` would simply check for a video object and forward it to the `SummaryWriter`. All other `KVWriter` classes would just ignore it. As mentioned I've done this already for me self. Here is the relevant commit:
https://github.com/SwamyDev/stable-baselines3/commit/d6b15416000011a5aeddd4af1bfb9f9a2d79d22f

Of course I'm open to any different approaches. For instance checking the type directly in the Formater classes is not good object oriented style and maybe an explicit video logging function would be better. However, it is relatively simple and get's the job done.

What do you think? Would this is be a small but useful contribution?

SwamyDev:
## Description
I've introduced a new Video type which simply consists of the frames and the requested frames per second of the video. This type can then be checked against in the logger's format-writer classes and treated accordingly. Currently only the `TensorBoardOutputFormat` supports writing Videos as it is easy to implement and Tensorboard already supports handling video data. All other formats raise a custom NotSupportedError reporting the unsupported format and suggesting how to exclude them from the `record` call. Tests have been added  to check that behavior. To allow video support the `moviepy` package has been added to the `extra` dependencies in the setup.py. Tensorboard needs this package to render videos otherwise it will produce a warning and just ignore the video data.

However, there are some options to consider:
- Removing the `moviepy` dependency from the `extra` dependencies and report a warning or raise an error if a video is logged to Tensorboard and `moviepy` is not installed. This way users can decide whether they want that feature or not
- Instead of capturing the rendered screens via callback extend the `evaluate_policy` and also return the rendered screens if the `render` parameter is True or set to "rgb_array" or a similar mechanism

What do you think? For this first implementation I went with the simplest and least intrusive options. 

## Motivation and Context
Logging an agent's trajectory as a video helps evaluate the agent. One can inspect directly how the agent is behaving during training and see if it is making progress. When tuning parameters or comparing different implementations it's of great help to directly see if the agent's performance is actually improving in addition to the usual reward and loss metrics. In summary, it provides a more complete picture of an agent's learning progress.
closes #186 

## Types of changes

## Checklist:


araffin:
>Removing the moviepy dependency from the extra dependencies and report a warning or raise an error if a video is logged to Tensorboard and moviepy is not installed.

yes, please do

>nstead of capturing the rendered screens via callback extend the evaluate_policy and also return the rendered screens if the render parameter is True or set to "rgb_array" or a similar mechanism

well, it is only three lines of code and actually shows how to use the callback of the `evaluate_policy`, so I wouldn't change that.

docs/guide/tensorboard.rst
```diff
@@ -80,3 +80,44 @@ Here is a simple example on how to log both additional tensor or arbitrary scala
...
+            self._eval_env = eval_env
+            self._check_freq = check_freq
+
+        def _on_step(self) -> bool:
+            if self.n_calls % self._check_freq == 0:
+                screens = []
+
+                def grab_screens(_locals: Dict[str, Any], _globals: Dict[str, Any]):
+                    e = _locals["env"]
+                    screen = e.render(mode="rgb_array")
```
araffin, line 114:
```suggestion
                def grab_screens(_locals: Dict[str, Any], _globals: Dict[str, Any]):
                    screen = self._eval_env.render(mode="rgb_array")
```
this should work, no? (as `self._eval_env` should be a reference to the environment)

docs/guide/tensorboard.rst
```diff
@@ -80,3 +80,44 @@ Here is a simple example on how to log both additional tensor or arbitrary scala
...
+                    screens.append(screen.transpose(2, 0, 1))
```
araffin, line 115: missing comment

docs/guide/tensorboard.rst
```diff
@@ -80,3 +80,44 @@ Here is a simple example on how to log both additional tensor or arbitrary scala
 
 
     model.learn(50000, callback=TensorboardCallback())
+
+Logging Videos
+--------------
+
+Tensorboard supports periodic logging of video data, which helps evaluating agents at various stages during training.
+Here is an example of how to render an episode and log the resulting video to tensorboard at regular intervals:
```
araffin, line 88: you can add a warning here too, saying that moviepy is required

docs/guide/tensorboard.rst
```diff
@@ -80,3 +80,44 @@ Here is a simple example on how to log both additional tensor or arbitrary scala
...
+    import torch as th
+
+    from stable_baselines3 import A2C
+    from stable_baselines3.common.callbacks import BaseCallback
+    from stable_baselines3.common.evaluation import evaluate_policy
+    from stable_baselines3.common.logger import Video
+    from typing import Any, Dict
+
+
+    class ReportTrajectoryCallback(BaseCallback):
```
araffin, line 102: maybe `VideoRecorderCallback` would be more appropriate?

docs/guide/tensorboard.rst
```diff
@@ -80,3 +80,44 @@ Here is a simple example on how to log both additional tensor or arbitrary scala
...
+                    screen = e.render(mode="rgb_array")
+                    screens.append(screen.transpose(2, 0, 1))
+
+                evaluate_policy(self.model, self._eval_env, callback=grab_screens)
+                self.logger.record("trajectory/video", Video(th.ByteTensor([screens]), fps=40))
+            return True
+
+
+    model = A2C("MlpPolicy", "CartPole-v1", tensorboard_log=f"runs/")
+    model.learn(total_timesteps=int(1e4), callback=ReportTrajectoryCallback(gym.make("CartPole-v1"), 1000))
```
araffin, line 123: maybe add a line to define the callback, to make it clearer?

araffin (changes requested):
Minor changes needed, otherwise LGTM =)

docs/guide/tensorboard.rst
SwamyDev: Yes it works and also nicely shortens the function

docs/guide/tensorboard.rst
SwamyDev: Done, I agree it makes the example more expressive

docs/guide/tensorboard.rst
SwamyDev: Yes!

stable_baselines3/common/logger.py
```diff
@@ -23,6 +23,28 @@
...
+    """
+    Video data class storing the video frames and the frame per seconds
+    """
+
+    def __init__(self, frames: th.Tensor, fps: Union[float, int]):
+        self.frames = frames
+        self.fps = fps
+
+
+class FormatUnsupportedError(NotImplementedError):
```
araffin, line 36: maybe say also which value is not supported by this format

araffin (commented):
I tested the feature, it works and it is quite nice for debugging =)
however, it is super-slow (0.1s per frame), is it expected? Maybe we could add a downsampling option?
Is there no other way to log videos? (it seems that we log a gif at the end, but we may have to re-implement the video transformation if we want to bypass PyTorch internals...)

stable_baselines3/common/logger.py
SwamyDev: You got me there, I was probably a bit lazy on the error message ^^.

SwamyDev:
> I tested the feature, it works and it is quite nice for debugging =)
Awesome! :)

> however, it is super-slow (0.1s per frame), is it expected? Maybe we could add a downsampling option?
Which environment did you test it on and which fps did you use? I've tested it on CartPole and Breakout where 40 fps gave me a smooth gif - even more on the fast side of things. Does the frame rate change for you when you change the fps parameter? I noticed that it gets into troubles when using really high fps rates. Internally it seems to use moviepy's `ImageSequenceClip` which just inverts the fps to get frame duration. I guess it runs into numerical problems when they get to small, but I'd need to further investigate that. For environments with long trajectories one easy option would be to just record every nth frame in the `grab_screens` callback - you wouldn't need all frames anyways to get a sense on how the agent is doing.

> Is there no other way to log videos? (it seems that we log a gif at the end, but we may have to re-implement the video transformation if we want to bypass PyTorch internals...)
I guess we could create our own gif from the frames and just use `add_image`, which is practically what `add_video` does via `moviepy`,

araffin:
>Which environment did you test it on and which fps did you use?

The one from the doc.
I mainly meant that writing the video is quite slow (the call to `add_video()` is the bottleneck), then it shows normally in tensorboard.

>For environments with long trajectories one easy option would be to just record every nth frame in the grab_screens callback 

That's what I meant by "downsampling".

stable_baselines3/common/logger.py
```diff
@@ -23,6 +23,28 @@
...
+    def __init__(self, unsupported_formats: Sequence[str], value: Any):
```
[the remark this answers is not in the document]
SwamyDev: Just to avoid calling the `str(value)` function on each calling site where this error is raised, but you're right - it is easier to read and also gives potential users of the error class the possibility to log customized value descriptions.

SwamyDev:
Ah I understand now. Yes that is quite slow, but to be honest, I've never noticed it before. In practice I don't use video logging that frequently during training that it actually has an impact on training time. However, of course I don't know if other user would want to log more frequently. Should I add the downsampling to the example, or is this something for the contribution repo?

araffin:
> Should I add the down-sampling to the example, or is this something for the contribution repo?

no, the full-featured callback should be for the contrib repo ;)

>I could also investigate a bit how to create a gif from frames more efficiently in memory? 

don't invest too much time, as I'm not even sure if we can have something faster.
Best would be to solve it from PyTorch/Tensorboard side.

stable_baselines3/common/logger.py
```diff
@@ -83,6 +105,9 @@ def write(self, key_values: Dict, key_excluded: Dict, step: int = 0) -> None:
             if excluded is not None and ("stdout" in excluded or "log" in excluded):
                 continue
 
+            if isinstance(value, Video):
+                raise FormatUnsupportedError(["stdout", "log"], str(value))
```
araffin, line 109: ok, I think there was a misunderstanding there.
why would you use `str(value)` instead of something more readable like `"video"`?

SwamyDev: Ah I understand now :). Well my reasoning was that I could just use the string representation of the value as it might contain interesting information of the state of the value object, but it does make the error message quite verbose.

araffin: ok, but then in case of the video, you will just check bytes, no?

SwamyDev:
> don't invest too much time, as I'm not even sure if we can have something faster.

A quick google research turned up a few quick solutions, but they'd all mean extra dependencies. Most people seem to use moviepy for this kind of task anyways, just like Tensorboard does. I could write a gif converter from scratch - it's shouldn't bee too hard, but I think out of scope for this PR (not to mention the maintenance burden). But we could come back to this in case we really needed it.

> Best would be to solve it from PyTorch/Tensorboard side.

Yes I agree. Also with the down-sample capability we could already address this issue on our end.

araffin (approved):
LGTM, thanks =)

araffin:
Small follow-up: after quick trials, it seems that the `th.ByteTensor([screens])` is the part that is quite slow too

PR #198, opened 2020-10-23
Fix off policy features extractor

### LINKED ISSUE #135: [Bug] Separate feature extractor for target network (SAC/TD3)

araffin:
## Description

Fixes the issue that the feature extractor was shared between network and target.
I also added a test to prevent any regression.

Note: for TD3, the effective polyak coefficient for the feature extractor will 2 * tau now
I hope it is not an issue?

## Motivation and Context

closes #135

## Types of changes

## Checklist:

### DIFF
--- a/stable_baselines3/sac/policies.py
+++ b/stable_baselines3/sac/policies.py
@@ -309,7 +309,7 @@ class SACPolicy(BasePolicy):
             self.critic = self.make_critic(features_extractor=None)
             critic_parameters = self.critic.parameters()
 
-        # Critic target should not share the feature extactor with critic
+        # Critic target should not share the features extractor with critic
         self.critic_target = self.make_critic(features_extractor=None)
         self.critic_target.load_state_dict(self.critic.state_dict())
 
--- a/stable_baselines3/sac/policies.py
+++ b/stable_baselines3/sac/policies.py
@@ -296,7 +296,12 @@ class SACPolicy(BasePolicy):
 
         self.critic = self.make_critic()
         self.critic_target = self.make_critic()
+        # Critic target should not share the feature extactor with critic
+        self.critic_target.features_extractor = self.features_extractor_class(
+            self.observation_space, **self.features_extractor_kwargs
+        )
         self.critic_target.load_state_dict(self.critic.state_dict())
+
         # Do not optimize the shared feature extractor with the critic loss
         # otherwise, there are gradient computation issues
         # Another solution: having duplicated features extractor but requires more memory and computation
--- a/stable_baselines3/sac/policies.py
+++ b/stable_baselines3/sac/policies.py
@@ -280,7 +280,13 @@ class SACPolicy(BasePolicy):
         }
         self.actor_kwargs.update(sde_kwargs)
         self.critic_kwargs = self.net_args.copy()
-        self.critic_kwargs.update({"n_critics": n_critics, "net_arch": critic_arch})
+        self.critic_kwargs.update(
+            {
+                "n_critics": n_critics,
+                "net_arch": critic_arch,
+                "share_features_extractor": share_features_extractor,
+            }
+        )
 
         self.actor, self.actor_target = None, None
         self.critic, self.critic_target = None, None
--- a/stable_baselines3/sac/policies.py
+++ b/stable_baselines3/sac/policies.py
@@ -90,7 +90,7 @@ class Actor(BasePolicy):
 
         if self.use_sde:
             latent_sde_dim = last_layer_dim
-            # Separate feature extractor for gSDE
+            # Separate features extractor for gSDE
             if sde_net_arch is not None:
                 self.sde_features_extractor, latent_sde_dim = create_sde_features_extractor(
                     features_dim, sde_net_arch, activation_fn
@@ -211,7 +211,7 @@ class SACPolicy(BasePolicy):
     :param clip_mean: Clip the mean output when using gSDE to avoid numerical instability.
     :param features_extractor_class: Features extractor to use.
     :param features_extractor_kwargs: Keyword arguments
-        to pass to the feature extractor.
+        to pass to the features extractor.
     :param normalize_images: Whether to normalize images or not,
          dividing by 255.0 (True by default)
     :param optimizer_class: The optimizer to use,
@@ -219,6 +219,8 @@ class SACPolicy(BasePolicy):
     :param optimizer_kwargs: Additional keyword arguments,
         excluding the learning rate, to pass to the optimizer
     :param n_critics: Number of critic networks to create.
+    :param share_features_extractor: Whether to share or not the features extractor
+        between the actor and the critic (this saves computation time)
     """
 
     def __init__(
@@ -239,6 +241,7 @@ class SACPolicy(BasePolicy):
         optimizer_class: Type[th.optim.Optimizer] = th.optim.Adam,
         optimizer_kwargs: Optional[Dict[str, Any]] = None,
         n_critics: int = 2,
+        share_features_extractor: bool = True,
     ):
         super(SACPolicy, self).__init__(
             observation_space,
@@ -258,17 +261,11 @@ class SACPolicy(BasePolicy):
 
         actor_arch, critic_arch = get_actor_critic_arch(net_arch)
 
-        # Create shared features extractor
-        self.features_extractor = features_extractor_class(self.observation_space, **self.features_extractor_kwargs)
-        self.features_dim = self.features_extractor.features_dim
-
         self.net_arch = net_arch
         self.activation_fn = activation_fn
         self.net_args = {
             "observation_space": self.observation_space,
             "action_space": self.action_space,
-            "features_extractor": self.features_extractor,
-            "features_dim": self.features_dim,
             "net_arch": actor_arch,
             "activation_fn": self.activation_fn,
             "normalize_images": normalize_images,
@@ -287,6 +284,7 @@ class SACPolicy(BasePolicy):
 
         self.actor, self.actor_target = None, None
         self.critic, self.critic_target = None, None
+        self.share_features_extractor = share_features_extractor
 
         self._build(lr_schedule)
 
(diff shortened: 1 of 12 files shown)


tests/test_cnn.py
```diff
@@ -35,3 +38,113 @@ def test_cnn(tmp_path, model_class):
...
+    kwargs = dict(buffer_size=250, learning_starts=100, policy_kwargs=dict(features_extractor_kwargs=dict(features_dim=32)))
+    model = model_class("CnnPolicy", env, seed=0, **kwargs)
+
+    patch_dqn_names_(model)
+
+    # Critic and target should be equal at the begginning of training
+    params_should_match(model.critic.parameters(), model.critic_target.parameters())
+
+    # TD3 has also a target actor net
+    if model_class == TD3:
```
AdamGleave, line 80: Special-casing this is annoying. How about having a list of attribute pairs to check? Then you can do:

```
for xattr, yattr in attr_pairs:
  params_should_match(getattr(model, xattr).parameters(), getattr(model, yattr).parameters())
```

You could then also eliminate `patch_dqn_names_`

tests/test_cnn.py
```diff
@@ -35,3 +38,113 @@ def test_cnn(tmp_path, model_class):
...
+    # Critic and target should differ
+    params_should_differ(model.critic.parameters(), model.critic_target.parameters())
+
+    if model_class == TD3:
+        params_should_differ(model.actor.parameters(), model.actor_target.parameters())
+
+        # Actor and critic feature extractor should be the same
+        check_td3_feature_extractor_match(model)
+
+    # Re-initialize and collect some random data (without doing gradient steps)
```
AdamGleave, line 94:
```suggestion
    # Re-initialize and collect some random data (without doing gradient steps, since 10 < learning_starts = 100)
```
Was initially confused why there are no gradient steps here

tests/test_cnn.py
```diff
@@ -35,3 +38,113 @@ def test_cnn(tmp_path, model_class):
...
+
+
+def check_td3_feature_extractor_match(model):
+    for (key, actor_param), critic_param in zip(model.actor_target.named_parameters(), model.critic_target.parameters()):
+        if "features_extractor" in key:
+            assert th.allclose(actor_param, critic_param)
+
+
+@pytest.mark.parametrize("model_class", [SAC, TD3, DQN])
+def test_feature_extractor_target_net(model_class):
```
AdamGleave, line 67: I found this function hard to follow. That's not a big deal, it's test code. I think suggestion (below) to take list of attribute-pairs on a per-algorithm case would reduce special cases making it easier to follow. But if it's still convoluted, I wonder if it's actually better to just bite some code duplication and have a different test for each class? (We might be able to factor out some common code.)

AdamGleave (changes requested):
Thanks for fixing this!

I'm a bit worried we could end up with similar bugs in the future. I've suggested below making the choice of whether to share weights explicit by a parameter in `make_critic` (or `make_actor` etc). I'd also be happy with other ways of making this less error prone -- e.g. having `feature_extractor` be separated out from `self.net_kwargs` and then explicitly copying it into `critic_kwargs` but not some new `critic_target_kwargs`.

I also made some suggestions on how to improve comprehensibility of the test code, though I think this is "nice to have" rather than essential.

tests/test_cnn.py
araffin: good point

stable_baselines3/sac/policies.py
```diff
@@ -296,7 +296,12 @@ def _build(self, lr_schedule: Callable) -> None:
 
         self.critic = self.make_critic()
         self.critic_target = self.make_critic()
+        # Critic target should not share the feature extactor with critic
+        self.critic_target.features_extractor = self.features_extractor_class(
```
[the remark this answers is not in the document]
araffin: I like the suggestion ;)

tests/test_cnn.py
araffin: I do agree this is quite hard to follow (I wrote it on the fly without the big picture in mind ^^")
On the other hand, the checks are tricky anyway... I will see what I can do, maybe one test per algorithm yes.

araffin:
I've added a new argument to control if the features extractor is shared or not.
Regarding the tests, I will see what I can do, not sure if we can make it easy to read.

araffin:
So I tried to duplicate some code to avoid special cases but it does not look so much better... (see below)

```python

@pytest.mark.parametrize("share_features_extractor", [True, False])
def test_features_extractor_target_td3(share_features_extractor):
    env = FakeImageEnv(screen_height=40, screen_width=40, n_channels=1, discrete=False)
    # Avoid memory error when using replay buffer
    # Reduce the size of the features
    kwargs = dict(
        buffer_size=250,
        learning_starts=100,
        policy_kwargs=dict(
            features_extractor_kwargs=dict(features_dim=32),
            share_features_extractor=share_features_extractor,
        ),
    )

    model = TD3("CnnPolicy", env, seed=0, **kwargs)

    if share_features_extractor:
        # Check that the objects are the same and not just copied
        assert id(model.policy.actor.features_extractor) == id(model.policy.critic.features_extractor)
        assert id(model.policy.actor_target.features_extractor) == id(model.policy.critic_target.features_extractor)
        # Actor and critic feature extractor should be the same
        td3_features_extractor_check = check_td3_feature_extractor_match
    else:
        # Actor and critic feature extractor should differ same
        td3_features_extractor_check = check_td3_feature_extractor_differ
        # Check that the object differ
        assert id(model.policy.actor.features_extractor) != id(model.policy.critic.features_extractor)
        assert id(model.policy.actor_target.features_extractor) != id(model.policy.critic_target.features_extractor)

    # Critic and target should be equal at the begginning of training
    params_should_match(model.critic.parameters(), model.critic_target.parameters())
    # TD3 has also a target actor net
    params_should_match(model.actor.parameters(), model.actor_target.parameters())

    model.learn(200)

    # Critic and target should differ
    params_should_differ(model.critic.parameters(), model.critic_target.parameters())
    params_should_differ(model.actor.parameters(), model.actor_target.parameters())
    td3_features_extractor_check(model)

    # Re-initialize and collect some random data (without doing gradient steps,
    # since 10 < learning_starts = 100)
    model = TD3("CnnPolicy", env, seed=0, **kwargs).learn(10)

    original_param = deepcopy(list(model.critic.parameters()))
    original_target_param = deepcopy(list(model.critic_target.parameters()))
    original_actor_target_param = deepcopy(list(model.actor_target.parameters()))

    # Deactivate copy to target
    model.tau = 0.0
    model.train(gradient_steps=1)

    # Target should be the same
    params_should_match(original_target_param, model.critic_target.parameters())
    params_should_match(original_actor_target_param, model.actor_target.parameters())
    td3_features_extractor_check(model)

    # not the same for critic net (updated by gradient descent)
    params_should_differ(original_param, model.critic.parameters())

    # Update the reference as it should not change in the next step
    original_param = deepcopy(list(model.critic.parameters()))
    original_actor_param = deepcopy(list(model.actor.parameters()))

    # Deactivate learning rate
    model.lr_schedule = lambda _: 0.0
    # Re-activate polyak update
    model.tau = 0.01

    model.train(gradient_steps=1)

    # Target should have changed now (due to polyak update)
    params_should_differ(original_target_param, model.critic_target.parameters())
    params_should_differ(original_actor_target_param, model.actor_target.parameters())
    # Critic should be the same
    params_should_match(original_param, model.critic.parameters())
    params_should_match(original_actor_param, model.actor.parameters())
    td3_features_extractor_check(model)


def test_features_extractor_target_dqn(model_class, share_features_extractor):
    env = FakeImageEnv(screen_height=40, screen_width=40, n_channels=1, discrete=True)
    # Avoid memory error when using replay buffer
    # Reduce the size of the features
    kwargs = dict(buffer_size=250, learning_starts=100, policy_kwargs=dict(features_extractor_kwargs=dict(features_dim=32)))

    model = DQN("CnnPolicy", env, seed=0, **kwargs)

    # Critic and target should be equal at the begginning of training
    params_should_match(model.q_net.parameters(), model.q_net_target.parameters())

    model.learn(200)

    # Critic and target should differ
    params_should_differ(model.q_net.parameters(), model.q_net_target.parameters())

    # Re-initialize and collect some random data (without doing gradient steps,
    # since 10 < learning_starts = 100)
    model = DQN("CnnPolicy", env, seed=0, **kwargs).learn(10)

    original_param = deepcopy(list(model.q_net.parameters()))
    original_target_param = deepcopy(list(model.q_net_target.parameters()))

    # Deactivate copy to target
    model.tau = 0.0
    model.train(gradient_steps=1)

    # Target should be the same
    params_should_match(original_target_param, model.q_net_target.parameters())
    # not the same for q net (updated by gradient descent)
    params_should_differ(original_param, model.q_net.parameters())

    # Update the reference as it should not change in the next step
    original_param = deepcopy(list(model.q_net.parameters()))

    # Deactivate learning rate
    model.lr_schedule = lambda _: 0.0
    # Re-activate polyak update
    model.tau = 0.01
    # Special case for DQN: target net is updated in the `collect_rollout()`
    # not the `train()` method
    model.target_update_interval = 1
    model._on_step()

    model.train(gradient_steps=1)

    # Target should have changed now (due to polyak update)
    params_should_differ(original_target_param, model.q_net_target.parameters())

    # Networks should be the same
    params_should_match(original_param, model.q_net.parameters())
```

stable_baselines3/sac/policies.py
```diff
@@ -283,24 +280,39 @@ def __init__(
...
+            # Do not optimize the shared features extractor with the critic loss
+            # otherwise, there are gradient computation issues
+            critic_parameters = [param for name, param in self.critic.named_parameters() if "features_extractor" not in name]
+        else:
+            # Create a separate features extractor for the critic
+            # this requires more memory and computation
+            self.critic = self.make_critic(features_extractor=None)
+            critic_parameters = self.critic.parameters()
+
+        # Critic target should not share the feature extactor with critic
```
AdamGleave, line 312:
```suggestion
        # Critic target should not share the features extractor with critic
```

stable_baselines3/td3/policies.py
```diff
@@ -135,35 +138,54 @@ def __init__(
...
         self.actor_target.load_state_dict(self.actor.state_dict())
+
         self.actor.optimizer = self.optimizer_class(self.actor.parameters(), lr=lr_schedule(1), **self.optimizer_kwargs)
-        self.critic = self.make_critic()
-        self.critic_target = self.make_critic()
+
+        if self.share_features_extractor:
+            self.critic = self.make_critic(features_extractor=self.actor.features_extractor)
+            # Critic target should not share the features extactor with critic
+            # but it can share it with the target as actor and critic are sharing
```
AdamGleave, line 179:
```suggestion
            # but it can share it with the actor target as actor and critic are sharing
```

AdamGleave (approved):
LGTM -- made two suggestions to fix typos but otherwise all good.

Shame splitting up the tests did not help, but I agree it seems better to keep it as-is.

PR #200, opened 2020-10-23
Add imitation library docs

shwang:
Adding quickstart for `imitation` based on discussion in https://github.com/HumanCompatibleAI/imitation/issues/190.


shwang:
@AdamGleave is there anything you'd want to have added to this page?

docs/guide/imitation.rst
```diff
@@ -0,0 +1,55 @@
...
+
+
+It also provides `CLI scripts <#Sacred-CLI-Quickstart>`_ for training and saving
+demonstrations from RL experts, and for training imitation learners on these demonstrations.
+
+
+Installation
+------------
+
+Requires Python 3.7+.
```
araffin, line 22: Why is that? What are you using which make it incompatible with python 3.6?

docs/guide/imitation.rst
```diff
@@ -0,0 +1,55 @@
...
+The `imitation <https://github.com/HumanCompatibleAI/imitation>`_ library implements
+imitation learning algorithms on top of Stable-Baselines3, including:
+
+  - Behavioral Cloning
+  - `DAgger <https://arxiv.org/pdf/1011.0686.pdf>`_ with synthetic examples
+  - `Adversarial Inverse Reinforcement Learning <https://arxiv.org/abs/1710.11248>`_ (AIRL)
+  - `Generative Adversarial Imitation Learning <https://arxiv.org/abs/1606.03476>`_  (GAIL)
+
+
+It also provides `CLI scripts <#Sacred-CLI-Quickstart>`_ for training and saving
```
araffin, line 15: This link seems wrong

araffin (commented):
Minor comments other LGTM, thanks =)

Also, could update the migration guide too? (linking to this page where it says that "GAIL is not supported in SB3")

docs/guide/imitation.rst
AdamGleave: `dataclasses` is one thing (used as a data type for transitions and trajectories).

docs/guide/imitation.rst
AdamGleave: I've fixed it.

docs/guide/imitation.rst
araffin: that's a nice feature =) (I did not know about, I've some hack to do that before)

AdamGleave (commented):
I've pushed some commits fixing syntax errors and updating the migration guide. LGTM now but I'll let @araffin do a final review. @shwang nothing else I'd add -- this is just meant to be a teaser for people to check out our codebase. I wonder if a code extract showing how easy it is to use the Python interface would be worthwhile -- but I worry it'll just get out of date (bad examples worse than no example).

araffin (approved):
LGTM, thanks =)
(that's only meant as a basis anyway, user interested should go to your repo for more details)

PR #143, opened 2020-08-15
[WIP] Assume environments are channels-first by default

qxcv:
## Description
This change removes all automatic image transposes, including automatic application of `VecTranspose`. Instead, environments are assumed to be channels-first; users with channels-last environments can wrap them with `VecTranspose` before passing them to SB3 algorithms. It also makes a few other changes that are necessary for compatibility with channels-first environments:

- `VecFrameStack` is updated to support channels-first environments (or channels in any position, really).
- Atari environment constructors now automatically wrap with `VecTranspose`.

## Motivation and Context

This PR is one way of addressing the feature request in #142. It also resolves the original confusion behind #103.

- [x] I have raised an issue to propose this change ([required](https://github.com/DLR-RM/stable-baselines3/blob/master/CONTRIBUTING.md) for new features and bug fixes)

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation (update in the documentation)

(this is not really a bug fix or a new feature, but rather a change of functionality)

## Checklist:

- [x] I've read the [CONTRIBUTION](https://github.com/DLR-RM/stable-baselines3/blob/master/CONTRIBUTING.md) guide (**required**)
- [x] I have updated the changelog accordingly (**required**).
- [x] My change requires a change to the documentation.
- [x] I have updated the tests accordingly (*required for a bug fix or a new feature*).
- [x] I have updated the documentation accordingly.
- [x] I have reformatted the code using `make format` (**required**)
- [x] I have checked the codestyle using `make check-codestyle` and `make lint` (**required**)
- [x] I have ensured `make pytest` and `make type` both pass. (**required**)

Issue #212, opened 2020-11-01
Transposing image environments even if it's not needed.


avandekleut:
Default in the base class is to always transpose image-based environments under the assumption that they are channels-last. 
This is not always true.

https://github.com/DLR-RM/stable-baselines3/blob/6327cc6156756f5cd0b1fb600c346fc21fbd7b11/stable_baselines3/common/base_class.py#L179-L182

Code to reproduce an issue: Make an environment from pixels with `CxHxW` ordering and use a `CnnPolicy`.

```python
import gym
import dmc2gym
from stable_baselines3 import SAC

env = dmc2gym.make(domain_name='point_mass', task_name='easy', seed=1, from_pixels=True, visualize_reward=False)
print(env.observation_space.shape)
model = SAC(
    "CnnPolicy",
    env=env,
)
```

Possible solutions, none of which are very elegant: 
- Check membership in a set of known environments that are channels-last and transpose
- Guess which dimension is channels (the smallest one?)

Miffyli:
Thanks for reporting this, but it is already covered by #142 with on-going PR #143. Lets move discussion to the PR and close this one.

PR #213, opened 2020-11-01
Avoid transposing channel-first envs

### LINKED ISSUE #142: [feature request] Channels-first environments without extra transpose wrapper
At present, image-based environments are automatically wrapped in a `VecTranspose` wrapper when they are passed to a policy. Likewise, observations passed to `policy.predict()` are automatically transposed so that the last channel appears first. This adds a bit of unnecessary overhead when working with channels-first environments: users need to wrap such environments in a special wrapper that puts channels last. The wrapper is then "undone" by `VecTranspose`, yielding the original (correct) observation. It would be nice if there was some way to support channels-first environments without two unnecessary transposes.

One way of fixing this would be to simply assume that environments are channels-first; environments that are not channels-first can be wrapped in a `VecTranspose` wrapper before being passed to an RL algorithm. This has a few advantages:

- It simplifies the behaviour of the code slightly, removing surprises like #103 and eliminating the need for transpose-aware checks like the one in `common.utils.check_for_correct_spaces()`
- It removes unnecessary overhead for channels-first environments
- It makes it easier for other libraries to interoperate with SB3 policies. For example, `imitation` trains SB3 policies using behavioural cloning, but that requires all of `imitation` to be aware of the heuristics that SB3 uses to decide whether to transpose images.

I've made a draft PR that does this in #143.

### LINKED ISSUE #148: [feature request] More Informative Errors on `is_image_space`
Currently,  errors resulting from failures of `is_image_space` as it is called [here](https://github.com/DLR-RM/stable-baselines3/blob/master/stable_baselines3/common/torch_layers.py#L66) in the NatureCNN torch layer are pretty uninformative. 

At the beginning of `NatureCNN`'s __init__, it calls: 

```
        assert is_image_space(observation_space), (
            "You should use NatureCNN "
            f"only with images not with {observation_space} "
            "(you are probably using `CnnPolicy` instead of `MlpPolicy`)"
        )
```

Within `is_image_space`, there are various failure modes, including: 
1. Observation space dtype isn't `uint8`
2. Observation space has wrong low/high values 
3. Observation space has the wrong channel orientation 

All of which are all currently grouped under one error template that assumes the user is failing to use an image observation (when it may just be that the image observation is currently formatted as a non-uint data type, or is accidentally pre-normalized to not be 0-255). 

I think this could be pretty easily resolved by just reformatting `is_image_space` to return both a boolean and a failure reason (which could be None in cases of non-failure), and that failure reason could be printed out on the NatureCNN level. If maintainers would be amenable to a solution of this general sort, I'd be happy to put together a PR for it.

Miffyli:
As discussed in #142  and #143 (and pointed out in recent issue again #212), currently code always transposes image-based environments even if they were channel-first already. This PR fixes that and also updates `VecFrameStack` to support stacking on first channel (supporting other dimensions requires bit more work).

~~@araffin few questions:~~
* ~~Should `VecFrameStack` automatically detect the axis to stack on or do we just rely on user input (this is hard to catch tho)?~~
* ~~You mentioned `predict()` should be updated but it seems like it already check if shapes match, so no need to update it for this.~~

## Motivation and Context
Not closing above issues as code without transpose checks could solve other confusions.

closes #142 
closes #143 
closes #148 

## Types of changes

## Checklist:


stable_baselines3/common/preprocessing.py
```diff
@@ -6,6 +6,19 @@
 from torch.nn import functional as F
 
 
+def is_image_space_channels_first(observation_space: spaces.Space) -> bool:
+    """
+    Check if an image observation space (see ``is_image_space``)
+    is channels-first (CxHxW, True) or channels-last (CxHxW, False).
```
araffin, line 12:
```suggestion
    is channels-first (CxHxW, True) or channels-last (HxWxC, False).
```

stable_baselines3/common/preprocessing.py
```diff
@@ -6,6 +6,19 @@
 from torch.nn import functional as F
 
 
+def is_image_space_channels_first(observation_space: spaces.Space) -> bool:
```
araffin, line 9:
```suggestion
def is_image_space_channels_first(observation_space: spaces.Box) -> bool:
```

stable_baselines3/common/preprocessing.py
```diff
@@ -6,6 +6,19 @@
...
+    """
+    Check if an image observation space (see ``is_image_space``)
+    is channels-first (CxHxW, True) or channels-last (CxHxW, False).
+
+    Use a heuristic that a channel-dimension is the smallest of the three.
```
araffin, line 14: you are actually not checking that...

stable_baselines3/common/vec_env/vec_frame_stack.py
```diff
@@ -9,19 +9,24 @@
...
+    Frame stacking wrapper for vectorized environment.
+    Frames are concatenated on desired axis (e.g. 4 RGB images -> 12 channels).
 
     :param venv: the vectorized environment to wrap
     :param n_stack: Number of frames to stack
+    :param channels_first: If True, stack on first image dimension, otherwise stack on the last (default)
     """
 
-    def __init__(self, venv: VecEnv, n_stack: int):
+    def __init__(self, venv: VecEnv, n_stack: int, channels_first: bool = False):
```
araffin, line 20: why not re-using you helper?
and maybe have a parameter `channels_order` that is optional and that can overwrite the result of the `is_channel_first()` helper?

araffin (commented):
Thanks, almost ready to merge. Some minor changes required.
And probably a separate test for `is_channel_first` (as the current implementation only works in some cases)

araffin:
>Should VecFrameStack automatically detect the axis to stack on or do we just rely on user input (this is hard to catch tho)?

see my comment ;)

>You mentioned predict() should be updated but it seems like it already check if shapes match, so no need to update it for this.

I actually forgot I did that ^^"

https://github.com/DLR-RM/stable-baselines3/blob/master/stable_baselines3/common/policies.py#L273

stable_baselines3/common/preprocessing.py
Miffyli: Oh wow proper brainderp here. Will fix this and add tests.

Miffyli:
That should resolve all the comments. Added tests for `is_image_space` while I was at it. 
Things got a touch complex (fragile?) for my liking, so if solutions seem too complex for the simple issues feel free to point out.

docs/misc/changelog.rst
```diff
@@ -6,22 +6,27 @@ Changelog
 Pre-Release 0.11.0a0 (WIP)
 -------------------------------
 
-
 Breaking Changes:
 ^^^^^^^^^^^^^^^^^
+- ``VecFrameStack`` now uses ``channels_order`` argument to tell if observations should be stacked
```
araffin, line 11: Is that really a breaking change as we are detecting the stack axis automatically?

araffin:
I updated the implementation, so it does not break things for non-image environments

tests/test_cnn.py
```diff
@@ -174,3 +176,71 @@ def test_features_extractor_target_net(model_class, share_features_extractor):
         params_should_match(original_actor_param, model.actor.parameters())
 
         td3_features_extractor_check(model)
+
+
+def test_channel_first_env(tmp_path):
+    # test_cnn uses environment with HxWxC setup that is transposed, but we
+    # also want to work with CxHxW envs directly without transpoing wrapper.
```
araffin, line 183:
```suggestion
    # also want to work with CxHxW envs directly without transposing wrapper.
```

docs/misc/changelog.rst
Miffyli: Some people (I know some) could be using FrameStackEnv on other-than-image-observations, like just on simple 1D observations, and with this update it will raise a an exception. Granted, this is not the case this stack-env is intended for but I can see why such use is tempting ^^

araffin (approved):
LGTM =)

PR #215, opened 2020-11-03
Add callable signatures to type annotations.

erniejunior:
## Description
I meant to review TD3 but ended up adding the signature to all callable type annotations in the codebase.

## Motivation and Context
Adding the signature to callables in the type annotations improves documentation and gives a much better idea what kind of function must be passed.

## Types of changes

## Checklist:

### DIFF
--- a/stable_baselines3/a2c/a2c.py
+++ b/stable_baselines3/a2c/a2c.py
@@ -1,4 +1,4 @@
-from typing import Any, Callable, Dict, Optional, Type, Union
+from typing import Any, Dict, Optional, Type, Union
 
 import torch as th
 from gym import spaces
@@ -7,7 +7,7 @@ from torch.nn import functional as F
 from stable_baselines3.common import logger
 from stable_baselines3.common.on_policy_algorithm import OnPolicyAlgorithm
 from stable_baselines3.common.policies import ActorCriticPolicy
-from stable_baselines3.common.type_aliases import GymEnv, MaybeCallback, LearningRateSchedule
+from stable_baselines3.common.type_aliases import GymEnv, LearningRateSchedule, MaybeCallback
 from stable_baselines3.common.utils import explained_variance
 
 
--- a/stable_baselines3/a2c/a2c.py
+++ b/stable_baselines3/a2c/a2c.py
@@ -55,7 +55,7 @@ class A2C(OnPolicyAlgorithm):
         self,
         policy: Union[str, Type[ActorCriticPolicy]],
         env: Union[GymEnv, str],
-        learning_rate: Union[float, Callable] = 7e-4,
+        learning_rate: Union[float, Callable[[float], float]] = 7e-4,
         n_steps: int = 5,
         gamma: float = 0.99,
         gae_lambda: float = 1.0,
--- a/stable_baselines3/a2c/a2c.py
+++ b/stable_baselines3/a2c/a2c.py
@@ -7,7 +7,7 @@ from torch.nn import functional as F
 from stable_baselines3.common import logger
 from stable_baselines3.common.on_policy_algorithm import OnPolicyAlgorithm
 from stable_baselines3.common.policies import ActorCriticPolicy
-from stable_baselines3.common.type_aliases import GymEnv, MaybeCallback
+from stable_baselines3.common.type_aliases import GymEnv, MaybeCallback, LearningRateSchedule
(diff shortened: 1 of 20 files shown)


erniejunior:
Note: I could not run `make pytest` locally but just executing pytest with my IDE worked just fine.

stable_baselines3/a2c/a2c.py
```diff
@@ -55,7 +55,7 @@ def __init__(
         self,
         policy: Union[str, Type[ActorCriticPolicy]],
         env: Union[GymEnv, str],
-        learning_rate: Union[float, Callable] = 7e-4,
+        learning_rate: Union[float, Callable[[float], float]] = 7e-4,
```
araffin, line 58: Maybe add that to the type aliases? (as it comes quite often)
for the name, I'm not sure, maybe `LearningRateOrSchedule`

stable_baselines3/common/utils.py
```diff
@@ -88,7 +88,7 @@ def get_schedule_fn(value_schedule: Union[Callable, float]) -> Callable:
     return value_schedule
 
 
-def get_linear_fn(start: float, end: float, end_fraction: float) -> Callable:
+def get_linear_fn(start: float, end: float, end_fraction: float) -> Callable[[float], float]:
```
araffin, line 91: same, maybe define a `Schedule` type alias

araffin (commented):
Just small remarks about new type aliases, other than that LGTM ;)

stable_baselines3/common/utils.py
erniejunior: I will go with `LearningRateSchedule` since we might have other kinds of schedules in the future.

araffin: in fact, for PPO, there is already the clipping ratio which could be a schedule (currently not implemented)
and in DQN, you have the exploration schedule for the epsilon-greedy policy

erniejunior:
What do you think about documenting the type aliases? This would be a nice place to state how the learning rate schedule work (inputs ranging from 0 to 1 etc). I could add it and also add the documentation from the `LearningRateSchedule` but I do not feel confident about documenting the other type aliases right now.

araffin:
>What do you think about documenting the type aliases?

Where do you want to put that? Comments in the file? or a file under `docs/` folder?
Yes, more documentation is always good ;)

stable_baselines3/sac/policies.py
```diff
@@ -16,6 +16,8 @@
 )
 
 # CAP the standard deviation of the actor
+from stable_baselines3.common.type_aliases import LearningRateSchedule
```
araffin, line 19: why is it not at the top?

stable_baselines3/td3/td3.py
```diff
@@ -21,7 +21,7 @@ class TD3(OffPolicyAlgorithm):
     Paper: https://arxiv.org/abs/1802.09477
     Introduction to TD3: https://spinningup.openai.com/en/latest/algorithms/td3.html
 
-    :param policy: The policy model to use (MlpPolicy, CnnPolicy, ...)
+    :param policy: The policy model to use (MlpPolicy, CnnPolicy, ...)  # TODO(max): is this true, the type annotations say otherwise
```
araffin, line 24: `MlpPolicy` is an alias of `TD3Policy` and `CnnPolicy` derives from `TD3Policy`

erniejunior:
I wanted to add a new type aliases section in common that just uses `..automodule::` to document the `type_aliases.py`. However I could not get it to work due to the following problem:

This is the docstring I use in `type_aliases.py`
```python
LearningRateSchedule = Callable[[float], float]
"""This is the learning rate schedule"""
```
However, this will only render `alias of typing.Callable` in HTML and not `This is the learning rate schedule` as expected. I only get it to render properly when I change the type alias to:
```python
LearningRateSchedule = Callable
```
without the signature. Do you have any idea for why this could be? I think I saw some issues where you were using some tricks to get the type annotations properly rendered in sphinx?

araffin:
>However, this will only render alias of typing.Callable in HTML and not This is the learning rate schedule as expected. I only get it to render properly when I change the type alias to:

I'm afraid there is no way around... For instance, I couldn't get it to work for the policies.
At least, it is documenting in the developer guide.

araffin (approved):
LGTM, thanks =)

PR #220, opened 2020-11-12
Use Monitor episode reward/length for `evaluate_policy`

### LINKED ISSUE #181: `evaluate_policy` reports preprocessed reward, whereas `rollout/ep_rew_mean` is unprocessed
**Describe the bug**

There are two kinds of reward that it might make sense to report:
  - The unwrapped reward. This is computed by the original environment. This defines the objective of the RL task, and what would usually be used to compute the policy return when e.g. benchmarking the resulting policy for papers.
  - The wrapped reward. This is the reward the RL algorithm actually uses for training, after any environment wrappers that mess with reward like `VecNormalize` or `AtariPreprocessing` have been applied. We hope the arg max of this is the same as the arg max of the original environment reward, but it need not be. Even if it is, they can vary substantially -- e.g. rescaling by a positive constant, or addition of shaping.

Currently, `common.evaluation.evaluate_policy` reports the wrapped reward. Most (all?) RL algorithms report the unwrapped reward, e.g. `OffPolicyAlgorithm._dump_logs` in `rollout/ep_rew_mean`. The difference is `common.evaluate.evaluate_policy` directly records reward and computes statistics; whereas the RL algorithms rely on the `"r"` and `"l"` keys inserted into the info-dict by `common.Monitor` which is usually the first wrapper applied to the environment (before any preprocessing).

In my opinion we would ideally report *both* types of reward (and episode length -- since wrappers can also affect `done`), in both the training and evaluation environment. But if we had to pick just one, I'd advocate for swapping the two -- report the reward used for training during training, and the reward usually used for evaluation for the evaluation environment.

Credit to @ejmichaud for first noticing this discrepancy.

**Code example**
This is visible in [train.py from Zoo](https://github.com/DLR-RM/rl-baselines3-zoo/blob/master/train.py) on Atari:
```
$ python train.py --algo dqn --env SeaquestNoFrameskip-v4
# [ ... elided ... ]
Eval num_timesteps=10000, episode_reward=52.00 +/- 48.33
Episode length: 527.20 +/- 139.00
New best mean reward!
----------------------------------
| eval/               |          |
|    mean_ep_length   | 527      |
|    mean_reward      | 52       |
| rollout/            |          |
|    ep_len_mean      | 2.16e+03 |
|    ep_rew_mean      | 87       |
|    exploration rate | 0.989    |
| time/               |          |
|    episodes         | 20       |
|    fps              | 717      |
|    time_elapsed     | 14       |
|    total timesteps  | 10698    |
----------------------------------
```
Note that not only do `eval/` and `rollout/` disagree on reward per timestep (because `AtariPreprocessing` does reward clipping), they also disagree on the episode length (because `AtariPreprocessing` converts loss of life into loss of an episode)! See [seaquest.log](https://github.com/DLR-RM/stable-baselines3/files/5347387/seaquest.log) for full log.

Note `train.py` does do some magic to disable reward normalization in `VecNormalize`. So this problem I think won't be visible in e.g. MuJoCo. Conceivably one could do something similar for `AtariWrapper` (set `terminal_on_life_loss=False` and `clip_reward=False`) -- but doing this for every wrapper seems error-prone, and some wrappers may just not support this out of the box.

Miffyli:
## Description
Design choice: Checking for Monitor wrapper reliably for both envs and vecenvs got tricky and messy, so instead I opted for "lazily" checking if the `"episode"` information is available in info and assume it is from a Monitor wrapper.

## Motivation and Context
closes #181 

## Types of changes

## Checklist:

### DIFF
--- a/docs/misc/changelog.rst
+++ b/docs/misc/changelog.rst
@@ -19,7 +19,7 @@ New Features:
   on the first or last observation dimension (originally always stacked on last).
 - ``common.env_util.is_wrapped`` and ``common.env_util.unwrap_wrapper`` functions for checking/unwrapping
   an environment for specific wrapper.
-- ``env_is_wrapped`` function for ``SubProcVecEnv`` for checking if its environments are wrapped
+- ``env_is_wrapped`` function for ``VecEnv`` for checking if its environments are wrapped
   with given wrappers.
 
 Bug Fixes:
--- a/docs/misc/changelog.rst
+++ b/docs/misc/changelog.rst
@@ -16,6 +16,7 @@ New Features:
   automatic check for image spaces.
 - ``VecFrameStack`` now has a ``channels_order`` argument to tell if observations should be stacked
   on the first or last observation dimension (originally always stacked on last).
+- ``evaluate_policy`` now returns rewards/episode lengths from a Monitor wrapper if one is present.
 
 Bug Fixes:
 ^^^^^^^^^^
--- a/docs/misc/changelog.rst
+++ b/docs/misc/changelog.rst
@@ -34,6 +34,7 @@ Deprecations:
 Others:
 ^^^^^^^
 - Add more issue templates
+- Add signatures to callable type annotations (@erniejunior)
 - Improve error message in ``NatureCNN``
 
 Documentation:
--- a/docs/misc/changelog.rst
+++ b/docs/misc/changelog.rst
@@ -8,8 +8,10 @@ Pre-Release 0.11.0a0 (WIP)
 
 Breaking Changes:
 ^^^^^^^^^^^^^^^^^
-- ``evaluate_policy`` now returns rewards/episode lengths from a Monitor wrapper if one is present.
-
+- ``evaluate_policy`` now returns rewards/episode lengths from a ``Monitor`` wrapper if one is present,
+  this allows to return the unnormalized reward in the case of Atari games for instance.
+- Renamed ``common.vec_env.is_wrapped`` to ``common.vec_env.is_vecenv_wrapped`` to avoid confusion
+  with the new ``is_wrapped()`` helper
 
 New Features:
 ^^^^^^^^^^^^^
@@ -17,10 +19,10 @@ New Features:
   automatic check for image spaces.
 - ``VecFrameStack`` now has a ``channels_order`` argument to tell if observations should be stacked
   on the first or last observation dimension (originally always stacked on last).
-- ``common.env_util.is_wrapped`` and ``common.env_util.unwrap_wrapper`` functions for checking/unwrapping
+- Added ``common.env_util.is_wrapped`` and ``common.env_util.unwrap_wrapper`` functions for checking/unwrapping
   an environment for specific wrapper.
-- ``env_is_wrapped`` function for ``VecEnv`` for checking if its environments are wrapped
-  with given wrappers.
+- Added ``env_is_wrapped()`` method for ``VecEnv`` to check if its environments are wrapped
+  with given Gym wrappers.
 
 Bug Fixes:
 ^^^^^^^^^^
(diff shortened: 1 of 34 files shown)


docs/misc/changelog.rst
```diff
@@ -16,6 +16,7 @@ New Features:
   automatic check for image spaces.
 - ``VecFrameStack`` now has a ``channels_order`` argument to tell if observations should be stacked
   on the first or last observation dimension (originally always stacked on last).
+- ``evaluate_policy`` now returns rewards/episode lengths from a Monitor wrapper if one is present.
```
araffin, line 19: that is a breaking change, no?

stable_baselines3/common/evaluation.py
```diff
@@ -49,15 +58,32 @@ def evaluate_policy(
...
             if render:
                 env.render()
-        episode_rewards.append(episode_reward)
-        episode_lengths.append(episode_length)
+
+        # Remove VecEnv stacking (if any)
+        if isinstance(env, VecEnv):
+            info = info[0]
+
+        if "episode" in info.keys():
```
araffin, line 73: did you try that on atari games?
I'm afraid it will have some weird behavior because the wrapper makes "end of life = end of episode", so you will have a mix of normalized and unnormalized reward, no?

One solution would be to deactivate that behavior when testing atari games (can be done in the zoo).

stable_baselines3/common/evaluation.py
```diff
@@ -49,15 +58,32 @@ def evaluate_policy(
...
+            # Monitor wrapper includes "episode" key in info if environment
+            # has been wrapped with it. Use those rewards instead.
+            episode_rewards.append(info["episode"]["r"])
+            episode_lengths.append(info["episode"]["l"])
+        else:
+            episode_rewards.append(episode_reward)
+            episode_lengths.append(episode_length)
+            warnings.warn(
```
araffin, line 81: we should only warn once, no?

araffin (commented):
Some point we need to discuss ;)

stable_baselines3/common/evaluation.py
Miffyli: Ah right, I felt like I forgot something when cooking this up... Good thing you mentioned this before I started running my Atari experiments where I use this ^^.

I would avoid Atari-specific tweaking and just go straight for bit more generic solution. I think a reliable check for ``Monitor`` wrapper would do the trick (need to think more on this).

stable_baselines3/common/evaluation.py
Miffyli: `warnings`, by default, suppresses further warnings from same code location so only one is shown, but we can add a simple if-check if you feel that would be less stressful for machine.

Miffyli:
CI will fail because I could not get typing stuff to work out (e.g. Monitor's). There also starts being a ton of circular imports which are _suuuuuper_-fun to deal with. I think making whole `type_aliases.py` file only "active" during typing would help in future, if that is possible in any way. 

@araffin Could you look into the typing things if the current things are alright and fix where necessary?

araffin:
>here also starts being a ton of circular imports which are suuuuuper-fun to deal with.

No worry, I will take a look ;) (i had some fun with that in the past)

araffin:
@Miffyli done ;)

but there some warnings not catched now...
one solution would be to add an argument like for the env checker (`warn=True`)
```
=============================== warnings summary ===============================
tests/test_callbacks.py: 6 warnings
tests/test_identity.py: 12 warnings
tests/test_spaces.py: 6 warnings
tests/test_utils.py: 3 warnings
tests/test_vec_normalize.py: 3 warnings
  /builds/araffin/stable-baselines3/stable_baselines3/common/evaluation.py:66: UserWarning: Evaluation environment is not wrapped with a ``Monitor`` wrapper. This may result in reporting modified episode lengths and rewards, if other wrappers happen to modify these. Consider wrapping environment first with ``Monitor`` wrapper.
    UserWarning,
```

>le only "active" during typing would help in future, if that is possible in any way.

the issue is then the automatic documentation which would fail (yes it is a mess ^^#)

Miffyli:
Thanks a ton! Now things should be in order :)

araffin (approved):
LGTM =)

PR #222, opened 2020-11-15
Fix DQN predict shape for single Gym env

### LINKED ISSUE #221: Why would the predict method of the DQN model return an array of multiple indices for the next action?
I am not yet able to reproduce this error with a simple example (given that I am not using a gym environment), but, at some point, the `predict` method is returning an array as the predicted action, while it does not return anything for the next state/observation, i.e., at some point, the following statement

```
action, _states = model.predict(next_observation)
```

returns `action = [0 1 0 0]` (whose type is `numpy.ndarray`) and `_states = None`. Why would this be the case?

I am using 

```
from stable_baselines3.dqn.dqn import DQN
from stable_baselines3.dqn.policies import MlpPolicy

model = DQN(MlpPolicy, env, verbose=1, exploration_fraction=0.1, tensorboard_log="some_folder")
model.learn(total_timesteps=10000)
```

Note that, before this happening, `model.predict` correctly returns only one index, which represents only 1 action, but `_states` is always `None` (not sure why too).

By looking at your implementation [here](https://github.com/DLR-RM/stable-baselines3/blob/master/stable_baselines3/dqn/dqn.py#L207), this should happen if `not deterministic and np.random.rand() < self.exploration_rate:`. Now, in my case, I don't set the parameter `deterministic`, whose default value is `False`, so `not deterministic` is true, so what is causing the problem is `np.random.rand() < self.exploration_rate`. The thing is: I never experienced this when I was playing with stable-baselines 2. Also, I don't understand your logic in that implementation anyway. Why would `n_batch = observation.shape[0]` and then sample multiple actions? I guess that if I set `deterministic=True`, then I will not experience this problem anymore. 

Should `deterministic` always be True after learning?

araffin:
## Motivation and Context

closes #221 

## Types of changes

## Checklist:

### DIFF
--- a/docs/misc/changelog.rst
+++ b/docs/misc/changelog.rst
@@ -20,6 +20,7 @@ New Features:
 Bug Fixes:
 ^^^^^^^^^^
 - Fixed bug where code added VecTranspose on channel-first image environments (thanks @qxcv)
+- Fixed ``DQN`` predict method when using single ``gym.Env`` with ``deterministic=False``
 
 Deprecations:
 ^^^^^^^^^^^^^
--- a/stable_baselines3/common/policies.py
+++ b/stable_baselines3/common/policies.py
@@ -3,7 +3,7 @@
 import collections
 from abc import ABC, abstractmethod
 from functools import partial
-from typing import Any, Callable, Dict, List, Optional, Tuple, Type, Union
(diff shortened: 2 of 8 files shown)


Miffyli (approved):
LGTM!

Issue #225, opened 2020-11-16
[Bug] arguments order error in explained_variance()


thisray:
Hi,

The calculation method of explained variance in the comment is: `1 - Var[y-ypred] / Var[y]`. In the scenario of value loss, I think the `y_true` is `return` and `y_pred` is `value`.

The calculation of explained variance in ppo.py:

https://github.com/DLR-RM/stable-baselines3/blob/master/stable_baselines3/ppo/ppo.py#L223
```
explained_var = explained_variance(self.rollout_buffer.returns.flatten(), self.rollout_buffer.values.flatten())
```

But the arguments in this function:

https://github.com/DLR-RM/stable-baselines3/blob/master/stable_baselines3/common/utils.py#L42
```
def explained_variance(y_pred: np.ndarray, y_true: np.ndarray) -> np.ndarray:
```

It seems the order is not right?

thanks!

Miffyli:
You are right, it seems to be the wrong way around. @araffin Can you double-confirm this, as you seem to be one who touched the code last? 

In case this is a bug, we need to calls to `explained_variance` for typos like this.

araffin:
Hello,
nice catch ;) (the good news is that this variable is only used for debugging purposes)

thisray:
Yes, it just a little bug. Many thanks for this library, it is very useful and make me learn a lot!

araffin:
>Yes, it just a little bug. Many thanks for this library, it is very useful and make me learn a lot!

I edited my message a bit late, but we would be happy if you could submit a PR that solves this issue ;)

PR #226, opened 2020-11-16
Fix for the arguments order in explained_variance

thisray:
This commit is for fixing the issue: https://github.com/DLR-RM/stable-baselines3/issues/225
Change: `explained_variance(y_pred, y_true)` -> `explained_variance(y_true, y_pred)`
This function is used in `ppo.py` and `a2c.py`. And I think this format (`true, pred`) is much like sklearn and torch style.

### DIFF
--- a/stable_baselines3/common/utils.py
+++ b/stable_baselines3/common/utils.py
@@ -39,7 +39,7 @@ def set_random_seed(seed: int, using_cuda: bool = False) -> None:
 
 
 # From stable baselines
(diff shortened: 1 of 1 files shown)


Miffyli:
Thanks for the PR! Please fill in the PR template and do the steps detailed there :).

thisray:
Thanks for your explanation, I will check the PR template.
I have checked there are only `ppo.py` and `a2c.py` using this `explained_variance()` function.
It seems changing the argument order in  `ppo.py` and `a2c.py` is better. I will do it latter.

araffin:
closing this in favor of #227 ;)

PR #239, opened 2020-11-22
Fix for saving big replay buffer, use pickle protocol>=4

### LINKED ISSUE #238: OverflowError : cannot serialize a bytes object larger than 4 GiB when saving replay buffer
python --version
Python 3.6.11
torch 1.6.0+cu101

The solution is probably this:
Use pickle with protocol=4, e.g.,
pickle.dump((X,c), f, protocol=4)

araffin:
## Motivation and Context

closes #238 

## Types of changes

## Checklist:

### DIFF
--- a/docs/misc/changelog.rst
+++ b/docs/misc/changelog.rst
@@ -32,6 +32,7 @@ Bug Fixes:
 - Fixed ``DQN`` predict method when using single ``gym.Env`` with ``deterministic=False``
 - Fixed bug that the arguments order of ``explained_variance()`` in ``ppo.py`` and ``a2c.py`` is not correct (@thisray)
 - Fixed bug where full ``HerReplayBuffer`` leads to an index error. (@megan-klaiber)
+- Fixed bug where replay buffer could not be saved if it was too big (> 4 Gb) for python<3.8 (thanks @hn2)
 
 Deprecations:
 ^^^^^^^^^^^^^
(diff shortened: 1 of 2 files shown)


Miffyli (approved):
Code checks out (not much to check :) ), however I would add a note to improve this implementation. Pickle is usually regarded slow (and not meant for such high-data-storage). I think buffers should be easy to store with numpy saves instead, which (could?) save some space with the compression and also make things less prone on inter-Python problems (like with the save formats a year ago).

araffin:
>Letting you merge this to see this comment ^^.

Alright ^^
Then I'm a bit confused, do you want me to add that note somewhere?

My personal take on that: most of the time, you will not save the replay buffer.
Yes, we could use numpy to save most of it, but again, we would face some issues later (when supporting dict for instance, or n-step replay).
Yes, pickle is slow but I don't see any use case for now where you need to save the replay buffer very frequently.

>inter-Python problems (like with the save formats a year ago).

if you use "raw" pickle, inter python is not an issue. The issue comes when you are using cloudpickle.

Miffyli:
> Then I'm a bit confused, do you want me to add that note somewhere?
Ah sorry. I meant to take a mental note on this, nothing bigger. If it works for now let it work, and next time it is fiddled it we can consider moving to numpy things.

Fair points on using pickle instead though, especially considering how in future things will get more complex.

PR #243, opened 2020-11-24
Dictionary Observations

### LINKED ISSUE #216: [Feature Request] basic dict/tuple support for observations
### 🚀 Feature
As mentioned in the [RoadMap](https://github.com/DLR-RM/stable-baselines3/issues/1), adding dict/tuple support for observations is a planned feature. This follows from the OpenAI gym api which has Tuple and Dict as possible observation spaces.

### Motivation

Currently, stablebaselines3 only supports one (image or a vector) observation. Extending this to Tuple/Dict observations would support for environments which have different inputs of data. 

### Current Plan
I plan on implementing this feature but I'd like to have some pointers on how to go about it.
Below is my current plan but I'd really like to verify it as a good way forward.

I think that I need to create a child class of `RolloutBufferSamples` which stores a list/dict of observations rather than a single observation. 

However, this may require adding a bool on the `rollout_buffer` itself so that the conversion to tensor ([see on_policy_algorithm.py](https://github.com/DLR-RM/stable-baselines3/blob/master/stable_baselines3/common/on_policy_algorithm.py#L155)), can be performed over each element of the list/dict.

From here, I think that the other necessary changes would permeate through the repository:
- add a "CombinedExtractor" in `torch_layers.py` that can take in multiple observations.
- add a new Policy for each algorithm to use the new extractors
- modify `util.py` and `preprocessing.py` to handle the new rollout type

Is this a good approach?

### LINKED ISSUE #284: [Bug] Infinize horizon tasks are handled like episodic tasks
Hi,
I wonder how to correctly use SAC with infinite horizon environments. I saw @araffin s answer to https://github.com/hill-a/stable-baselines/issues/776 where he points out that algorithms are step-based. Our environments could always return `done = False`, but we would have to reset the environment manually then. As a consequence, we would add transitions to the replay buffer going from the last state to the initial state, which is bad. 

Is the only solution to include a time-feature? That means messing with the `observation_space` size and handling dict spaces correctly + explaining what this "time-feature" is in papers. Let me know if I've missed a thread treating this issue already :smile:  
Greetings!

### 🐛 Bug / Background
My understanding is that SAC skips the target if `s'` is a terminal state:
```python
q_backup = replay_data.rewards + (1 - replay_data.dones) * self.gamma * target_q
```
In infinite horizon tasks, we wrap our env with `gym.wrappers.TimeLimit`, which sets `done = True` when the maximum episode length is reached. This stops the episode in SAC and the transition is saved in the replay buffer for learning. 

However, according to "Time Limits in Reinforcement Learning" (https://arxiv.org/abs/1712.00378), we should not see that last state as a "terminal" state, since the termination has nothing to do with the MDP. If we ignore this, we are doing "state aliasing" and violating the Markov Property.

### LINKED ISSUE #287: [Question] Using images to train DDPG+HER agent
Hello,
I’m trying to train a DDPG+HER agent that interacts with a custom environment and that takes in input the RGB image of the environment.
From what I understand, in the previous version of stable baselines only 1D observation spaces were supported in HER (as indicated also in **HERGoalEnvWrapper**), thus excluding image observations

           if len(goal_space_shape) == 2:
                assert goal_space_shape[1] == 1, "Only 1D observation spaces are supported yet"
            else:
                assert len(goal_space_shape) == 1, "Only 1D observation spaces are supported yet"

In this new version of stable baselines, I see no evident assertion against using 2D spaces but in **ObsDictWrapper**, the observation and goal dimensions are taken from the first dimension shape only

        if isinstance(self.spaces[0], spaces.Discrete):
            self.obs_dim = 1
            self.goal_dim = 1
        else:
            self.obs_dim = venv.observation_space.spaces["observation"].shape[0]
            self.goal_dim = venv.observation_space.spaces["achieved_goal"].shape[0]

### Question
Is it possible to train a DDPG+HER agent from images using the implementation of stable baselines 3?

J-Travnik:
In machine learning, input comes in the form of matrices. Typically, models take in 1 matrix at a time, such as in the case of image classification where matrix containing the input image is given to the model and the model classifies the image.
However, there are many situations in which taking multiple inputs is necessary. One example is when training a reinforcement learning agent and the observations that the agent sees comes in the form of an image (e.g., camera, grid sensor, etc) and a vector describing the agent's state (e.g., current position, health, etc). In this situation, it is necessary to feed 2 inputs to the model. This PR addresses this.

## Description
* added example environments with multi-input observations
* added `DictReplayBuffer` and `DictRolloutBuffer` to handle dictionaries
* added `CombinedExtractor` feature extractor that handles generic dictionary data
* added `StackedObservations` and `StackedDictObservations` to decouple data stacking from the `VecFrameStack` wrapper
* added `test_dict_env.py` test
* added a `is_vectorized_env()` method per observation space type in `common\utils.py`

## Motivation and Context
- [x] I have raised an issue to propose this change ([link](https://github.com/DLR-RM/stable-baselines3/issues/216))

closes #216 

closes #287 (image support for HER)
closes #284 (for off-policy algorithms)

## Types of changes

## Checklist:

- [x] I have updated the tests accordingly
- [x] I have reformatted the code using `make format`
- [x] I have checked the codestyle using `make check-codestyle` and `make lint`

## TODOs
- [x] check that documentation is properly updated
- [x] check that dict with vectors only is the same as mlp policy + vector flatten
- [x] Update env checker
- [x] (optional) refactor HER: https://github.com/DLR-RM/stable-baselines3/tree/feat/refactor-her
- [x] test A2C/PPO/SAC alone with GoalEnv


Miffyli:
Quick comment before further commits: Make sure your formatter has line lengths set to 127, not 80. Seems like your formatter is splitting lines of allowed length (e.g. [this one](https://github.com/DLR-RM/stable-baselines3/pull/243/files#diff-fe0422470391efdc71a3583fbfb041b2b35d8399288bb5d2a3802002b1708061L170)).

J-Travnik:
@araffin 
> After a quick look, I'm not sure we need that much complexity.
> You could probably use the BitFlippingEnv together with a small dummy Custom Env.

Thats a good point though, I did want to make sure that it worked with Images though as that was my main use case. I think its important to cover this case (at least in a test) to make sure that VecEnvStacking and multiple environments all played well together when using images. Thoughts?

> it would nice to have at least one common "combined extractor " for the Box.

I agree though I still see value in having the option to separate them. Perhaps this could come in the form of a bool with true default of "auto_concatenate_box_obs". 

> After a quick look at your PR, I think it would be better to almost duplicate the buffer classes, because having too many branches (with the if else) is hard to follow.

Hmm, that seems like a decent idea. Specifically you mean to create:

```
class DictReplayBuffer(ReplayBuffer):...

class DictRolloutBuffer(RolloutBuffer):...
```
?


>I can run it on my side if you want and push the results. But it would be better for you to setup black and isort at least (you can take a look at the makefile for the details of the commands).

Yea, that'd be awesome. I will figure out black and stuff on my end. @Miffyli thanks for the pointer!

araffin:
> I think its important to cover this case (at least in a test) to make sure that VecEnvStacking and multiple environments all played well together when using images. Thoughts?

two things: if you want to test performances, then you need your environment (but we may skip the test if it is slow to run).
If you want only to check that the code run, you can use a dummy env that outputs dummy images like this one: https://github.com/DLR-RM/stable-baselines3/blob/master/stable_baselines3/common/identity_env.py#L107
And of course, we want to check that it runs at least with different type of input (otherwise we could just use a "flatten" wrapper).

So, yes, we need to test performance to validate the implementation, but I would not run the test for each commit (maybe use `pytest.mark.slow` to skip it).

>. Perhaps this could come in the form of a bool with true default of "auto_concatenate_box_obs".

sounds like a good compromise.

>Hmm, that seems like a decent idea. Specifically you mean to create:

yes!

Miffyli:
> I agree though I still see value in having the option to separate them. Perhaps this could come in the form of a bool with true default of "auto_concatenate_box_obs".

After bit of internal discussion we can leave this out for now and, at least initially, tell people to do it on the environment side (or alternatively provide some wrapper that concatenates arrays of specific keys together). Less messy code to worry about (e.g. images of different resolution).

Bit of a heads up: This update would be very desired, but being a big change there will be bunch of reviewing and comments ^^

J-Travnik:
@Miffyli Yea, no worries. I'd rather take the time and to review and fix changes to figure out something that is stable.

And fair enough on the "auto_concatenate_box_obs". Potentially it can take the form of a VecEnv wrapper so that the tensors passed to the `CombinedExtractor` can be used readily.

H-Park:
Tentative timeframe of this getting merged? I need this feature to start the literature grounded action masking PR

Miffyli:
@H-Park We will first do first release (v1.0) properly before merging this, soooo most daring estimate is just before end of the year, but more likely in the beginning of January.

 On preliminary glances this looks overall the way that things will work, so I would semi-reliably say you can base your work on how things work in @J-Travnik's branch :).

araffin:
@J-Travnik I reformated the code and corrected one typing issue in preprocessing, however, it complains about `VecFrameStack` now
I also fixed the check for HER, so the tests for HER should pass now ;)

@H-Park there is no deadline for this feature (at least after v1.0), it will take some time to have a working and clean solution.
But in the meantime, you could start working on @J-Travnik branch ;) (even though it may change).

multi_input_tests.py
```diff
@@ -0,0 +1,87 @@
+import argparse
```
araffin, line 1: this file should be removed at the end no?

Also, I recommend you to use `pytest -k keyword` to run only part of the test.

stable_baselines3/common/policies.py
```diff
@@ -260,14 +275,29 @@ def predict(
         #     state = self.initial_state
         # if mask is None:
         #     mask = [False for _ in range(self.n_envs)]
-        if isinstance(observation, dict):
+        # Need to check the observation if its a ObsDictWrapper
+
+        if isinstance(self.observation_space, ObsDictWrapper):
```
araffin, line 280: you should use `is_vec_env_wrapped()`

stable_baselines3/common/policies.py
```diff
@@ -278,12 +308,18 @@ def predict(
...
                     observation = transpose_obs
+        else:
+            observation = np.array(observation)
 
         vectorized_env = is_vectorized_observation(observation, self.observation_space)
 
-        observation = observation.reshape((-1,) + self.observation_space.shape)
+        if isinstance(observation, dict):
+            for key, obs in observation.items():
+                observation[key] = th.as_tensor(observation[key]).to(self.device)
```
araffin, line 318: don't have an helper for that now?

stable_baselines3/common/preprocessing.py
```diff
@@ -117,6 +133,9 @@ def get_obs_shape(observation_space: spaces.Space) -> Tuple[int, ...]:
     elif isinstance(observation_space, spaces.MultiBinary):
         # Number of binary features
         return (int(observation_space.n),)
+    elif isinstance(observation_space, spaces.Dict):
+        return {key: subspace.shape for (key, subspace) in observation_space.spaces.items()}
```
araffin, line 137: I think it should be a recursive call, no?

stable_baselines3/common/base_class.py
```diff
@@ -197,7 +197,9 @@ def _wrap_env(env: GymEnv, verbose: int = 0, monitor_wrapper: bool = True) -> Ve
             env = VecTransposeImage(env)
 
         # check if wrapper for dict support is needed when using HER
-        if isinstance(env.observation_space, gym.spaces.dict.Dict):
+        if isinstance(env.observation_space, gym.spaces.dict.Dict) and set(env.observation_space.spaces.keys()) == set(
```
araffin, line 200: note for later: we should probably check that this is a "first-level" dict and throw an error if there is more than one level
The env checker should be updated too

multi_input_tests.py
J-Travnik: I can convert this file to be a test and place it with the rest of the tests and label it `pytest.mark.slow` as suggested earlier. Would that be a good idea going forward?

stable_baselines3/common/base_class.py
J-Travnik: I wonder if we need this check at all. Once the DictBuffers are in, would it make the ObsDictWrapper redundant?

J-Travnik:
I added `DictReplayBuffer` and `DictRolloutBuffer` so that there is only one "if" statement in the policy to choose which one. They are child classes of their respective parents so the typing is preserved throughout the code base.
I'll go back and update the documentation for them probably early next week.

`test_offpolicy_normalization` fails now but only for HER and I am not completely sure why yet but I feel like there is an assumption of having ObsDictWrapper somewhere.

Going forward, what should the usage of the ObsDictWrapper be if dict observations can be handled by buffers?
I think there are some functions in the ObsDictWrapper that can be moved to a util script but in general I am not sure.

multi_input_tests.py
araffin: yes!

stable_baselines3/common/base_class.py
araffin: I meant that as an independent check, your current implementation only supper first-level dict, no?

`ObsDictWrapper` is meant to disappear in the future (when we refactor HER), but let's keep that update separate as this PR is already quite big.

stable_baselines3/common/buffers.py
```diff
@@ -396,3 +449,288 @@ def _get_samples(self, batch_inds: np.ndarray, env: Optional[VecNormalize] = Non
...
+class DictReplayBuffer(ReplayBuffer):
+    """
+    Replay buffer used in off-policy algorithms like SAC/TD3.
+
+    :param buffer_size: Max number of element in the buffer
+    :param observation_space: Observation space
+    :param action_space: Action space
+    :param device:
+    :param n_envs: Number of parallel environments
+    :param optimize_memory_usage: Enable a memory efficient variant
```
araffin, line 463: I would drop `optimize_memory_usage` support for DictReplayBuffer, no?
It feels like it adds quite some complexity.

Miffyli: I agree we should keep things as simple as possible first, and then later add any further optimizations/features once base implementations.

stable_baselines3/common/policies.py
```diff
@@ -260,14 +275,29 @@ def predict(
...
-        else:
-            observation = np.array(observation)
+        elif isinstance(observation, dict):
+            # need to copy the dict as the dict in VecFrameStack will become a torch tensor
+            observation = copy.deepcopy(observation)
+            for key, obs in observation.items():
+                obs_space = self.observation_space.spaces[key]
+                if is_image_space(obs_space):
+                    obs = np.array(obs)
+                    if not (obs.shape == obs_space.shape or obs.shape[1:] == obs_space.shape):
```
araffin, line 289: looks like duplicated code from above... maybe a separate (private) helper would be better?

stable_baselines3/common/vec_env/vec_frame_stack.py
```diff
@@ -28,68 +28,166 @@ def __init__(self, venv: VecEnv, n_stack: int, channels_order: Optional[str] = N
...
-        for i, done in enumerate(dones):
-            if done:
-                if "terminal_observation" in infos[i]:
-                    old_terminal = infos[i]["terminal_observation"]
-                    if self.channels_first:
-                        new_terminal = np.concatenate(
-                            (self.stackedobs[i, :-stack_ax_size, ...], old_terminal), axis=self.stack_dimension
-                        )
+
+        if isinstance(self.venv.observation_space, spaces.Box):
```
araffin, line 93: Again, looks like a lot of branching... maybe it would also make sense to create a separate wrapper for clarity?

stable_baselines3/common/base_class.py
J-Travnik: Yea, I only support first-level dicts. Ok, I understand :)

stable_baselines3/common/buffers.py
J-Travnik: I can default `optimize_memory_storage` to False for now for it to be implemented/verified later. Sound like a plan?

stable_baselines3/common/policies.py
J-Travnik: True, I created `try_transpose_img_observation(observation, observation_space)` on `BasePolicy` but we can move it to a util script later.

stable_baselines3/common/vec_env/vec_frame_stack.py
J-Travnik: Although I agree its a lot of branching, if there was a separate wrapper would it mean that we'd have to keep track of whether the environment is a `VecFrameStack` as opposed to a `DictVecFrameStack`?.

Additionally, from a usability point of view, assuming that Dicts are supported "natively" because the correct Buffer class is chosen based on the observation space (e.g., `DictReplayBuffer`), I would expect that `VecFrameStack` to have similar functionality so as to avoid:
```
env = ...
if isinstance(env.observation_space, gym.spaces.Dict):
    env = DictVecFrameStack(env, n_stack=4)
else:
    env = VecFrameStack(env, n_stack=4)
```

Is there interest in an alternative that doesn't require an another wrapper class?

Potentially there could be a class that holds the data of `VecFrameStack.stacked_obs` and `VecFrameStack` can call methods upon the data. Thoughts?

araffin:
quick remark: black is run with a line length of 127 (cf contrib guide and PR template)
also, please use the latest version of it

stable_baselines3/common/vec_env/vec_frame_stack.py
araffin: Good point.
Maybe as an intermediate solution we could have private implementations for each case and `VecFrameStack` just do:

```python
class VecFrameStack(VecEnvWrapper):
    def __init__(self, venv: VecEnv, n_stack: int, channels_order: Optional[str] = None):
        if isinstance(venv.observation_space, spaces.Box):
            return VecFrameStackBox(venv, n_stack, channels_order)

        elif isinstance(wrapped_obs_space, spaces.Dict):
            return VecFrameStackDict(venv, n_stack, channels_order)
        else:
            raise Exception("VecFrameStack only works with gym.spaces.Box and gym.spaces.Dict observation spaces")
```

araffin: btw, do we want to stack every key (then we would need to check that each key is a Box) or allow to stack only some keys?

J-Travnik: > we could have private implementations for each case ...

I don't think returning the private implementations like this will work. I think there are (and if not there could be) checks to see if the environment has a `VecFrameStack` wrapping. In this case, both `VecFrameStackBox` and `VecFrameStackDict` would have to be child classes of `VecFrameStack` which is confusing if their parent `VecFrameStack` depends upon them. Additionally the `__init__` method does not return anything so this would cause a `TypeError`. 

As an alternative, I was imagining having a private class of `StackedObservations` that wraps the data. Something like:

```
class StackedObservations:
    def __init__(self, num_stacks, observation_space, channels_order):
        ...

    def reset(self, observation):
        ...
        return self.stackedobs

    def update(self, observations, dones, info):
        ....
        return self.stackedobs, info

class StackedDictObservations(StackedObservations):
    ...
```

This would replace `compute_stacking` completely avoiding code duplication and enabling reuse `StackedObservations` outside of  `VecFrameStack` and could be used as:

```
class VecFrameStack(VecEnvWrapper):
    def __init__(self, venv: VecEnv, n_stack: int, channels_order: Optional[str] = None):
        if isinstance(venv.observation_space, spaces.Box):
            self.stacked_observations= StackedObservations(num_stacks, observation_space, channels_order)
        elif isinstance(venv.observation_space, spaces.Dict):
            self.stacked_observations= StackedDictObservations(num_stacks, observation_space, channels_order)
        else:
            raise Exception("VecFrameStack only works with gym.spaces.Box and gym.spaces.Dict observation spaces")

    def step_wait(self):
        observations, rewards, dones, infos = self.venv.step_wait()
        stacked_obs, infos = self.stacked_observations.update(observations, dones, infos)
        return stacked_obs, rewards, dones, infos

    def reset(self):
        observation = self.venv.reset()
        return self.stacked_observations.reset(observation)
```

In the future, this allows for custom stacking techniques but I do think that by default we should stack every key. Generic implementations are more easily customized than selective ones IMO. 

Thoughts?

araffin: >In the future, this allows for custom stacking techniques but I do think that by default we should stack every key. Generic implementations are more easily customized than selective ones IMO.
>Thoughts?

>As an alternative, I was imagining having a private class of StackedObservations that wraps the data. Something like:

Looks like a good compromise ;)

J-Travnik: Deal ;)

J-Travnik: Put this in the latest commit. I still use `compute_stacking` within the new `StackedObservation` class to reduce code duplication. I think it really cleans up `VecFrameStack` . Perhaps one can pass in their own kind of stacking class to `VecFrameStack` but I don't really want the PR to get even bigger than it is so this can be done in the future if desired.

araffin:
@J-Travnik I fixed all typing issues, the CI should pass now ;).

Please fix all the warnings from the linter (you can take a look a the CI output but you can also run it locally).
Please also revert back all the formatting changes (adding trailing commas) that you have done on part of the code that it not related to this PR ;).

docs/misc/changelog.rst
```diff
@@ -8,6 +8,7 @@ Pre-Release 0.11.0a1 (WIP)
 
 Breaking Changes:
 ^^^^^^^^^^^^^^^^^
+- Breaks HER as HER needs to be updated to use the new dictionary observations
```
araffin, line 11: this is not the case anymore ;)

stable_baselines3/common/multi_input_envs.py
```diff
@@ -0,0 +1,193 @@
...
+    # 7         Vec * 2/3   Img * 0
+    # 8         Vec * 2/3   Img * 1/3
+    # 9         Vec * 2/3   Img * 2/3
+    # 10        Vec * 2/3   Img * 3/3
+    # 11        Vec * 3/3   Img * 0
+    # 12        Vec * 3/3   Img * 1/3
+    # 13        Vec * 3/3   Img * 2/3
+    # 14        Vec * 3/3   Img * 3/3
+
+    def __init__(self, num_col=4, num_row=4, random_start=True, noise=0.0):
```
araffin, line 38: please use type hints

stable_baselines3/common/multi_input_envs.py
```diff
@@ -0,0 +1,193 @@
...
+        self.state = 0
+        self.action2str = ["left", "down", "right", "up"]
+        self.noise = noise
+        self.init_possible_transitions()
+
+        self.init_state_mapping(num_col, num_row)
+
+        self.max_state = len(self.state_mapping) - 1
+
+    def random_upsample_img(self, v_rng=(0, 255), initial_size=(4, 4), up_size=(20, 20)):
```
araffin, line 64: missing docstrings and type hints

stable_baselines3/common/multi_input_envs.py
```diff
@@ -0,0 +1,193 @@
+import gym
+import numpy as np
+
+
+class SimpleMultiObsEnv(gym.Env):
```
araffin, line 5: maybe we should put all custom envs in an env folder...
What do you think @Miffyli ?

stable_baselines3/common/off_policy_algorithm.py
```diff
@@ -154,13 +154,23 @@ def __init__(
...
         self.set_random_seed(self.seed)
-        self.replay_buffer = ReplayBuffer(
-            self.buffer_size,
-            self.observation_space,
-            self.action_space,
-            self.device,
-            optimize_memory_usage=self.optimize_memory_usage,
-        )
+
+        if isinstance(self.observation_space, gym.spaces.Dict):
```
araffin, line 158: looks like duplicate code... probably using the if/else to choose the class and then pass the same parameters to the constructors

stable_baselines3/common/off_policy_algorithm.py
```diff
@@ -431,7 +465,11 @@ def collect_rollouts(
                         reward_ = self._vec_normalize_env.get_original_reward()
                     else:
                         # Avoid changing the original ones
-                        self._last_original_obs, new_obs_, reward_ = self._last_obs, new_obs, reward
+                        self._last_original_obs, new_obs_, reward_ = (
```
araffin, line 468: as mentioned in my comments in the main, please revert such formatting change (where a trailing comma was added) on code not related to your PR ;)

stable_baselines3/common/policies.py
```diff
@@ -237,9 +247,19 @@ def _predict(self, observation: th.Tensor, deterministic: bool = False) -> th.Te
         :return: Taken action according to the policy
         """
 
+    @staticmethod
+    def try_transpose_img_observation(obs: np.ndarray, observation_space: gym.spaces.Space):
```
araffin, line 251: missing docstring

stable_baselines3/common/preprocessing.py
```diff
@@ -61,6 +65,13 @@ def is_image_space(observation_space: spaces.Space, channels_last: bool = True,
     return False
 
 
+def has_image_space(observation_space: spaces.Dict):
```
araffin, line 68: missing docstring and return type

stable_baselines3/common/preprocessing.py
```diff
@@ -61,6 +65,13 @@ def is_image_space(observation_space: spaces.Space, channels_last: bool = True,
...
+    for key, subspace in observation_space.spaces.items():
+        if is_image_space(subspace):
+            return True
+    return False
+
+
 def preprocess_obs(obs: th.Tensor, observation_space: spaces.Space, normalize_images: bool = True) -> th.Tensor:
```
araffin, line 75: looks like the type hints are not valid anymore with Dict obs...

stable_baselines3/common/stacked_observations.py
```diff
@@ -0,0 +1,202 @@
...
+        self.n_stack = n_stack
+        (
+            self.channels_first,
+            self.stack_dimension,
+            self.stackedobs,
+            self.repeat_axis,
+        ) = self.compute_stacking(num_envs, n_stack, observation_space, channels_order)
+
+    @staticmethod
+    def compute_stacking(
```
araffin, line 42: missing docstring

stable_baselines3/common/stacked_observations.py
```diff
@@ -0,0 +1,202 @@
...
+            channels_first = channels_order == "first"
+
+        # This includes the vec-env dimension (first)
+        stack_dimension = 1 if channels_first else -1
+        repeat_axis = 0 if channels_first else -1
+        low = np.repeat(observation_space.low, n_stack, axis=repeat_axis)
+        stackedobs = np.zeros((num_envs,) + low.shape, low.dtype)
+        return channels_first, stack_dimension, stackedobs, repeat_axis
+
+    def stack_observation_space(self, observation_space):
```
araffin, line 71: missing type hints (dito for the other methods)

stable_baselines3/common/torch_layers.py
```diff
@@ -172,11 +177,10 @@ def __init__(
         # Iterate through the shared layers and build the shared parts of the network
         for idx, layer in enumerate(net_arch):
             if isinstance(layer, int):  # Check that this is a shared layer
-                layer_size = layer
                 # TODO: give layer a meaningful name
-                shared_net.append(nn.Linear(last_layer_dim_shared, layer_size))
+                shared_net.append(nn.Linear(last_layer_dim_shared, layer))  # add linear of size layer
```
araffin, line 181: why this change?
to avoid one assignment?

stable_baselines3/common/multi_input_envs.py
Miffyli: I agree. Normally a single file (e.g. `envs.py`) would suffice but these classes get lengthy with docstrings etc, so a separate folder `common/envs/` would be nice. We could also consider moving testing environments there so such envs can be used elsewhere as well (local tests with new algos, demonstrations, etc).

stable_baselines3/common/utils.py
```diff
@@ -317,3 +342,14 @@ def polyak_update(params: Iterable[th.nn.Parameter], target_params: Iterable[th.
         for param, target_param in zip_strict(params, target_params):
             target_param.data.mul_(1 - tau)
             th.add(target_param.data, param.data, alpha=tau, out=target_param.data)
+
+
+def obs_as_tensor(
```
araffin, line 347: missing docstring

stable_baselines3/common/off_policy_algorithm.py
Miffyli: I think this is issue with line lengths? Make sure you provide `-l 127` in arguments when running `black`.

araffin: which version of black is that? do you specify the `-l 127` argument?
and to reformat, you need to remove the last trailing comma before calling black, otherwise, black will keep it as-is.

We are using `black==20.8b1`.

in the worst case, I will see how to checkout part of the file the way they are in master.

JadenTravnik: I use black version `20.8b1` as well. I was forgetting to run black at all so I added a hook to VS code which runs it "on save". So since I wasn't running `black` explicitly, it was using the default 88. Before committing, I'd then run `black -l 127 .` but I am still potentially missing something as the mentioned comma issue resulted after doing this.

On that note though, is there consideration for using a `pyproject.toml` to have the black configuration explicit in the repo? (See [this comment](https://github.com/psf/black/issues/290#issuecomment-679841004))

stable_baselines3/common/torch_layers.py
JadenTravnik: Pretty much yea. I'm indifferent to it :p

stable_baselines3/common/off_policy_algorithm.py
araffin: >On that note though, is there consideration for using a pyproject.toml to have the black configuration explicit in the repo? (See this comment)

not really as we have already a `setup.cfg` file and that the `Makefile` is normally in charge of handling those default (although it works only on linux/mac os and probably on Windows Subsystem for Linux ).
but we may consider it in the future.

If possible, I would highly recommend to run `make commit-checks` before committing (it runs every checks except the tests), but it may be trickier on windows...

araffin: and what is your python version? (I'm using python 3.6 locally)

JadenTravnik: > If possible, I would highly recommend to run make commit-checks before committing
Will try

I'm on python 3.8

docs/misc/changelog.rst
```diff
@@ -19,13 +19,17 @@ New Features:
   automatic check for image spaces.
 - ``VecFrameStack`` now has a ``channels_order`` argument to tell if observations should be stacked
   on the first or last observation dimension (originally always stacked on last).
+- Add ``DictRolloutBuffer`` to support dictionary observations (need to be tested in ReplayBuffer)
```
araffin, line 22: please don't forget to add your username (and at the bottom too ;))
entries without usernames are from maintainers.

stable_baselines3/common/buffers.py
```diff
@@ -396,3 +429,293 @@ def _get_samples(self, batch_inds: np.ndarray, env: Optional[VecNormalize] = Non
...
+        super(DictReplayBuffer, self).__init__(buffer_size, observation_space, action_space, device, n_envs=n_envs)
+
+        assert isinstance(self.obs_shape, dict), "DictReplayBuffer must be used with Dict obs space only"
+        assert n_envs == 1, "Replay buffer only support single environment for now"
+
+        # Check that the replay buffer can fit into the memory
+        if psutil is not None:
+            mem_available = psutil.virtual_memory().available
+
+        if optimize_memory_usage:
```
araffin, line 470: I would rather put an assert and remove all the code related to it

stable_baselines3/common/preprocessing.py
J-Travnik: Should this wait for when nested dicts are supported?

J-Travnik:
@araffin Is anything left to do on this submission?

araffin:
>@araffin Is anything left to do on this submission?

Well, more tests (only PPO is tested for now), fix all the warnings from the linter, and update the documentation.

I will try to take a deeper look this week (or maybe today).

araffin:
Sorry for the delayed answer, I'm quite busy right now...

Quick remarks:
- why not put all custom envs in an `envs` folder inside the `common/` folder?
- why the stacked observation code in not in `common/vecenvs/` ?
- we still need to update the documentation and at least add an example
- thanks for updating the tests =) (I also noticed that the current implementation will crash if using a `GoalEnv` with PPO because of the additional checks we do for HER, I will see how much effort it is to change HER in another PR)

J-Travnik:
@araffin 

> why not put all custom envs in an envs folder inside the common/ folder?

Done, moved over the BitFlipEnv as well. Saw that there were dummy ones in the tests. Should they be moved over as well?

> why the stacked observation code in not in common/vecenvs/ ?
> We still need to update the documentation and at least add an example

I moved the StackedObservation code over and started updating the documentation for the StackedObs and an example but I wanted to reference the Envs (BitFlipEnv etc) but I am not familiar enough with sphinx to know how to make a new Env page or reference it from the StackedObs documentation.

araffin:
>Saw that there were dummy ones in the tests. Should they be moved over as well?

I'm not sure... they are not real env whereas BitFlippingEnv is an actual one for instance.

>I moved the StackedObservation code over and started updating the documentation for the StackedObs and an example but I wanted to reference the Envs (BitFlipEnv etc) but I am not familiar enough with sphinx to know how to make a new Env page or reference it from the StackedObs documentation.

perfect, I fixed the different issues ;)


I hope to have more time end of this week or next week to continue the review ;)
Anyway, thank you already for your good work =)

J-Travnik:
@Miffyli @araffin @PartiallyTyped Hi! Hope y'all enjoyed the holidays.

I think this PR is worth some attention as it creates a kind of bottleneck for other features that either rely on it like
- recurrent policies
- HER
- Invalid Action Masking

or may conflict with it:
- n-step buffers from #81. I am not certain there will be a conflict but I assume so because both PRs introduce child class buffers. 

Although I agree that a 1.0 release is a priority, my concern is mainly that some features are or will become dependent on this PR before they can be developed further. Perhaps I am mistaken but that is my reasoning to ask:
What is the best course of action regarding merging this into master?

stable_baselines3/common/torch_layers.py
```diff
@@ -224,6 +228,76 @@ def forward(self, features: th.Tensor) -> Tuple[th.Tensor, th.Tensor]:
...
+        super(CombinedExtractor, self).__init__(observation_space, features_dim=features_dim)
+
+        extractors = {}
+
+        total_concat_size = 0
+        for (key, subspace) in observation_space.spaces.items():
+
+            if is_image_space(subspace):
+                n_input_channels = subspace.shape[0]
+                cnn = nn.Sequential(
```
Miffyli, line 265: This could be replaced with NatureCNN, no? This network seems small and I'd opt for something time-tested, and original DQN network has served people well so far.

Miffyli:
Hey @J-Travnik! Indeed we had very snowy holidays here in north, hope you had similar experiences :)

I skimmed over the code and left one comment regarding default networks. I also started going through docstrings and all that, but stopped before I got too wild with it. Do you mind if I push some changes directly to branch to help getting this PR done rather than just commenting? 

@araffin I would like you to review this PR but I will be going over it in detail as well (I really would love to have this so I do not have to mess around with branches :') ).

stable_baselines3/common/torch_layers.py
Miffyli: (After update) As in could this not use the NatureCNN extractor instead of creating the network separately here? I see it could save bit of duplicate code :). This could even be an argument to the extractor so one can change CNN extractor easier.

J-Travnik:
@Miffyli Glad to hear. Same here :) 

I changed the CNN architecture to be that of the NatureCNN (I don't use the CNN Extractor though as extractors with child extractors seems like a hard thing to manage).

Also feel free to push commits directly to the PR. 👍

araffin:
Hello and happy new year!

I'm slowly back from holidays (there was a big stack of notifications waiting for me, I'm still unstacking it...).

>I think this PR is worth some attention as it creates a kind of bottleneck for other features that either rely on it like

Overall, do not worry about other PR, it's better to take the time and do the thing the right way rather than hurry up and change the implementation later.

> recurrent policies

I'm not sure why RNN would rely on that PR.

>HER

yes, I'll probably start that in parallel as we may need it sooner than I expected.

>Invalid Action Masking

this is an experimental feature that belongs to sb3-contrib (https://github.com/Stable-Baselines-Team/stable-baselines3-contrib/issues/16) and that does not necessary need dict obs.

>or may conflict with it:
>n-step buffers from #81. I am not certain there will be a conflict but I assume so because both PRs introduce child class buffers.

Do not worry about that one, which is quite experimental.

>What is the best course of action regarding merging this into master?

I need to go through the code again... I'm doing that today and then we may still have some exchanges until we converged ;)

docs/guide/algos.rst
```diff
@@ -19,7 +19,8 @@ TD3          ✔️          ❌            ❌                ❌
 
 
 .. note::
-    Non-array spaces such as ``Dict`` or ``Tuple`` are not currently supported by any algorithm.
+   ``Tuple`` spaces are not supported by any environment however ``Dict`` spaces of ``Box`` spaces are. 
+   ``Dict`` spaces of containing other kinds of spaces (e.g., ``Discrete``) have not yet been explored.
```
araffin, line 23: why not?
we should have tests for that (using dummy dict env)

docs/common/envs.rst
```diff
@@ -0,0 +1,31 @@
...
+
+
+SimpleMultiObsEnv
+-----------------
+
+.. autoclass:: SimpleMultiObsEnv
+  :members:
+
+
+NineRoomMultiObsEnv
```
araffin, line 27: do we really need the two?

stable_baselines3/common/envs/multi_input_envs.py
```diff
@@ -0,0 +1,266 @@
...
+                self.state_mapping.append({"vec": col_vecs[i], "img": row_imgs[j]})
+
+    def get_state_mapping(self) -> Dict[str, np.ndarray]:
+        """
+        Uses the state to get the observation mapping and applies noise if there is any.
+
+        :return: observation dict {'vec': ..., 'img': ...}
+        """
+        state_dict = self.state_mapping[self.state]
+        if self.noise > 0:
```
araffin, line 123: is that really needed?

stable_baselines3/common/envs/multi_input_envs.py
```diff
@@ -0,0 +1,266 @@
...
+        """
+        Generated a random image and upsample it.
+
+        :param value_range: The range of values for the img
+        :param initial_size: The initial size of the image to generate
+        :param up_size: The size of the upsample
+        :return: upsampled img
+        """
+        im = np.random.randint(value_range[0], value_range[1], initial_size, dtype=np.int32)
+        return np.array(
```
araffin, line 88: is that double for loop really needed?

stable_baselines3/common/envs/multi_input_envs.py
```diff
@@ -0,0 +1,266 @@
...
+       | 8|_9_10_|11|
+       |12 13  14 15|
+       ¯¯¯¯¯¯¯¯¯¯¯¯¯¯
+
+    start is 0
+    states 5, 6, 9, and 10 are blocked
+    goal is 15
+    actions are = [left, down, right, up]
+
+    simple linear state env of 15 states but encoded with a vector and an image observation
```
araffin, line 27: so the image is just a copy of the vector?
wouldn't be better to have complementary information in order to check that the implementation is actually working?

stable_baselines3/common/policies.py
```diff
@@ -260,30 +287,35 @@ def predict(
         #     state = self.initial_state
         # if mask is None:
         #     mask = [False for _ in range(self.n_envs)]
-        if isinstance(observation, dict):
+        # Need to check the observation if its a ObsDictWrapper
+
+        # Special Case for GoalEnv (using HER normally)
+        if isinstance(observation, dict) and set(observation.keys()) == set(["observation", "desired_goal", "achieved_goal"]):
```
araffin, line 293: note for myself: this breaks the code for A2C/PPO

stable_baselines3/common/preprocessing.py
```diff
@@ -95,11 +114,18 @@ def preprocess_obs(obs: th.Tensor, observation_space: spaces.Space, normalize_im
     elif isinstance(observation_space, spaces.MultiBinary):
         return obs.float()
 
+    elif isinstance(observation_space, spaces.Dict):
+        for key, _obs in obs.items():
```
araffin, line 118: why not a recursive call to support more types?

stable_baselines3/common/preprocessing.py
araffin: this is to support not only Box spaces, not for nested Dict (I don't think we should support nested dict)

stable_baselines3/common/torch_layers.py
```diff
@@ -224,6 +228,88 @@ def forward(self, features: th.Tensor) -> Tuple[th.Tensor, th.Tensor]:
...
+    :param combined_net_arch: Architecture of the combined network module which calculates the final feature extracted
+    """
+
+    def __init__(
+        self,
+        observation_space: gym.spaces.Dict,
+        features_dim: int = 64,
+        cnn_output_dim: int = 64,
+        mlp_output_dim: int = 64,
+        mlp_net_arch: List[int] = [64, 64],
```
araffin, line 254: please avoid mutables as default, see https://github.com/satwikkansal/wtfpython#-beware-of-default-mutable-arguments

araffin (commented):
I've started updating the doc (there was missing things here and there), I was wondering, did you check the performance of the current implementation?
i.e., can it solve something that is not a toy example and that involve the combination of vector obs and image?

stable_baselines3/common/torch_layers.py
```diff
@@ -224,6 +228,89 @@ def forward(self, features: th.Tensor) -> Tuple[th.Tensor, th.Tensor]:
...
+                        activation_fn,
+                        squash_output=False,
+                    )
+                )
+
+                total_concat_size += mlp_output_dim
+
+        self.extractors = nn.ModuleDict(extractors)
+
+        self.combined = nn.Sequential(
```
araffin, line 305: why do you have that?
it looks like a duplicate with the network that will be used after... (and cnn features are already smaller)

docs/guide/algos.rst
J-Travnik: I haven't had a moment to try other kinds of spaces yet but will try to this week using the dummy dict env. It will be a good test to see if frame stacking will work out of the box (pun intended).

docs/common/envs.rst
J-Travnik: Probably not. I'll take it out on the next push.

stable_baselines3/common/envs/multi_input_envs.py
J-Travnik: No, I was originally going to add something that upsamples a simple grid representation of the grid world but I think there are better options for testing the dict observations.

stable_baselines3/common/envs/multi_input_envs.py
J-Travnik: The image is not a copy of the vector as they are both independently randomly generated.
For every row there is an image, and for every column there is a vector. Only by combining them can an agent infer what state they are in.

stable_baselines3/common/torch_layers.py
J-Travnik: I wasn't sure what the best approach was. I can remove the combined mlp altogether and return the `th.cat` of the encoded tensors. What would you recommend?

J-Travnik:
@araffin 
> I've started updating the doc (there was missing things here and there), I was wondering, did you check the performance of the current implementation?
> i.e., can it solve something that is not a toy example and that involve the combination of vector obs and image?

I am still in the process of verifying this on my end but it is in a private environment. As an alternative, I was considering [gym_miniworld](https://github.com/maximecb/gym-miniworld) but I am open to other environments to validate this.

stable_baselines3/common/envs/multi_input_envs.py
J-Travnik: Removed in latest commit.

stable_baselines3/common/envs/multi_input_envs.py
J-Travnik: Removed in latest commit

J-Travnik:
So I have added a `check_channels` parameter to the `CombinedExtractor`. I have had a lot of issues trying to uses stacked observations with images. I think that the definition of what an image is in `is_image_space()` in preprocessing.py, is a bit too constrained. For instance, if we are using a depth based image, this might not be in `uint8` format. Thoughts?

araffin:
>As an alternative, I was considering gym_miniworld but I am open to other environments to validate this.

For debugging, your private env is fine, but yes for final performance check, I would go for `gym_miniworld` (should be easy to integrate to the rl zoo)

>I have had a lot of issues trying to uses stacked observations with images. 

I will have a look, but we can probably leave that for now (the stacked vector was a requirement, but the version for images a plus)

>Also, I have noticed that not all of the tests seem to run, it looks like everything runs in alphabetical order until tests\test_logger.py then it stops. Not sure what is happening there.

that's weird, at least all tests run on github CI: https://github.com/DLR-RM/stable-baselines3/pull/243/checks?check_run_id=1705928341

I will probably take a look on what you did today, otherwise, it will be on monday ;)

stable_baselines3/common/torch_layers.py
araffin: yes please, the current structure of SB3 networks are composed of two parts:
the feature extractor (a flatten layer (for vector obs), a CNN or here a combination) and then a mlp used to predict the action, the value from the extracted features.

araffin: writing that, I realized that we now have an inconsistency with how SB3 works with vector obs (here we have an additional MLP instead of just a flatten layer)

Miffyli:
>  I think that the definition of what an image is in is_image_space() in preprocessing.py, is a bit too constrained. For instance, if we are using a depth based image, this might not be in uint8 format. Thoughts?

Dropping in to comment on this. While I agree it is restrictive, it is also a good time-proven safeguard against dumb data. It is easier to enforce specific data-type and normalize inside network than add checks/support for all data-types, and without this check people often end up not normalizing data (a very common mistake). This also makes storing data easier/faster. 

If more support is added, this should be an another issue/PR for better discussion. I think we should leave it as it is for now.

docs/guide/algos.rst
araffin: spoiler alert: it does not ^^"
but I did some bug fixes and now it should ;)

araffin:
@J-Travnik I did some (nasty) bug fixes ;) (tensor was modified by reference) and added some tests for mixed dict env (mixing vector, discrete obs and image).
support for channel last env is yet to be done
and we need to update the env checker + add a check to ensure the env has only one level (otherwise throw an error and advise the user to flatten it).

J-Travnik:
> @J-Travnik I did some (nasty) bug fixes ;) (tensor was modified by reference) and added some tests for mixed dict env (mixing vector, discrete obs and image).
> support for channel last env is yet to be done
> and we need to update the env checker + add a check to ensure the env has only one level (otherwise throw an error and advise the user to flatten it).

Good catches! Thanks for getting to the mixed dict env. I did try to make time last week but alas, life happened.
That sounds like some reasonable checks. I'll try to find time on my side to do so.

stable_baselines3/common/envs/multi_input_envs.py
araffin: took me some time to understand it ^^" now it makes sense and I've added comments too

Miffyli:
Sorry for delays again. I tried applying this on ViZDoom health-gathering task (not solvable unless you provide "health" information. Gym-miniworld has similar env but it does not provide enough challenge out-of-the-box nor the health variable (at least directly). I only could do one repetition as I have little compute to spare but results in this env are very determistic, so I call this a win:

results

Using image alone does not work, but feeding required health information with dict-obs learns faster as expected! At the very least now Image + 1D vector input works as expected. I include code here in case somebody wants to use it, but I can run more formal experiments when needed. [vizdoom_hgs_test.zip](https://github.com/DLR-RM/stable-baselines3/files/5863447/vizdoom_hgs_test.zip)


**Bug:** Code initially failed on failed convolutional layers ("filter size exceeds..."). Seems like image transposing was not applied despite env using channels-last notation. Changing to channels-first inside environment works.

J-Travnik:
@Miffyli 
Indeed, I can confirm both channel-last issue and the learning using a private env over multiple runs. On my end, I specifically add `check_channels` to the `features_extractor_kwargs`. Is there interest in adding a `channels_last` parameter as well?

Thanks so much for the updates @araffin. I've been swamped with other commitments so it is very appreciated :)

Miffyli:
> Indeed, I can confirm both channel-last issue and the learning using a private env over multiple runs. On my end, I specifically add check_channels to the features_extractor_kwargs. Is there interest in adding a channels_last parameter as well?

Ah good point! I would actually make this `True` by default and add a highlighted note about it in the docs. This way the default behaviour is same as with rest of the code ("CnnPolicy"), but with highlight in docs someone wanting to use custom images can find easily.

araffin:
> On my end, I specifically add check_channels to the features_extractor_kwargs. Is there interest in adding a channels_last parameter as well?

This check is deactivated by default for CnnPolicies normally.

>Ah good point! I would actually make this True by default and add a highlighted note about it in the docs.

I think when I tried, this break things in an unexpected way.

But why can't we add automatic support for channel-last env? (this would require to tweak `VecTranspose` a little, but that's all, no?)

>Only use MLP extractor if observation is Box with ndim=1

As I mentioned the default is just a flatten layer, to be consistent with the `MlpPolicy`.

>Only use CNN extractor if observation is image in shape that can be fed to CNN

this is already the case (but the transpose part, with `VecTranspose` is missing)

>Add optional extractor_modules: Dict[str, BasicFeatureExtractor] argument to CombinedExtractor, which can be used to provide custom extractors. If key matches observation space's key, use that extractor.

well, in that case, better for the user to define a custom feature extractor, no? (we can add an example in the doc)

>Thanks so much for the updates @araffin. I've been swamped with other commitments so it is very appreciated :)

de rien ;)

ErikVester:
Hey all, 

thank you for this great project and the continuous work on it! I am really looking forward to this merge as I am working with Dict observations. I had a short question that's very related to this, so thought I'd just ask it here. Hope that's ok, if not, let me know and I'll remove it! 

If I have an observation that contains two boxes; one  is a 3D array encoding of a grid world and the other is an array with 2 integers which represent some information about the agent. Would it be possible to just flatten these two arrays and concatenate them and then use the Mlppolicy? Is this in some way worse than using the actual dict observations? 

This would really help me out!  

Kind regards, 

Erik

araffin:
>Hope that's ok, if not, let me know and I'll remove it!

for one question, it is ok but later, issues are made for that ;)

>Would it be possible to just flatten these two arrays and concatenate them and then use the Mlppolicy? Is this in some way worse than using the actual dict observations?

yes, you can flatten the observation (there is a wrapper in gym for that) but you will lose the spatial information that could be used by the CNN, and that the mlp will have to learn (which also means more parameters here).

In any case, you can also already use that branch if you want (see install from source guide in the documentation).

ErikVester:
Hi, 

I still had one questions related to my previous question, which might be of interest for this implementation. 

Considering the case where we have a dict observation of an image and some other information in the form of an array of integers. If this array is relatively small (in my case it's only 2 integers), does this lead to any problems due to the different output sizes that are concatenated for the combined MLP? In my case this will lead to concatenating an output of size 64 with an output of size 2. 

Would a solution be to first 'explode' the small array to a bigger one and adding this to the observation instead of the small array? 

Kind regards,

Erik

ErikVester:
Not sure if I am right, but it seems to me as if currently the CombinedExtractor doesn't make an MLP for the vector shaped input, but just flattens it:

image

The given mlp_net_arch  is also never used I think.

araffin:
>Not sure if I am right, but it seems to me as if currently the CombinedExtractor doesn't make an MLP for the vector shaped input, but just flattens it:

yes, that's intended (see comments about it above) so we match the behavior of normal mlp policy.
Also, we will remove the mlp of the feature extractor.

>The given mlp_net_arch is also never used I think.

yes, we need to clean that.

Miffyli:
I added proper support for vectranspose, completed some TODOs and simplified CombinedExtractor as discussed above (only applies NatureCNN on images and flattens rest, does not add any other layers). I checked  the performance with above ViZDoom test and learning curves remain same, so nothing broke so far :)

JadenTravnik:
@Miffyli @araffin Yo, I've been swamped with other things but I wanted to reach out and say thank you so much for continuing this PR and in general updating the repository to finally get to v 1.0. Just reading through the blog its easy to see that its an incredible milestone and the hard work and dedication that you and the other collaborators have put in will benefit the entire community. Thank you so much. :)

Miffyli:
@JadenTravnik Thank you for the kind words! It is equally exciting to see the library being used even in video game companies (at least I am super excited about it :D). No worries about not being too active on this PR, you did a big service already! We will finish this up with arrafin as one of the next steps.

araffin:
@Miffyli when finishing the PR for HER (#351 ), I noticed, that the `cnn_output_dim` was set to 64 instead of the 512 that goes with the default `NatureCnn`, is there a reason for that? (reduce computation/memory?)

Miffyli:
@araffin No particular reason (I think this stems from the original dict obs PR), but I'd say it is for reducing computation/memory. Having more than one image inputs mapped to length 512 vectors would get pretty expensive. Maybe set the default to 128 or 256 (64 might be bit too small)?

araffin:
>but I'd say it is for reducing computation/memory. 

That's what I thought too.

>Maybe set the default to 128 or 256 (64 might be bit too small)?

yes, I would go for 128 or 256 too.

araffin:
@J-Travnik @Miffyli The refactoring of HER is done, tested and merged in here (I also added proper timeout handling).
We mostly need to do a final review of this PR and it is good to be merged =)

stable_baselines3/common/vec_env/vec_transpose.py
```diff
@@ -14,9 +17,20 @@ class VecTransposeImage(VecEnvWrapper):
...
-        assert is_image_space(venv.observation_space), "The observation space must be an image"
+        assert is_image_space(venv.observation_space) or isinstance(
+            venv.observation_space, spaces.dict.Dict
+        ), "The observation space must be an image or dictionary observation space"
 
-        observation_space = self.transpose_space(venv.observation_space)
+        if isinstance(venv.observation_space, spaces.dict.Dict):
+            self.image_space_keys = []
+            observation_space = deepcopy(venv.observation_space)
+            for key, space in observation_space.spaces.items():
```
araffin, line 27: here we are assuming that all image spaces follow the same convention (which is reasonable), shall we do a check? (it will crash later on anyway)

JadenTravnik: I think its worth doing the check here in the init. We have everything to make an informative error message so we might as well.

JadenTravnik (commented):
@araffin The integration of HER looks awesome. :)
I did a pass and didn't find much to comment on so koodos 👍

stable_baselines3/common/torch_layers.py
```diff
@@ -93,7 +94,11 @@ def forward(self, observations: th.Tensor) -> th.Tensor:
 
 
 def create_mlp(
-    input_dim: int, output_dim: int, net_arch: List[int], activation_fn: Type[nn.Module] = nn.ReLU, squash_output: bool = False
+    input_dim: int,
```
JadenTravnik, line 97: I can't comment on none modified code but above this line on line 69 is a comment about probably using an MlpPolicy when the `asset is_image_space(observation_space)` fails. Perhaps we should change the error to include MultiInputPolicy?

Miffyli (approved):
Looks positively brilliant to me! 👍 . If no further changes are required (see above comment), feel free to merge @araffin 

@J-Travnik 🍻 / ☕ / 🍵 / hot-chocos on me if we ever get to meet in person! Thanks for all your time and patience! ^^

araffin:
Finally merged :tada: 
Thanks for the PR =D!

J-Travnik:
@Miffyli @araffin This is wonderful :)
Looking forward to hot cocos in the future ;)

PR #245, opened 2020-11-28
TD3 Code review

### LINKED ISSUE #17: Review of Existing Code

### LINKED ISSUE #331: [Bug] action_noise never got used in HER
### 🐛 Bug
Hi when I was trying to use HER with DDPG, I noticed that the action noise I feed into HER never got used by DDPG. 
Please correct me if I am wrong, if you take a look at the code on 197 at HER.py
```
            rollout = self.collect_rollouts(
                self.env,
                n_episodes=self.n_episodes_rollout,
                n_steps=self.train_freq,
                action_noise=self.action_noise,
                callback=callback,
                learning_starts=self.learning_starts,
                log_interval=log_interval,
            )

```
I have printed out self.action_noise, it is NONE. the action_noise got fed into self.model during the init part (line 103 as kwargs), but I think it never got assigned to it self as an attribute, maybe we shall do self.model.action_noise there? Let me know if I am mistaken.

ernestum:
## Description
Discussion MR for the TD3 implementation

## Motivation and Context

This code review is part of the review of existing code #17.

Also fixes issues with terminal observation.

closes #17
closes #331 

## Types of changes

## Checklist:


stable_baselines3/common/policies.py
```diff
@@ -119,12 +119,11 @@ def extract_features(self, obs: th.Tensor) -> th.Tensor:
         preprocessed_obs = preprocess_obs(obs, self.observation_space, normalize_images=self.normalize_images)
         return self.features_extractor(preprocessed_obs)
 
-    def _get_data(self) -> Dict[str, Any]:
+    def _get_data_to_reconstruct_model(self) -> Dict[str, Any]:
```
araffin, line 122: suggestion: `get_constructor_parameters()` but I'm ok with this name too ;)

araffin:
@ernestum are you done with the edits?

ernestum:
No I still wanted to implement the more convenient way to specify the training frequency based on either steps or episodes. I am on it right now.

I noticed that SAC does not support updating after n episodes (just after n steps). Was it supposed to be that way? Because after the patch that will be possible.

ernestum:
Another question regarding [this line](https://github.com/DLR-RM/stable-baselines3/blob/52003e0bd8ff7b284b96c240bb98e06dd2cc6100/stable_baselines3/td3/td3.py#L163) (and all other lines where a `polyak_update` is done):
```python
if gradient_step % self.policy_delay == 0:
    ...
    polyak_update(...)
```
Note that this triggers a polyak update in each first iteration of the outer `for gradient_step in range(gradient_steps)` loop since `0 % x == 0` for all `x`.
 In the case where we set `train_freq=1` the policy is not delayed since it will be polyak-updated after each step. I guess this should be more like `if self.total_train_steps % self.policy_delay == 0:` right?

Miffyli:
> I noticed that SAC does not support updating after n episodes (just after n steps). Was it supposed to be that way? Because after the patch that will be possible.

Jumping in here to comment on this. Is this something people do on robotics side? I have personally not seen "update every N episodes" type of setups, but I mainly focus on games and non-robotics envs. It sounds very sensitive to problems (e.g. agent gets better -> episodes last longer -> less updates), and support for this could be confusing. If support for this is added it should be blatantly clear how "update every N steps" and "update every N episodes" function together (e.g. do they override each other? In which order? Do they both apply?)

araffin:
>I noticed that SAC does not support updating after n episodes (just after n steps). Was it supposed to be that way? Because after the patch that will be possible.

why not? (I've been using that feature on a real robot...)

> In the case where we set train_freq=1 the policy is not delayed since it will be polyak-updated after each step. I guess this should be more like if self.total_train_steps % self.policy_delay == 0: right?

oh... nice catch ;)

> Is this something people do on robotics side? 

yes, mostly because everything happens in real time, you cannot afford having inconsistent and slow control frequency because your policy is updating (whereas in simulation, you can stop the simulation until you are ready again).

>If support for this is added it should be blatantly clear how "update every N steps" and "update every N episodes" function together (e.g. do they override each other? In which order? Do they both apply?)

they are already implemented, and there is a big warning when one is trying to use both at the same time.
@ernestum proposal was in fact to make things clearer.

stable_baselines3/common/off_policy_algorithm.py
```diff
@@ -83,9 +81,8 @@ def __init__(
         batch_size: int = 256,
         tau: float = 0.005,
         gamma: float = 0.99,
-        train_freq: int = 1,
+        train_freq: Union[int, Tuple[int, str]] = (1, "episode"),
```
araffin, line 84: please keep the default ("step")

stable_baselines3/common/off_policy_algorithm.py
```diff
@@ -250,11 +239,17 @@ def learn(
         callback.on_training_start(locals(), globals())
 
         while self.num_timesteps < total_timesteps:
+            if isinstance(self.train_freq, int):
```
araffin, line 242: missing comment

araffin: but I would do that transform directly in the constructor, no?

stable_baselines3/common/off_policy_algorithm.py
ernestum: I was assuming the default should be 1 step? So either `1` or `(1, "step")` should be fine? Why are you suggesting 1 episode? Or am I getting something wrong here?

araffin: yes, the default should be one step (but that's not what it is right now, no?) I read `(1, "episode"),`

stable_baselines3/common/off_policy_algorithm.py
ernestum: You mean I should add a comment describing the translation from the old style to the new style?

I deliberately kept the transition here since otherwise we would 1. need to store `n_steps` and `n_episodes` in the `OffPolicyAlgorithm` and 2. the user could not change the `train_freq` in hindsight or during training. Its unlikely that it is needed but WHEN you need it it is really annoying.

stable_baselines3/common/off_policy_algorithm.py
```diff
@@ -128,23 +125,15 @@ def __init__(
...
-            warnings.warn(
-                "You passed a positive value for `train_freq` and `n_episodes_rollout`."
-                "Please make sure this is intended. "
-                "The agent will collect data by stepping in the environment "
-                "until both conditions are true: "
-                "`number of steps in the env` >= `train_freq` and "
-                "`number of episodes` > `n_episodes_rollout`"
-            )
+        if isinstance(train_freq, tuple) and train_freq[1] not in ("step", "episode"):
+            warnings.warn(f"The units of the train_freq must be either `step` or `episode` not `{train_freq[1]}`!")
```
araffin, line 136: why not an assert?

stable_baselines3/common/off_policy_algorithm.py
araffin: >You mean I should add a comment describing the translation from the old style to the new style?

yes

>1. need to store n_steps and n_episodes in the OffPolicyAlgorithm 

this is already the case (`self.n_episodes_rollout` and `self.train_freq`)

> 2. the user could not change the train_freq in hindsight or during training.

it can if we store it.
Also, if needed, the user can direct call `collect_rollout` alone if needed.

Last thing, I think the check can be simplified if we have an assert about `self.train_freq[1]` before (see comment above ;))

ernestum: nope `self.n_episodes_rollout` is gone now. only `self.train_freq` is left and it is split up into old-style `n_steps` and `n_episodes` before passing it to `collect_rollouts` since I was too lazy to also change that implementation too (but we should do that in the future).

I did not quite get what you mean with the user calling `collect_rollout` directly.

Good point about normalizing `self.train_freq` to be a tuple in the constructor. I will do that.

stable_baselines3/common/off_policy_algorithm.py
ernestum: Good point I will change that.

stable_baselines3/common/off_policy_algorithm.py
araffin: what I meant was keeping `self.n_episodes_rollout` (that would be nice to load old models too) but computing the value of `self.train_freq` and `self.n_episodes_rollout` in the constructor.

>I did not quite get what you mean with the user calling collect_rollout directly.

the same way the user can call `model.train()` (just updating the model), the user can also directly call `model.collect_rollout()` which the parameters the user wants.

stable_baselines3/common/off_policy_algorithm.py
ernestum: Fixed it

stable_baselines3/common/off_policy_algorithm.py
```diff
@@ -128,23 +125,17 @@ def __init__(
...
-            warnings.warn(
-                "You passed a positive value for `train_freq` and `n_episodes_rollout`."
-                "Please make sure this is intended. "
-                "The agent will collect data by stepping in the environment "
-                "until both conditions are true: "
-                "`number of steps in the env` >= `train_freq` and "
-                "`number of episodes` > `n_episodes_rollout`"
-            )
+        if isinstance(self.train_freq, int):
+            self.train_freq = (self.train_freq, "step")
```
araffin, line 136: what do you think of using NamedTuple? so later you can do `train_freq.n` and `train_freq.type` instead of `train_freq[1]` ?

ernestum: Sure but can I have `train_freq.quantity` and `train_freq.unit`?

araffin: `quantity` sounds a bit weird to me... I think you wrote `frequency` in the docstring, no?

stable_baselines3/common/off_policy_algorithm.py
ernestum: I would be reluctant to add the public `self.n_episodes_rollout` back. This way we basically build an API with two ways to specify the number of epochs between two training steps and introduce back the issue that the value of `self.n_episodes_rollout` can conflict with `self.train_freq`. If you like I can come up with a separate solution to ensure backwards compatibility with old stored models.

Since, as you pointed out, `model.collect_rollout()` is part of the public API and it still has the old-style `n_episodes` and `n_steps` interface that one should actually be updated to the new tuple format as well and I should stop being lazy!

araffin: >I would be reluctant to add the public self.n_episodes_rollout back

Fair enough ;)

>nterface that one should actually be updated to the new tuple format as well and I should stop being lazy!

if you do so, you may also need to remove the `replay_buffer` parameter ;) (which was here for historic reasons)

araffin:
The current failure on the CI is due to a newer version of numpy, we should replace `np.bool` by `bool` and that should solve it ;)

stable_baselines3/common/off_policy_algorithm.py
ernestum: You are right. I was borrowing terminology from [here](https://en.wikipedia.org/wiki/Unit_of_measurement) and messed it up. In our case the `train_freq` is the quantity, your `n` is the magnitude unit of measurement must be either `step` or `episode` (but this could be extended in the future). So I propose `train_freq.magnitude` and `train_freq.unit`.

ernestum:
> The current failure on the CI is due to a newer version of numpy, we should replace `np.bool` by `bool` and that should solve it ;)

Thanks for the hint I will look into fixing it tomorrow.

stable_baselines3/common/off_policy_algorithm.py
ernestum: noted ...

stable_baselines3/common/off_policy_algorithm.py
araffin: I would really for go `train_freq.frequency` or `train_freq.interval` (so discrete values, that's why I proposed the `n` at first, from the maths side) as `magnitude` sounds like a continuous value to me.
and yes for the `train_freq.unit` ;)

araffin:
@AdamGleave my last commit (https://github.com/DLR-RM/stable-baselines3/pull/245/commits/d83cdc9687b57b70ab1559fb398b3833e7bc0d2b) might affect your imitation learning lib ;) (it's about having the right terminals)

araffin (approved):
LGTM, thanks for the fixes and the discussion =)

AdamGleave:
> @AdamGleave my last commit ([d83cdc9](https://github.com/DLR-RM/stable-baselines3/commit/d83cdc9687b57b70ab1559fb398b3833e7bc0d2b)) might affect your imitation learning lib ;) (it's about having the right terminals)

Thanks for flagging! Tagging @shwang FYI: we have previously been seeing unnormalized terminal observations, this PR will fix that.

PR #270, opened 2020-12-23
Improve error messages when PPO effective batch size is 1 and when last mini-batch is truncated

### LINKED ISSUE #265: [Bug] Silent NaNs in PPO Loss Calculation if n_steps=1 and n_envs=1
### 🐛 Bug

This is somewhere between a bug and a request for more informative errors: 

When n_steps and n_envs are both set to 1, the batch returned by a rollout buffer [here](https://github.com/DLR-RM/stable-baselines3/blob/master/stable_baselines3/ppo/ppo.py#L160) will be of length 1. This will make the `advantages` calculation return `nan` values, since the normalization step involves calculating the standard deviation of advantages, which is undefined for a single element.

I recognize that this small of a setting is definitely an edge case (I ran into it during testing, when we were setting all values quite low for speed reasons), so I'm not sure it makes sense to have logic for this case, but at minimum, I think it would be beneficial to have some kind of explicit warning that checks if `actions` or `advantages` have a single element, so that there's a clear indication of the source of the issue, rather than having to follow a breadcrumb trail of `nan`s from some higher abstraction level of code

### To Reproduce

```import gym

env = gym.make('CartPole-v1')
model = PPO('MlpPolicy', env, verbose=1, n_steps=1)
model.learn(total_timesteps=10)
```
This will fail with an unclear error : 
```
RuntimeError: invalid multinomial distribution (encountering probability entry < 0)
```
If you put a debugger or insert logging statements at ppo.py:170, you'll be able to see that (1) `len(advantages) = 1` and consequently (2) `advantages.std() = nan`, which first arises as a visible bug when you try to collect an on-policy rollout after your first training step, since the `nan` values in loss propagate into `nan` parameter values. 

### Expected behavior

Either (1) explicit support for training on effective batches of size 1, or (2) a clearer and earlier error when you attempt to construct an algorithm object with `n_steps=1` and `n_envs=1`, informing the user that the case isn't supported.

### System Info

Describe the characteristic of your environment:
 * Describe how the library was installed (pip, docker, source, ...): Cloned from fork of current master, installed via `pip` 
 * GPU models and configuration: N/A 
 * Python version: 3.7.0 
 * PyTorch version: 1.7.1 
 * Gym version: 0.17.3
 * Versions of any other relevant libraries: N/A

### Checklist

decodyng:
## Description
At a high level, this PR adds clearer error and warning handling around situations where the relationship between the max size of your env steps buffer (n_steps*n_envs) might cause problems or generated unexpected behavior. 

Specifically, it: 
1. Raises an error when the size of your env steps buffer (and thus the maximum possible size of your training batch) is 1, since PPO training code calculates a standard deviation, which is undefined for a single-length batch 
2. Raises a warning when the size of your env steps buffer does not divide evenly into your passed-in batch size, which result in some fraction of batches (potentially up to 100% of batches) being truncated. 

## Motivation and Context
Closes #265 
- [x ] I have raised an issue to propose this change ([required](https://github.com/DLR-RM/stable-baselines3/blob/master/CONTRIBUTING.md) for new features and bug fixes)

## Types of changes

I'm not really sure which of these boxes apply for adding both a warning and a raised error. I'm fairly confident that in the scenario where I raise an error, code will break subsequently, so in that sense it's not a breaking change, but my change does prompt the error to happen _earlier_ than it otherwise would have, so some code that e.g. only constructed a model and never tried to actually train it might error where it hadn't before. 

## Checklist:


decodyng:
I'm not sure if this change is in fact actually enough of a new feature that you would want an explicit test confirming that a Warning gets raised in a scenario where you would expect one; if so, I would be happy to create such a test.

araffin:
>an explicit test confirming that a Warning gets raised in a scenario where you would expect one; if so, I would be happy to create such a test.

yes, please do ;)

stable_baselines3/ppo/ppo.py
```diff
@@ -1,3 +1,5 @@
+import warnings
+from math import floor
```
araffin, line 2: you don't need that normally, for instance `3 // 2` is the euclidian division of 3 by 2 ;)

stable_baselines3/ppo/ppo.py
```diff
@@ -28,7 +30,8 @@ class PPO(OnPolicyAlgorithm):
     :param learning_rate: The learning rate, it can be a function
         of the current progress remaining (from 1 to 0)
     :param n_steps: The number of steps to run for each environment per update
-        (i.e. batch size is n_steps * n_env where n_env is number of environment copies running in parallel)
+        (i.e. replay buffer size is n_steps * n_env where n_env is number of environment copies running in parallel)
+        NOTE: n_steps*n_env cannot be equal to 1, since that would imply a maximum training batch size of 1
```
araffin, line 34:
```suggestion
        (i.e. rollout buffer size is n_steps * n_envs where n_envs is number of environment copies running in parallel)
        NOTE: n_steps * n_envs must be greater than 1 (because of the advantage normalization)
        See https://github.com/pytorch/pytorch/issues/29372
```

stable_baselines3/ppo/ppo.py
```diff
@@ -115,7 +118,22 @@ def __init__(
                 spaces.MultiBinary,
             ),
         )
-
+        if self.env is not None:
+            # Checks for scenarios when we are collecting steps from a live environment (self.env not None)
+
+            total_env_steps = len(self.env.envs) * self.n_steps
```
araffin, line 124: Please use `env.num_envs`

stable_baselines3/ppo/ppo.py
```diff
@@ -115,7 +118,22 @@ def __init__(
...
+            assert total_env_steps > 1, (
+                f"n_steps*n_envs must be greater than 1. " f"Currently n_steps={self.n_steps} and n_envs={len(self.env.envs)}"
+            )
+            untruncated_batches = floor(total_env_steps / batch_size)
+            if total_env_steps % batch_size != 0:
+                warnings.warn(
+                    f"You have specified a desired batch size of {batch_size},"
```
araffin, line 131:
```suggestion
                    f"You have specified a mini-batch size of {batch_size},"
```

stable_baselines3/ppo/ppo.py
```diff
@@ -115,7 +118,22 @@ def __init__(
...
+                    f" but because your buffer of environment steps will be of size n_steps*n_envs, which"
```
araffin, line 132:
```suggestion
                    f" but because the `RolloutBuffer` is of size n_steps*n_envs, which"
```

stable_baselines3/ppo/ppo.py
```diff
@@ -115,7 +118,22 @@ def __init__(
...
+                    f" is currently {total_env_steps}, after every {untruncated_batches} untruncated batches,"
```
araffin, line 133: Maybe something simpler: the last mini-batch will be truncated and will be of size ... instead of ....

decodyng: My worry with that framing is that I think the notion of "last mini-batch" is potentially a bit unclear, since we mean it here in the context of "the last mini-batch of every buffer of environment steps," but I don't think that's necessarily the most intuitive parsing (my default, if I saw this error without the context I have, would be to assume it meant the last mini-batch in the whole of training). 

I also worry that explicitly using the framing of "the last mini-batch" to also cover scenarios where we know that this truncation will happen to _every_ mini-batch (because buffer size < batch size) would be extra confusing, since on a sort of language pragmatics level, I expect to only hear something referred to as the "last <thing>" if it's a situation where there's more than one <thing> present.

stable_baselines3/ppo/ppo.py
decodyng: Ah, right, I'd forgotten that functionality was still preserved there in Python3

decodyng:
I've added a test; wasn't sure what the right file was for it to live in, so I've stuck it in `test_utils` for now, but would be happy to move it if you think somewhere else is more appropriate

araffin (approved):
LGTM, thanks =)

PR #303, opened 2021-01-26
Add Support for Text Records to Logger, Add Hint on How To Access SummaryWriter in Docs.

### LINKED ISSUE #286: [Feature Request] Add Support for More Tensorboard Summary Formats (e.g. Text)
### 🚀 Feature

Add support for text (and potentially other formats) to TensorBoardOutputFormat. If there is some interest, I would be happy to create a PR.

### Motivation

Since torch offers a wide variety of tensorboard formats, It would make sense to offer some way of using most of these in stable baselines. In particular stable baselines should really support text to cater to the large community that is applying RL to Natural Language Processing.

### Pitch

Allowing text could be done by modifying the  TensorBoardOutputFormat.write(...) method. If we wanted to add support for more complex formats ([graphs](https://pytorch.org/docs/stable/tensorboard.html#torch.utils.tensorboard.writer.SummaryWriter.add_graph), [embeddings](https://pytorch.org/docs/stable/tensorboard.html#torch.utils.tensorboard.writer.SummaryWriter.add_embedding), etc.) it might be easier to just add a way for power users to directly access the underlying SummaryWriter (see alternatives below).

### Alternatives

More generally speaking it may make sense to allow users direct access to the summary writer in some cases. Right now the only way to access the writer that I have found is as follows (this is very hacky):

``` python
class Callback(BaseCallback):
  def add_text_record(self, tag: str, value: str, step: int) -> None:
     # could probably be made prettier with lambdas but there seems to be no way around iterating over the formatters.
      formatter_found = False
      for formatter in self.logger.Logger.CURRENT.output_formats:
          if isinstance(formatter, TensorBoardOutputFormat):
              formatter_found = True
              formatter.writer.add_text(tag, value, step)
              formatter.writer.flush()
  
      if not formatter_found:
          self.logger.warning("Could not log text to tensorboard. Could not find tb formatter in logger.")
```

### Checklist

lorenz-h:
## Description

Adds support for text records to ``JSONOutputFormat``, ``CSVOutputFormat``, ``HumanOutputFormat`` and ``TensorBoardOutputFormat``.  ``JSONOutputFormat`` and ``TensorBoardOutputFormat`` rely on the escaping done by ``json.dumps()`` and the ``SummaryWriter`` respectively. For the two other formatters we manually escape funky characters (e.g. color escapes) as follows: 
``` python
value = value.encode("unicode_escape").decode("utf-8")
```
To ignore any separators (,) within the text, it is wrapped in double quotes for the ``CSVOutputFormat``.
Secondly, as discussed in #286 I added a small section to the bottom of the docs/guide/tensorboard page, which describes how advanced users can directly access the Tensorboard SummaryWriter to log advanced metrics such as graphs and embeddings.

## Motivation and Context
Closes #286. stable-baselines3 should support text to cater to the large community that is applying RL to Natural Language Processing.

## Types of changes

## Checklist:


lorenz-h:
Surprisingly, one of the tests does not pass. As far as I can tell, this test should have nothing to do with the changes that I have made. @araffin could you verify if this test also fails for you?
```
FAILED tests/test_identity.py::test_discrete[env2-PPO] - AssertionError: Mean reward below threshold: 87.20 < 90.00
```

araffin:
>Surprisingly, one of the tests does not pass on my local machine (Ubuntu 18.04.5). As far as I can tell, this test should have nothing to do with the changes that I have made. @araffin could you verify if this test also fails for you?

it does not fail on my side, nor on GitHub ;) (and the failure not big: " 87.20" is very close to 90)

lorenz-h:
In that case - at least from my side - the PR is ready for review.

araffin:
>In that case - at least from my side - the PR is ready for review.

Almost, new tests are missing ;)

lorenz-h:
I added some test that validate the behaviour with some realistic unicode sequences (which include newlines, carriage returns, commas, semicolons etc.). Do you think this will be sufficient?

docs/guide/tensorboard.rst
```diff
@@ -225,3 +225,38 @@ Here is an example of how to render an episode and log the resulting video to Te
     model = A2C("MlpPolicy", "CartPole-v1", tensorboard_log="runs/", verbose=1)
     video_recorder = VideoRecorderCallback(gym.make("CartPole-v1"), render_freq=5000)
     model.learn(total_timesteps=int(5e4), callback=video_recorder)
+
+
+Directly Accessing The Summary Writer
+-------------------------------------
+
+If you would like to log arbitrary data (in one of the formats supported by torch.utils.tensorboard), you
```
araffin, line 233: maybe a link to https://pytorch.org/docs/stable/tensorboard.html would be better?

docs/guide/tensorboard.rst
```diff
@@ -225,3 +225,38 @@ Here is an example of how to render an episode and log the resulting video to Te
...
+
+    class SummaryWriterCallback(BaseCallback):
+        def _on_training_start(self):
+            output_formats = self.logger.Logger.CURRENT.output_formats
+            # Save reference to tensorboard formatter object
+            # note: the failure case (not formatter found) is not handled here, should be done with try/except.
+            self.tb_formatter = next(formatter for formatter in output_formats if isinstance(formatter, TensorBoardOutputFormat))
+
+        def _on_step(self) -> bool:
+            if self.n_calls % self._render_freq == 0:
```
araffin, line 256: where is `_render_freq` defined?

docs/guide/tensorboard.rst
```diff
@@ -225,3 +225,38 @@ Here is an example of how to render an episode and log the resulting video to Te
...
+                self.tb_formatter.writer.add_text("direct_access", "this is a value", self.n_calls)
```
araffin, line 257: `self.n_calls` or `self.n_timesteps`?

btw, did you test that code?

docs/misc/changelog.rst
```diff
@@ -27,6 +27,8 @@ New Features:
 - Wrap the environments automatically with a ``Monitor`` wrapper when possible.
 - ``EvalCallback`` now logs the success rate when available (``is_success`` must be present in the info dict)
 - Added new wrappers to log images and matplotlib figures to tensorboard. (@zampanteymedio)
+- Add support for text records to ``Logger``. (@lorenz-h)
+- Add hint on how to access the tensorboard summary writer directly to docs. (@lorenz-h)
```
araffin, line 31: the second entry should go under documentation ;)

stable_baselines3/common/logger.py
```diff
@@ -137,16 +137,22 @@ def write(self, key_values: Dict, key_excluded: Dict, step: int = 0) -> None:
...
+            elif isinstance(value, Figure):
                 raise FormatUnsupportedError(["stdout", "log"], "figure")
 
-            if isinstance(value, Image):
+            elif isinstance(value, Image):
                 raise FormatUnsupportedError(["stdout", "log"], "image")
 
-            if isinstance(value, float):
+            elif isinstance(value, str):
+                value = value.encode("unicode_escape").decode(
```
araffin, line 150: you should maybe define a function as it is re-used multiple times, no?

and it is common to log text with special characters?

tests/test_logger.py
```diff
@@ -114,6 +115,9 @@ def test_main(tmp_path):
     record("a", 5.5)
     dump()
     info("^^^ should see a = 5.5")
+    record("f", "this text \n \r should appear in one line")
```
araffin, line 118: I need to double check if this uses all the output formatter

araffin (commented):
Overall, LGTM, only some minor comments ;)

docs/guide/tensorboard.rst
lorenz-h: I usually use n_calls and it is what is used in [other parts of the docs](https://stable-baselines3.readthedocs.io/en/master/guide/examples.html?highlight=callback#using-callback-monitoring-training) ? I tested it but oviously but simplified it a bit too much before copying it to the docs. I'll test the exact code Ill push in the next commit.

docs/misc/changelog.rst
lorenz-h: Adressed in ffacc5f8af4c77cfea7b1b9c24ecbd7bb3bbbc4d

docs/guide/tensorboard.rst
lorenz-h: My bad, I forgot to add it. Adressed in ffacc5f8af4c77cfea7b1b9c24ecbd7bb3bbbc4d

docs/guide/tensorboard.rst
lorenz-h: Good point. Added it in ffacc5f8af4c77cfea7b1b9c24ecbd7bb3bbbc4d

stable_baselines3/common/logger.py
lorenz-h: Since we only use it twice I did not want to clutter up the `KVWriter` class. I can add it if you prefer.

docs/guide/tensorboard.rst
lorenz-h: Updated and fully tested the example in ffacc5f8af4c77cfea7b1b9c24ecbd7bb3bbbc4d .

araffin: so yes, n calls is used for checkpointing, and n timesteps is usually used for logging (to tensorboard notably)

stable_baselines3/common/logger.py
araffin: yes, please do, and outside the KVWriter probably

lorenz-h: The main idea behind this is to escape newlines, carriage returns and tabs that may be present in the original text.

lorenz-h: Moved to function in 474ce91344558b1f21da7a06ec3dcf17563eb616

docs/guide/tensorboard.rst
```diff
@@ -225,3 +225,43 @@ Here is an example of how to render an episode and log the resulting video to Te
...
+        def _on_training_start(self):
+            self._log_freq = 1000
+
+            output_formats = self.logger.Logger.CURRENT.output_formats
+            # Save reference to tensorboard formatter object
+            # note: the failure case (not formatter found) is not handled here, should be done with try/except.
+            self.tb_formatter = next(formatter for formatter in output_formats if isinstance(formatter, TensorBoardOutputFormat))
+
+        def _on_step(self) -> bool:
+            if self.num_timesteps % self._log_freq == 0:
```
araffin, line 261: sorry, I wasn't clear, `n_calls` is fine for checking when to log (otherwise it gets tricky with multiple envs).
However, I think num timesteps should be used to report to tensorboard (and to be able to compare to different runs with different number of envs ;))

To sum up:
- we need to change line 261 but keep line 262 as is ;)

lorenz-h: That's great to know - fixed in 4f65c7f5897415bab88127e6dfb57288e66bebec

docs/guide/tensorboard.rst
```diff
@@ -225,3 +225,43 @@ Here is an example of how to render an episode and log the resulting video to Te
...
+
+
+
+    model = SAC("MlpPolicy", "Pendulum-v0", tensorboard_log="/tmp/sac/", verbose=1)
+
+
+    class SummaryWriterCallback(BaseCallback):
+
+        def _on_training_start(self):
+            self._log_freq = 1000
```
araffin, line 253:
```suggestion
            self._log_freq = 1000  # log every 1000 calls
```
would be cleaner in the __init__ but I think it's ok for a quick example ;)

lorenz-h: absolutely - just wanted to keep the example short.

stable_baselines3/common/logger.py
```diff
@@ -300,13 +316,23 @@ def write(self, key_values: Dict[str, Any], key_excluded: Dict[str, Union[str, T
...
+            elif isinstance(value, Image):
                 raise FormatUnsupportedError(["csv"], "image")
 
-            if value is not None:
+            elif isinstance(value, str):
+                # escape any funky characters that may break the csv
+                value = escape_unicode(value)
+
+                # escape quotechars by prepending them with another quotechar
+                value = value.replace(self.quotechar, self.quotechar + self.quotechar)
```
araffin, line 330: this is new, no (I saw only the escape unicode before)? is that tested?

stable_baselines3/common/logger.py
```diff
@@ -336,7 +362,10 @@ def write(self, key_values: Dict[str, Any], key_excluded: Dict[str, Union[str, T
                 continue
 
             if isinstance(value, np.ScalarType):
-                self.writer.add_scalar(key, value, step)
+                if isinstance(value, str):
```
araffin, line 365: a `str` is a `np.ScalarType` ?

lorenz-h: I was surprised too. But [yes](https://stackoverflow.com/questions/21968643/what-is-a-scalar-in-numpy). I double checked this myself.

stable_baselines3/common/logger.py
lorenz-h: I tested it by hand and in automatic testing it should be covered by test_logger.py:read_log ? I added it in 474ce91344558b1f21da7a06ec3dcf17563eb616

araffin: but without the quotechar escape, shouldn't the test fail? (or it is just for better visualization?)
because your tests passed before that addition, no?

stable_baselines3/common/logger.py
araffin: interesting, but for clarity the if could be moved outside, no? (and replace the if by a elif for the scalar type)
or at least add a comment ;)

stable_baselines3/common/logger.py
lorenz-h: Just realized that the test passed because the test input doesn't contain a quotechar (duh). I'll add it.

stable_baselines3/common/logger.py
lorenz-h: I did not want to move it outside using elif in case someone at a later point changes the ordering of the elifs without realizing that `str` is a `np.ScalarType`.

lorenz-h: Added a comment in 8dbe360b84a88cadbae7ee6fb319aec823ffdf53

stable_baselines3/common/logger.py
lorenz-h: Added the new test input in 8dbe360b84a88cadbae7ee6fb319aec823ffdf53

araffin:
I did some tests and I'm not sure that the `escape_unicode` and other safety are needed (the tests pass when I remove them and they produce the same output...)
So either we remove those or we add a test that actually would fail without that safety...

Also, I realized that string logging was already supported for stdout (it is the default case in fact) and I removed the double quotes for clarity.

araffin:
You can ignore the failure on the CI due to a newer version of numpy, see https://github.com/DLR-RM/stable-baselines3/pull/245#issuecomment-770283257 ;)

lorenz-h:
Perhaps I was being overly cautious. In 101787c104147a5ff0baaba48e27f5c6406de249 you passed the raw string to the logger in test_logger.py:test_main. This might be a good solution. Hence, In b8f868489ff4bd2b089ee71e87ef4164887e3191 I replaced the unicode escaping with repr() for both the human and the csv output format. However, if you still think this excessive, I'll just remove the escaping entirely. This would mean that if users log newlines they may corrupt the log overviews that we print during training.

araffin:
>you passed the raw string to the logger in test_logger.py:test_main.

I did that because I wanted to break the logger (but did not managed ^^")

>However, if you still think this excessive, I'll just remove the escaping entirely.

I would favor `str` over `repr` as `str` is meant to be human readable vs `repr` is used for debug (see [example](https://www.geeksforgeeks.org/str-vs-repr-in-python/)).


In the meantime, I'll merge a fix for numpy soon ;) (see #307 )

lorenz-h:
I have removed `repr()` from `HumanOutputFormat` and `CSVOutputFormat`.

araffin (approved):
LGTM, thanks =)

Issue #172, opened 2020-10-01
[Enhancement] Support for fallbacks when loading cloudpickle items fail


Miffyli:
Related issue #171 

As seen in #171, cloudpickle/pickle can easily fail when transferring models between Python versions (and in case some other shenanigans). There is currently no way to address this issue as loading of model tries to un-pickle all pickle objects and does not catch errors.

Two suggestions to deal with this, both of which would be nice to have:
1) Provide `custom_objects` dictionary (similar to what we had in [SB2](https://stable-baselines.readthedocs.io/en/master/modules/a2c.html#stable_baselines.a2c.A2C.load)), which will override any objects read from file. This could be used to even prevent loading the item from the file, in which case all JSON-ed items can be used as before.
2) Update `set_parameters` function to only load parameters from the file but not the `data` objects (which include pickled items). This was included in SB2.

@ManifoldFR If you had other suggestions please feel free to throw them here :)

ManifoldFR:
From what I saw the problem I encountered is with the learning rate and clip ratio schedulers (for PPO), in the particular case where a constant scheduler is crated using `constant_fn` (I don't know about other cases though) because pickling/unpickling functions is brittle. I think we can look at how PyTorch supports saving state dicts for [schedulers](https://github.com/pytorch/pytorch/blob/master/torch/optim/lr_scheduler.py). From the source I see they use classes and the `__dict__` attribute to save/load state dicts.

ManifoldFR:
I started a branch for schedulers. I think we should *not* subclass from PyTorch's `_LRScheduler` straightaway to get the learning rate because they're designed to step with the optimizer.
I think we can take inspiration from their design though:
* use a class with `state_dict`/`load_state_dict` methods
* update the scheduler parameters using their `step()` method, update the param groups from there

Miffyli:
I wouldn't just focus on schedules though, as other objects are stored as pickled objects (no easy alternative). I am not sure having non-pickled versions of all objects is possible, and it would not be easy at the very least. I'd focus on the fallback tools to help reconstructing models if loading fails (part of the reason why currently the saved model contains some description of the pickled object). However if you can find something that could be used to achieve less-brittle serialization I am all ears!

aliamiri1380:
this can create a file from model, but it's too large

```
import dill
with open('nnn.pkl', 'wb') as f:
    dill.dump(model, f)
```

araffin:
[...]

Issue #171, opened 2020-10-01
Python 3.8: "an integer is required (got type bytes)" when loading models saved under older Python versions


ManifoldFR:
**Describe the bug**

There is an issue loading trained agent files in Python 3.8.

**Code example**

I checked this on the `rl-baselines3-zoo` repository.

```bash
(base) manifold$ python enjoy.py --algo sac --env Pendulum-v0
Loading latest experiment, id=1
Traceback (most recent call last):
  File "enjoy.py", line 225, in <module>
    main()
  File "enjoy.py", line 141, in main
    model = ALGOS[algo].load(model_path, env=env, **kwargs)
  File "/home/manifold/stable-baselines3/stable_baselines3/common/base_class.py", line 561, in load
    data, params, pytorch_variables = load_from_zip_file(path, device=device)
  File "/home/manifold/stable-baselines3/stable_baselines3/common/save_util.py", line 390, in load_from_zip_file
    data = json_to_data(json_data)
  File "/home/manifold/stable-baselines3/stable_baselines3/common/save_util.py", line 164, in json_to_data
    deserialized_object = cloudpickle.loads(base64_object)
TypeError: an integer is required (got type bytes)
```

**System Info**
Describe the characteristic of your environment:
 * SB3 downloaded from git trunk
 * Python version 3.8.5
 * PyTorch version 1.6.0
 * Gym version: latest from git
 * Cloudpickle 1.6.0

**Additional context**

I printed out info and saw that it occurs when loading the data key `learning_rate` on the SAC example above. For policies trained using recent versions of SB3 it happens for `lr_schedule`.

ManifoldFR:
After further investigation, this also applies to `clip_range` in PPO.

I think the common point is that these can all be callables, right?

Miffyli:
Does the same happen with saving/loading models with stable-baselines3 alone without the files in zoo? The code for saving and loading has changed since those zoo models were trained so it might have broken a thing or two.

ManifoldFR:
It happens for models trained using a version of SB3 pulled from Git a week ago, albeit with Python 3.7

Miffyli:
Just to clarify: You train a model with older version of SB3 (week ago), save it, update to current SB3 (master) and loading fails?

araffin:
> Just to clarify: You train a model with older version of SB3 (week ago), save it, update to current SB3 (master) and loading fails?

the issue comes with the pretrained models  (using python 3.6) that are available in the zoo.

[...]

PR #336, opened 2021-03-01
Add custom objects support + bug fix

araffin:
## Description

Addresses part of #171 and #172

Also fixes a bug when using images and DQN at test time.

Hacks for loading with python 3.8+ things saved in python 3.6/3.7:
```python
  # For python3.8+
  custom_objects = {
      "learning_rate": 0.0,
      "lr_schedule": lambda _: 0.0,
      "clip_range": lambda _: 0.0,
  }
```

## Motivation and Context

## Types of changes

## Checklist:


tests/test_cnn.py
```diff
@@ -25,14 +25,25 @@ def test_cnn(tmp_path, model_class):
...
     assert is_vecenv_wrapped(model.get_env(), VecTransposeImage)
 
     obs = env.reset()
 
+    # Test stochastic predict with channel last input
+    if model_class == DQN:
+        model.exploration_rate = 0.9
+
+    for _ in range(10):
+        model.predict(obs, deterministic=False)
```
Miffyli, line 45: A curiosity: Why 10 steps, and why not 10 steps for deterministic=True? Just to be on save side we could do 10 steps for the True case as well (in case first `predict` modifies something and breaks next one)

tests/test_save_load.py
```diff
@@ -183,17 +183,17 @@ def test_set_env(model_class):
     # create model
     model = model_class("MlpPolicy", env, policy_kwargs=dict(net_arch=[16]), **kwargs)
     # learn
-    model.learn(total_timesteps=300)
```
Miffyli, line 186: A sidenote: Maybe we should invest bit of time to set all tests to default to "cpu" by default? All the cuda-initialization takes time and GPUs slow down small-network agents. Maybe leave "auto" for CnnPolicy tests.

Miffyli:
Just few notes but other than that LGTM. Kind of funny the custom_object support was already there but we just did not pass the parameters forward :')

tests/test_cnn.py
araffin: pretty arbitrary, this is mostly to avoid any regression on the bug I fixed for DQN, and as I set the exploration rate to 0.9, I still to sample the predict method enough times to be sure that the random action part will be called...

tests/test_save_load.py
araffin: doing such thing would prevent us from finding bugs that are due to cuda only... (for instance a tensor that is not moved to the correct device)

araffin:
>Kind of funny the custom_object support was already there but we just did not pass the parameters forward :')

yep, I was quite surprised when I discovered that trying to load model with python 3.8

Miffyli (approved):
LGTM with the above comments handled

Rocamonde:
I feel like this should be documented somewhere. Loading the standard huggingface models from sb3 with the textbook examples gives the AssertionError:

```python
env = gym.make('LunarLander-v2')
policy = A2C.load('./models/a2c/LunarLander-v2_1/LunarLander-v2.zip', env=env)
```

Totally not obvious that one should write instead
```python
env = gym.make('LunarLander-v2')
policy = A2C.load('./models/a2c/LunarLander-v2_1/LunarLander-v2.zip', env=env, custom_objects={'learning_rate': 0.0})
```

The error message should be changed and this should be explained in the guide IMO (or just update the huggingface models).

qgallouedec:
Which SB3/python version do you use? I'm pretty sure I've fixed this in #1123.

Rocamonde:
@qgallouedec Oh okay great! My version is at least a month old, due to dependencies. Good that this has already been looked into! 

Also, apologies that I have not been looking into some pending PRs for sb3, I have been quite busy with work recently. I'll aim to get up to speed this week and see if there's anything I can still help with.

Issue #297, opened 2021-01-19
[Feature Request] (V1.1+) Unified framework for multi-goal algorithms


ferreirajoaouerj:
I saw that `dict` support is almost ready, I think it's worth considering this framework, for V1.1+, before a new `HER` implementation appears.

All credits goes to @spitis, @Takonan, et al., for proposing this framework in the awesome paper [Maximum Entropy Gain Exploration for Long Horizon Multi-goal Reinforcement Learning](https://arxiv.org/pdf/2007.02832.pdf).

### 🚀 Feature

Most of the algorithms that emerged after HER differ in 3 functions:

1. **select(*args)**: chooses which `desired_goal` will be used by the agent in each episode (HER uses the `desired_goal` given by the environment).
2. **reward(state, action, achieved_goal)**: how to calculate the reward (HER uses the `compute_reward` method directly).
3. **relabel(buffer, *args)**: how data/trajectory augmentation will occur (HER's best strategy uses future `achieved_goals`).

Each of this components have massive consequences in the algorithm performance (see figure attached), especially for long-horizon tasks.

### Motivation

It can be easy to implement new multi-goal algorithms just by changing these 3 functions. 

### Pitch

Implement the API through functions or methods.

### Alternatives

1. Keep it as it is.
2. Implement a subset (i.e. don't mess with **reward**).

### Additional context

The figure below is an experiment (source: [Maximum Entropy Gain Exploration for Long Horizon Multi-goal Reinforcement Learning](https://arxiv.org/pdf/2007.02832.pdf)) comparing different algorithms for a maze navigation task.

![image](https://user-images.githubusercontent.com/63377970/105107392-c7495c00-5a96-11eb-9a76-1158fd3426f6.png)

### Checklist

- [x]  I have checked that there is no similar [issue](https://github.com/DLR-RM/stable-baselines3/issues) in the repo (**required**)

araffin:
Hello,

thank you for raising this issue.
I'm aware of the MEGA paper (although I did not read it thoroughly) and that would be a good addition for the [contrib repo](https://github.com/Stable-Baselines-Team/stable-baselines3-contrib) ;)

>Most of the algorithms that emerged after HER differ in 3 functions:

this would be part of the buffer no? As the rest would look the same from the algorithm perspective (as we are only relabeling transitions).

>select(*args): chooses which desired_goal will be used by the agent in each episode (HER uses the desired_goal given by the environment).

[...]

PR #351, opened 2021-03-15
Refactor HER

### LINKED ISSUE #284: [Bug] Infinize horizon tasks are handled like episodic tasks

araffin:
**NOTE: this branch will be merged with #243 not master**

## Description

Refactoring of HER (need to account for #297 ) using proper obs dict support.
We could also do the timelimit support (see #284 ) as we are breaking the replay buffer api (we need to provide upgrade script)

closes https://github.com/DLR-RM/stable-baselines3/issues/287 (image support for HER)
closes #284 (for off-policy algorithms)

TODO:
- [x] Check HER performance
- [x] Investigate offline replay buffer speed drop
- [x] Add documentation
- [x] Perf check for HER + TimeLimit handling (PyBullet)

## Motivation and Context

## Types of changes

## Checklist:


araffin:
@ferreirajoaouerj could you take a look at this PR? (this shoudl address your concern from https://github.com/DLR-RM/stable-baselines3/issues/297).
`HER` is now just a replay buffer class, so implementing multi-goal algorithms should be simpler, as one would need to define only a new buffer class.

ferreirajoaouerj:
I'll take a look, thank you for addressing this feature.

ferreirajoaouerj:
Sorry the delay. Nice that you managed to put everything into the replay buffer. The `select` in #297 function needs to change the `observation['desired_goal']` at the beginning of the episode. I was thinking about the least intrusive way to do that. Maybe the best option is to add an env wrapper, then is easy to also implement `reward` and we don't need to mess with `collect_rollouts` for each off-policy algo. Anyway, we could explore these options in `contrib` since the baseline HER is already up and running.

araffin:
Thanks for taking a look =)

>he select in #297 function needs to change the observation['desired_goal'] at the beginning of the episode. I was thinking about the least intrusive way to do that. Maybe the best option is to add an env wrapper, then is easy to also implement reward and we don't need to mess with collect_rollouts for each off-policy algo. 

You are right, an env wrapper would be the way to go for changing/providing the desired goal at the beginning of an episode.

For the reward, the `compute_reward` is only used in the `HerReplayBuffer` so it should be quite easy to overwrite ;)

PR #363, opened 2021-03-24
Fix default arguments + add bugbear

araffin:
## Description

Fixed potential issue when calling off-policy algorithms with default arguments multiple times (the size of the replay buffer would be the same)
and add bug bear extension to test requirements.

## Motivation and Context

## Types of changes

## Checklist:


docs/misc/changelog.rst
```diff
@@ -15,12 +15,14 @@ New Features:
 
 Bug Fixes:
 ^^^^^^^^^^
+- Fixed potential issue when calling off-policy algorithms with default arguments multiple times (the size of the replay buffer would be the same)
```
Miffyli, line 18: Bit of clarification: does this happen when off-policy algo is constructed, and in what kind of scenario? This is more of a question for the thread, but could also be updated here in changelog.

araffin: For a concrete example:

```python
class A:
    def __init__(self, a=int(1e6)):
        self.a = a


obj = A()
obj2 = A()

print(id(obj.a), id(obj2.a))
# obj.a and obj2.a are pointing to the same object
assert id(obj.a) == id(obj2.a)
```

In our very special case, it is ok-ish because `buffer_size` is an int (so if you modify the value, it won't modify the value of the other object, but it is just a bad practice that can be very painful to debug.

Miffyli (approved):
Thanks for the clarification! LGTM :)

PR #375, opened 2021-03-31
Add test for GAE + rename `RolloutBuffer.dones` for clarification

araffin:
## Description

- Added a test for GAE computation
- Add comment for return computation
- Renamed `dones` for clarity 

## Motivation and Context

see #105

## Types of changes

## Checklist:


araffin:
Note: after some quick trials on `CartPole-v1`, the formulation `self.returns = self.advantages + self.values` seems to give me consistently better results.

Need to be confirmed with some Breakout/Pong run I think (or any harder env, maybe `HalfCheetahBullet` would do the trick)

Miffyli:
Could you provide references on what would be the "classical" way of computing returns you mention? While yes, this is not what you usually see, the one used inside GAE computation should be theoretically valid (e.g. if you set `gae_lambda` to one far end, you get the normal, discounted return).

stable_baselines3/common/base_class.py
```diff
@@ -130,7 +130,7 @@ def __init__(
         self.tensorboard_log = tensorboard_log
         self.lr_schedule = None  # type: Optional[Schedule]
         self._last_obs = None  # type: Optional[np.ndarray]
-        self._last_dones = None  # type: Optional[np.ndarray]
+        self._episode_starts = None  # type: Optional[np.ndarray]
```
Miffyli, line 133: I would name this `_last_episode_starts` for consistency, bit verbose, but at least tells its meaning :)

docs/misc/changelog.rst
```diff
@@ -4,18 +4,20 @@ Changelog
 ==========
 
 
-Release 1.1.0a1 (WIP)
+Release 1.1.0a2 (WIP)
 ---------------------------
 
 Breaking Changes:
 ^^^^^^^^^^^^^^^^^
+- renamed ``_last_dones`` and ``RolloutBuffer`` ``dones`` to ``_episode_starts`` and ``episode_starts``
```
Miffyli, line 12:
```suggestion
- Renamed ``_last_dones`` and ``dones`` to ``_episode_starts`` and ``episode_starts`` in ``RolloutBuffer``.
```

araffin:
>Could you provide references on what would be the "classical" way of computing returns you mention?

- GAE paper: equation 28
https://arxiv.org/abs/1506.02438

- Sutton Barto book: "9.3.   STOCHASTIC-GRADIENT AND SEMI-GRADIENT METHODS": both MC adn TD(0) estimates are shown


I guess using the GAE(lambda) + Value ends up being TD(lambda) target, so it should be fine but need a comment, no?

araffin:
Looking at the original commit (https://github.com/openai/baselines/commit/da997060461e3cbf54ca4dc7a67081a731fb6b3b), the comment says "Compute target value using TD(lambda) estimator, and advantage with GAE(lambda)", so yes it is using TD(lambda) and not MC for value estimation.

The line of code:
https://github.com/openai/baselines/blob/ea25b9e8b234e6ee1bca43083f8f3cf974143998/baselines/ppo1/pposgd_simple.py#L66

Miffyli:
@araffin 

That is very good to know, and also good to know your TD(0) experiment (as I understood it) had lower results (which was to be expected, as I understood TD(lambda) should work better) 👍 . A comment on this would make things clear, and maybe even clarify that this is used because it was used by openai baselines and SB2.

araffin:
>That is very good to know, and also good to know your TD(0) experiment (as I understood it) 

I tested MC estimates (as it is the most common way usually)

>that this is used because it was used by openai baselines and SB2.

this is in the docstring but not so clear. I will update it ;)

araffin:
So, I took some time and found the answer in David Silver lecture 4, slide 47 (page 51 in the PDF) "Telescoping in TD(λ)"), it comes from an non-intuitive telescope: https://www.davidsilver.uk/wp-content/uploads/2020/03/MC-TD.pdf

He shows that $G(\lambda) - V(s)$ lead to the GAE formula.

>That is very good to know, and also good to know your TD(0) experiment (as I understood it) had lower results (which was to be expected, as I understood TD(lambda) should work better

TD(1) is MC estimate (non-intuitive when looking at the definition)
TD(0) is one-step estimate with bootstrapping

Miffyli:
Added some clarifications to docs. Two major comments:

1) The GAE-test is to ensure GAE-computation works out correctly, right? Two worries here:
    a) What is the implementation in test based on (i.e. can we trust it?)
    b) I tried changing the code to compute GAE over three episodes (one rollout lasts for `max_steps * 3`). Only first two tests pass, but remaining do not. I figure this should pass if the implementations were correct?
2) Should `dones` in `compute_returns_and_advantages` arguments be renamed to `last_episode_starts`? This is to clarify `dones` and `self.episode_starts` have same meaning, but sadly this mean updating code around the repository to reflect this change.

araffin:
> a) What is the implementation in test based on (i.e. can we trust it?)

Different things. As you noticed, I took the idea from https://github.com/DLR-RM/stable-baselines3/issues/105, so that if there is an off-by-one error, the test will fail (I did some testing trying to re-introduce the bug and it failed as expected).

Then, I did some unit testing (GAE(1) is MC estimate) while writing the test (+ manual debugging).
For the lambda-return implementation, I based it on SB3... mostly because we did a lot of testing with that code (and if it is wrong, well, it is wrong but it works ^^").

> b) I tried changing the code to compute GAE over three episodes (one rollout lasts for max_steps * 3). Only first two tests pass, but remaining do not. I figure this should pass if the implementations were correct?

Good point. I think I did not write the test for more than one episode, and in fact, it should probably fail without modification (that was to keep things simple and easy to monitor).

>hould dones in compute_returns_and_advantages arguments be renamed to last_episode_starts?

I did not rename it because I think it has a different role compared to the episode start (is the very last step of the rollout terminal?), and it is used only once (I think I wrote that in the docstring).

Miffyli:
> Good point. I think I did not write the test for more than one episode, and in fact, it should probably fail without modification (that was to keep things simple and easy to monitor).

Since main code should support this, I think the test should also check for this. In addition, I think it should also test other reward models than "last step rewards 1.0" (e.g. "first step rewards 1.0" and "all steps reward 1.0"), to cover edge-cases and most common cases. I have framework for this and multi-episode test (which fail), which I can push if you'd like.

> I did not rename it because I think it has a different role compared to the episode start (is the very last step of the rollout terminal?), and it is used only once (I think I wrote that in the docstring).

Ah right, now I see it 👍 . You are right they have different meanings. Lets leave it as is.

araffin:
>I have framework for this and multi-episode test (which fail), which I can push if you'd like.

I would prefer if you push tests that pass :p

Miffyli:
Sorry for taking this long! I have added a test for multi-episode rollout that passes. If it checks out, feel free to accept and merge 👍

Issue #90, opened 2020-07-07
[Question/Discussion] Comparing stable-baselines3 vs stable-baselines


AlessandroZavoli:
Did anybody compare the training speed (or other performance metrics) of SB and SB3 for the implemented algorithms (e.g., PPO?)
Is there a reason to prefer either one for developing a new project?

PartiallyTyped:
SB3 is in active development whereas SB2(SB) is in maintenance mode. I use SB3 for my projects since it is more modular and less cluttered than SB2 simply because of dynamic computation graphs, the experience collected by the members in implementing the algorithms, and a well thought out design.

Miffyli:
To add to comment above, some of the methods are as of writing slower (at least without tuning e.g. number of threads), but we are still in process of going over them and optimizing for speed and matching the performance of SB2 implementations.

araffin:
Hello,

I'm glad that you ask ;)

As mentioned by @PartiallyTyped , SB3 is now the project actively developed by the maintainers.
It does not have all the features of SB2 (yet) but is already ready for most usage.

>Did anybody compare the training speed (or other performance metrics) of SB and SB3 for the implemented algorithms (e.g., PPO?)

We have two related issues for that: #49 #48 
The algorithms have been benchmarked recently in a [paper](https://arxiv.org/abs/2005.05719) for the continuous case and I have already successfully used SAC on real robots.
Because PyTorch uses dynamic graph, you have to expect a small slow down (we plan to use the jit improve the speed in the future #57 ) and you may have to plan with `torch.set_num_thread()` to have the best speed. One exception is DQN which is significantly faster in SB3 because of the new replay buffer implementation.

> Is there a reason to prefer either one for developing a new project?

The main advantage of SB3 is that it was re-built (almost) from scratch, trying not to reproduce the errors made in SB2.
That means much clearer code, more test coverage and higher quality standard (with the use of typing notably). 
Unless you need to use RNN, I would highly recommend you to use SB3.

If you change the internals, you may expect some changes (they will be documented anyway) until the v1.0 is released (see issue #1 and code review #17 ).
If you use only the "user api" (without changing the internals), then not much should change and I would highly recommend you to use the [rl zoo](https://github.com/DLR-RM/rl-baselines3-zoo) that should cover most needs (and that is up to date with the best practices for using SB3).

It is also in the roadmap to document the differences between SB2 and SB3.

[...]

PR #405, opened 2021-04-24
increase default ppo batch size from 64 to 256

matwilso:
PPO can work poorly with small batch sizes, so 256 is probably a better default for people than 64. Also, 256 is the old default batch size in the original stable baselines (https://github.com/DLR-RM/stable-baselines3/issues/90#issuecomment-742525593), as well SAC currently.

I thought this was a minor enough change not to require the normal PR process. But let me know if you want me to go through that.

### DIFF
--- a/stable_baselines3/ppo/ppo.py
+++ b/stable_baselines3/ppo/ppo.py
@@ -71,7 +71,7 @@ class PPO(OnPolicyAlgorithm):
         env: Union[GymEnv, str],
         learning_rate: Union[float, Schedule] = 3e-4,
         n_steps: int = 2048,
-        batch_size: Optional[int] = 64,
+        batch_size: Optional[int] = 256,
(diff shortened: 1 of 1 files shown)


Miffyli:
Hello. Please open up an issue discussing this change. The "non-issue PRs" are only reserved for hotfixes and very small updates like documentation typo fixes. This would change the behaviour of existing codebases, so it needs to be properly discussed.

araffin:
Hello,

I'm not sure where you have seen that SB2 PPO had 256 as default batch size (the comment you linked was only valid for a particular example).
Make sure to read our migration guide: https://stable-baselines3.readthedocs.io/en/master/guide/migration.html
and afterward open an issue (as mentioned by @Miffyli ) if you think there is still one.

PS: please read our contributing guide and keep the PR template next time.

Issue #109, opened 2020-07-16
Allow for logger setup before calling BaseAlgorithm.learn()


qxcv:
**Describe the bug**

At present, SB3 algorithms automatically configure the logger on the first call to `{On,Off}PolicyAlgorithm.learn()`. Unfortunately, this prevents the logger from being configured manually beforehand—if the logger is already configured, then the first call to `.learn()` will overwrite that old configuration. This is a problem for code bases that wish to use the SB3 logger for both SB3 RL algorithms and for other non-SB code (for example, we use the SB logger for both discriminator training and RL in the GAIL implementation from [imitation](https://github.com/HumanCompatibleAI/imitation/)). Ideally, custom logger configuration should be moved out of the `.learn()` method, or the logger configuration update inside `.learn()` should be made optional via a keyword argument to `PPO`.

**Code example**

The following code demonstrates the issue I am trying to overcome:

```python
import gym
import stable_baselines3
from stable_baselines3.common import logger 

# show initial config
print("Logger outputs at startup:", logger.Logger.CURRENT.output_formats)

# set up logger
logger.configure("./path/to/my/logs", ["stdout", "tensorboard"])
print("Logger outputs before training:", logger.Logger.CURRENT.output_formats)

# now instantiate & train
env = gym.make("CartPole-v1")
algorithm = stable_baselines3.PPO("MlpPolicy", env)
algorithm.learn(32)
print("Logger outputs after training:", logger.Logger.CURRENT.output_formats)
```

Current output:

```
Logger outputs at startup: [<stable_baselines3.common.logger.HumanOutputFormat object at 0x7f553f7a3550>]
Logging to ./path/to/my/logs
Logger outputs before training: [<stable_baselines3.common.logger.HumanOutputFormat object at 0x7f553e938d10>, <stable_baselines3.common.logger.TensorBoardOutputFormat object at 0x7f55abd95f90>]
Logger outputs after training: []
```

Instantiating PPO and calling `.learn()` clears the logger configuration. The desired output is instead:

```
Logger outputs at startup: [<stable_baselines3.common.logger.HumanOutputFormat object at 0x7f553f7a3550>]
Logging to ./path/to/my/logs
Logger outputs before training: [<stable_baselines3.common.logger.HumanOutputFormat object at 0x7f553e938d10>, <stable_baselines3.common.logger.TensorBoardOutputFormat object at 0x7f55abd95f90>]
[...]

PR #407, opened 2021-04-25
Use Python logger instead of print

### LINKED ISSUE #128: Use python logger instead of print
Currently, we use `self.verbose` and `print` for logging and debug info but a cleaner way would be to use the logging package from python.

tfurmston:
## Description

This PR addresses issue 128, [https://github.com/DLR-RM/stable-baselines3/issues/128](https://github.com/DLR-RM/stable-baselines3/issues/128). 

To do so, I have introduced the use of [https://environ-config.readthedocs.io/en/stable/index.html](environ-config), which is used to handle the environment settings, such as logging config.

## Motivation and Context

closes #128 

## Types of changes

## Checklist:

- [] My change requires a change to the documentation.

Note: I have not done the following yet, with reasons provided:

- Changelog has yet to be updated. I am awaiting confirmation that the contributions in this PR are acceptable.
- `make type` seems to be failing, but it is not clear it is related to my changes. I will have a look into this once I have some feedback on the contributions themselves.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -82,6 +82,7 @@ setup(
         "pandas",
         # Plotting learning curves
         "matplotlib",
+        "environ-config>=~20.1",
     ],
     extras_require={
         "tests": [
--- a/stable_baselines3/__init__.py
+++ b/stable_baselines3/__init__.py
@@ -1,3 +1,4 @@
+import logging.config
 import os
 
 from stable_baselines3.a2c import A2C
@@ -6,7 +7,10 @@ from stable_baselines3.dqn import DQN
 from stable_baselines3.her import HER
 from stable_baselines3.ppo import PPO
 from stable_baselines3.sac import SAC
+from stable_baselines3.settings import settings
 from stable_baselines3.td3 import TD3
+
+logging.config.dictConfig(config=settings.LOGGING)
 
 # Read version from file
 version_file = os.path.join(os.path.dirname(__file__), "version.txt")
--- a/stable_baselines3/a2c/a2c.py
+++ b/stable_baselines3/a2c/a2c.py
@@ -45,7 +45,6 @@ class A2C(OnPolicyAlgorithm):
     :param create_eval_env: Whether to create a second environment that will be
         used for evaluating the agent periodically. (Only available when passing string for the environment)
     :param policy_kwargs: additional arguments to be passed to the policy on creation
-    :param verbose: the verbosity level: 0 no output, 1 info, 2 debug
     :param seed: Seed for the pseudo random generators
     :param device: Device (cpu, cuda, ...) on which the code should be run.
         Setting it to auto, the code will be run on the GPU if possible.
@@ -71,7 +70,6 @@ class A2C(OnPolicyAlgorithm):
         tensorboard_log: Optional[str] = None,
         create_eval_env: bool = False,
         policy_kwargs: Optional[Dict[str, Any]] = None,
-        verbose: int = 0,
         seed: Optional[int] = None,
         device: Union[th.device, str] = "auto",
         _init_setup_model: bool = True,
@@ -91,7 +89,6 @@ class A2C(OnPolicyAlgorithm):
             sde_sample_freq=sde_sample_freq,
             tensorboard_log=tensorboard_log,
             policy_kwargs=policy_kwargs,
-            verbose=verbose,
             device=device,
             create_eval_env=create_eval_env,
             seed=seed,
--- a/stable_baselines3/common/base_class.py
+++ b/stable_baselines3/common/base_class.py
@@ -38,7 +38,6 @@ from stable_baselines3.common.vec_env import (
 )
 from stable_baselines3.common.vec_env.obs_dict_wrapper import ObsDictWrapper
 
-
 _logger = logging.getLogger(__name__)
 
 
--- a/stable_baselines3/common/base_class.py
+++ b/stable_baselines3/common/base_class.py
@@ -1,6 +1,7 @@
 """Abstract base classes for RL algorithms."""
 
 import io
+import logging
 import pathlib
 import time
 from abc import ABC, abstractmethod
@@ -38,16 +39,17 @@ from stable_baselines3.common.vec_env import (
 from stable_baselines3.common.vec_env.obs_dict_wrapper import ObsDictWrapper
 
 
-def maybe_make_env(env: Union[GymEnv, str, None], verbose: int) -> Optional[GymEnv]:
+_logger = logging.getLogger(__name__)
+
+
+def maybe_make_env(env: Union[GymEnv, str, None]) -> Optional[GymEnv]:
     """If env is a string, make the environment; otherwise, return env.
 
     :param env: The environment to learn from.
-    :param verbose: logging verbosity
     :return A Gym (vector) environment.
     """
     if isinstance(env, str):
-        if verbose >= 1:
-            print(f"Creating environment from the given name '{env}'")
+        _logger.info(f"Creating environment from the given name '{env}'")
         env = gym.make(env)
     return env
 
@@ -64,7 +66,6 @@ class BaseAlgorithm(ABC):
         it can be a function of the current progress remaining (from 1 to 0)
     :param policy_kwargs: Additional arguments to be passed to the policy on creation
     :param tensorboard_log: the log location for tensorboard (if None, no logging)
-    :param verbose: The verbosity level: 0 none, 1 training information, 2 debug
     :param device: Device on which the code should run.
         By default, it will try to use a Cuda compatible device and fallback to cpu
         if it is not possible.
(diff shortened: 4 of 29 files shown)


araffin (commented):
Hello,
thanks for the PR.
As a general remark, please do not make any backward incompatible change (and you should not have to change the tests normally).
It seems you also updated the internal logger, which is outside the scope of this PR.

As raised by @Miffyli , I'm not sure about the usefulness of an additional dependency (it look like what you are doing can be done using `os.environ` variable (aka env variable).

tfurmston:
OK, so it seems I have misunderstood the purpose/scope of the issue. I thought it was to introduce standard python logging, which my PR provides a pretty standard example. (If you want users to be able to configure the logging, then I think a centralised settings it a pretty standard approach too.) However, it seems that this is not the scope of the issue. I'm going to close the PR.

> As a general remark, please do not make any backward incompatible change (and you should not have to change the tests normally).

Given the scope of the issue, as I understood it, I don't see how you can introduce a standard logging whilst not having to change the tests. In particular, any sensible use of standard logging would not pass a `verbose` parameter around, and hence I removed it and updated the tests.

> As raised by @Miffyli , I'm not sure about the usefulness of an additional dependency (it look like what you are doing can be done using os.environ variable (aka env variable).

Some of the advantages of this library are listed on the page, [https://pypi.org/project/environ-config/](https://pypi.org/project/environ-config/)

araffin:
>OK, so it seems I have misunderstood the purpose/scope of the issue. I thought it was to introduce standard python logging, which my PR provides a pretty standard example

to clarify things, we currently have two "loggers" inside SB3, one is used for info/debug about SB3 internal and is controlled using the verbose argument, one is under `common/logger.py` and is used to report agent training progress.

> It seems you also updated the internal logger, which is outside the scope of this PR.

I made that remark probably too quickly...

>Some of the advantages of this library are listed on the page, https://pypi.org/project/environ-config/

I do not doubt that such library can be very useful when dealing with configs, however, here we have only one parameter (`verbose`), so it sounds a bit overkill to add a dependency just for it.

>Given the scope of the issue, as I understood it, I don't see how you can introduce a standard logging whilst not having to change the tests. In particular, any sensible use of standard logging would not pass a verbose parameter around, and hence I removed it and updated the tests.

tests need to be changed with care, because in some cases we test loading/saving of parameters, not necessary the logging system. Keeping the tests as is would also allow to spot any backward incompatible change.

>I'm going to close the PR.

my first comments was probably a bit rough (and only after taking a quick look at what you did)...
I really appreciate the work you have done and would like it to go towards a mergable PR ;)
My first remark were mostly here to prevent you from going to far away from what we would expect and saving you some time.

tfurmston:
OK, I thought my PR way off topic somehow. If you feel it is worthwhile, then I am very happy to finish off the PR.

Other than the extra dependency, is there anything that needs to be resolved? If so, can you please let me know. 

I would suggest that there is still a `settings` module in which all of the required environment variables in the project are collated into a single settings class. This will centralise the settings, it just won't have the additional benefits of environ-config or the additional dependency. Are you happy with this approach?

tfurmston:
As a passing comment, I know it is out of scope for this PR, is the purpose of the "logger" in `common/logger.py` simply to keep the model metrics logging separate from the other logging?

Miffyli:
> Other than the extra dependency, is there anything that needs to be resolved? If so, can you please let me know.

To focus on the `self.verbose` and use of `print` to log things, I would revert the changes on `common/logger.py` things.

> I would suggest that there is still a settings module in which all of the required environment variables in the project are collated into a single settings class. This will centralise the settings, it just won't have the additional benefits of environ-config or the additional dependency. Are you happy with this approach?

Something like this might be added in future if we start to need more settings, but it should not be added for this single thing prematurely. 

An ideal and backwards-compatible change would something like this:
- `verbose` parameter sets the level of Python logging shown (e.g. using `logging.setLevel` function to omit debug level messages when `verbose < 2`).
- Replace all `if verbose > #: print("stuff")` things with loggers.

> As a passing comment, I know it is out of scope for this PR, is the purpose of the "logger" in common/logger.py simply to keep the model metrics logging separate from the other logging?

Yes, model logging tool is more specific to showing key-value pairs, and outputting them in different form for later processing and/or human consumption.

tfurmston:
> Something like this might be added in future if we start to need more settings, but it should not be added for this single thing prematurely.

There are actually other settings that are in other parts of the code base already. In particular, `SB3_LOGDIR` and `SB3_LOG_FORMAT`, so it would be more that one setting. It would simply be a class that loads all of the environment variables in on place and can then be used across the rest of the project. It's introduction would be pretty minimal, I think. I can remove it if you want.

> An ideal and backwards-compatible change would something like this:
>  verbose parameter sets the level of Python logging shown (e.g. using logging.setLevel function to omit debug level messages when verbose < 2).
>     Replace all if verbose > #: print("stuff") things with loggers.

This change would not be backwards-compatible. Introducing configurable standard python logging will require people to setting an environment variable for the logging level and also for the removal of the `verbose` parameter. Both of these are not backwards-compatible. 

As I mentioned before, I don't think there is a sensible way to introduce configurable standard python logging in a way that does not require these changes, i.e., that is backwards-compatible. If you don't want a backward-incompatible change in this functionality at the moment, which is completely understandable, then I suggest the issue/PR is closed for the moment and the changes better scoped in a new issue.

araffin:
>There are actually other settings that are in other parts of the code base already. In particular, SB3_LOGDIR and SB3_LOG_FORMAT, so it would be more that one setting. 

for those, I would in fact remove them (they are currently not used and overwritten when creating the logger), best would be to allow the user to change that through a method call (we have the issue for that: https://github.com/DLR-RM/stable-baselines3/issues/109, probably for a separate PR).

tfurmston:
> for those, I would in fact remove them (they are currently not used and overwritten when creating the logger), best would be to allow the user to change that through a method call (we have the issue for that: #109, probably for a separate PR).

OK, I see. Thanks for clarifying. In that case, if you want people to be able to configure the logging level, then we could just 
load the environment variable in `stable_baselines3/__init__.py` when the logger is configured, and not have a settings class at all. Would that be ok?

Miffyli:
> Introducing configurable standard python logging will require people to set an environment variable for the logging level and also require the removal of the verbose parameter. Both of these are not backwards-compatible.

What I meant was to keep the verbose parameter (for now), but it only sets the logging level under the hood (e.g. in the `BaseAlgorithm`. User can also set the logging level prior to creating the algorithm with `logging` library, or by the single environment variable. I am not sure which should overwrite which one.

@araffin can you give a final opinion how this should be done? :)

tfurmston:
> What I meant was to keep the verbose parameter (for now), but it only sets the logging level under the hood (e.g. in the BaseAlgorithm. User can also set the logging level prior to creating the algorithm with logging library, or by the single environment variable. I am not sure which should overwrite which one.

This would be significant divergence from how loggers are typically set up, to be honest. I also think it would be error prone and lead to confusing behaviour. 

Python logging works on a tree structure (w.r.t. modules) . When you configure the logging at a node in the tree, it then configures all of the loggers in the sub-nodes. This is why it is advisable to configure the logging at the root level of a library. 

Also, it precludes anything like the suggestion `User can also set the logging level prior to creating the algorithm with logging library, ` from working in practice.

Miffyli:
> Python logging works on a tree structure (w.r.t. modules) . When you configure the logging at a node in the tree, it then configures all of the loggers in the sub-nodes. This is why it is advisable to configure the logging at the root level of a library.

Ah, right, this is good to know! I have only used logging in rather small projects and never in bigger scale ^^. I will let @araffin provide the guidelines to actually make this work in a neat order.

araffin:
>Python logging works on a tree structure (w.r.t. modules) . When you configure the logging at a node in the tree, it then configures all of the loggers in the sub-nodes. This is why it is advisable to configure the logging at the root level of a library.

I didn't know that (I've also used python logging only for small projects too).

>As I mentioned before, I don't think there is a sensible way to introduce configurable standard python logging in a way that does not require these changes, i.e., that is backwards-compatible.

Looking back at what you did and the changes needed, I think you are right, this would lead to breaking changes.
I'm also thinking of per object verbosity (for instance, each callback has a different `verbose` parameter), so I'm thinking more and more that python logging won't help us there :/ (compared to what I thought originally).

@tfurmston thanks for the discussion and thanks for the PR =) but actually I'm afraid we would have to stick we our current `verbose` system for now.

However, I still plan to work (or welcome help) on our internal logger (add a method to allow the user to configure it): https://github.com/DLR-RM/stable-baselines3/issues/109

tfurmston:
> > Python logging works on a tree structure (w.r.t. modules) . When you configure the logging at a node in the tree, it then configures all of the loggers in the sub-nodes. This is why it is advisable to configure the logging at the root level of a library.
> 
> I didn't know that (I've also used python logging only for small projects too).
> 
> > As I mentioned before, I don't think there is a sensible way to introduce configurable standard python logging in a way that does not require these changes, i.e., that is backwards-compatible.
> 
> Looking back at what you did and the changes needed, I think you are right, this would lead to breaking changes.
> I'm also thinking of per object verbosity (for instance, each callback has a different `verbose` parameter), so I'm thinking more and more that python logging won't help us there :/ (compared to what I thought originally).
> 
> @tfurmston thanks for the discussion and thanks for the PR =) but actually I'm afraid we would have to stick we our current `verbose` system for now.
> 
> However, I still plan to work (or welcome help) on our internal logger (add a method to allow the user to configure it): #109

ok, no worries. 

Shall I close this PR then?

PR #419, opened 2021-05-05
Update PPO KL Divergence Estimator

### LINKED ISSUE #417: [Feature Request] Better KL Divergence Approximation
**Important Note: We do not do technical support, nor consulting** and don't answer personal questions per email.
Please post your question on the [RL Discord](https://discord.com/invite/xhfNqQv), [Reddit](https://www.reddit.com/r/reinforcementlearning/) or [Stack Overflow](https://stackoverflow.com/) in that case.

### 🚀 Feature

Use a better estimate for the KL Divergence in the PPO algorithm. 

The estimator I propose is the reverse estimator here:  
image

This has a lower variance since the extra term is negatively correlated and also is positive semi-definite (see motivation section).

### Motivation

In the PPO algorithm, there is a KL limit method used as a final block to large updates to the policy in a single timestep. The line used is: `approx_kl_divs.append(th.mean(rollout_data.old_log_prob - log_prob).detach().cpu().numpy())`. This is an unbiased estimator, but it has large variance since it can take on negative values (as opposed to the actual KL Divergence measure)! This can cause problems in the check: `if self.target_kl is not None and np.mean(approx_kl_divs) > 1.5 * self.target_kl:` since this doesn't consider the large negative values that `approx_kl_divs` could take. 

### Pitch

I want to replace the KL Divergence equation currently used in the PPO algorithm with the better approximation described above.

NOTE: This is basically a 1 line change.

### Alternatives

N/A

### Additional context

image

### Checklist

09tangriro:
## Description
Add the Schulman KL Divergence estimator and moved the KL check before the gradient update.

See the below image for examples comparing before (`old_approx_kl`) and after (`approx_kl`) in a little practical experiment.
image

For more robust comparison of values given, see the below graph:
image

I have also validated that the training does indeed stop when the KL Divergence is too large.

## Motivation and Context

closes #417 

## Types of changes

## Checklist:

### DIFF
--- a/docs/misc/changelog.rst
+++ b/docs/misc/changelog.rst
@@ -11,6 +11,7 @@ Breaking Changes:
 ^^^^^^^^^^^^^^^^^
 - Renamed ``_last_dones`` and ``dones`` to ``_last_episode_starts`` and ``episode_starts`` in ``RolloutBuffer``.
 - Updated the KL Divergence estimator in the PPO algorithm to be positive definite and have lower variance (@09tangriro)
+- Updated the KL Divergence check in the PPO algorithm to be before the gradient update step rather than after end of epoch (@09tangriro)
 
 New Features:
 ^^^^^^^^^^^^^
--- a/docs/misc/changelog.rst
+++ b/docs/misc/changelog.rst
@@ -10,6 +10,7 @@ Release 1.1.0a5 (WIP)
 Breaking Changes:
 ^^^^^^^^^^^^^^^^^
 - Renamed ``_last_dones`` and ``dones`` to ``_last_episode_starts`` and ``episode_starts`` in ``RolloutBuffer``.
+- Updated the KL Divergence estimator in the PPO algorithm to be positive definite and have lower variance (@09tangriro)
 
 New Features:
 ^^^^^^^^^^^^^
--- a/stable_baselines3/ppo/ppo.py
+++ b/stable_baselines3/ppo/ppo.py
@@ -231,13 +231,20 @@ class PPO(OnPolicyAlgorithm):
 
                 loss = policy_loss + self.ent_coef * entropy_loss + self.vf_coef * value_loss
 
+                # Calculate approximate form of reverse KL Divergence for early stopping
+                # see issue #417: https://github.com/DLR-RM/stable-baselines3/issues/417
+                # and Schulman blog: http://joschu.net/blog/kl-approx.html 
+                with th.no_grad():
+                    r = rollout_data.old_log_prob - log_prob
+                    approx_kl_div = th.mean((th.exp(r) - 1) - r).cpu().numpy()
+                    approx_kl_divs.append(approx_kl_div)
+
                 # Optimization step
                 self.policy.optimizer.zero_grad()
                 loss.backward()
                 # Clip grad norm
                 th.nn.utils.clip_grad_norm_(self.policy.parameters(), self.max_grad_norm)
                 self.policy.optimizer.step()
-                approx_kl_divs.append(th.mean(rollout_data.old_log_prob - log_prob).detach().cpu().numpy())
 
 
             if self.target_kl is not None and np.mean(approx_kl_divs) > 1.5 * self.target_kl:
--- a/stable_baselines3/ppo/ppo.py
+++ b/stable_baselines3/ppo/ppo.py
@@ -243,7 +243,8 @@ class PPO(OnPolicyAlgorithm):
 
                 if self.target_kl is not None and approx_kl_div > 1.5 * self.target_kl:
                     continue_training = False
-                    print(f"Early stopping at step {epoch} due to reaching max kl: {approx_kl_div:.2f}")
+                    if self.verbose >= 1:
+                        print(f"Early stopping at step {epoch} due to reaching max kl: {approx_kl_div:.2f}")
                     break
 
                 # Optimization step
--- a/stable_baselines3/ppo/ppo.py
+++ b/stable_baselines3/ppo/ppo.py
@@ -242,6 +242,7 @@ class PPO(OnPolicyAlgorithm):
                     approx_kl_divs.append(approx_kl_div)
 
                 if self.target_kl is not None and approx_kl_div > 1.5 * self.target_kl:
+                    continue_training = False
                     print(f"Early stopping at step {epoch} due to reaching max kl: {approx_kl_div:.2f}")
                     break
 
--- a/stable_baselines3/ppo/ppo.py
+++ b/stable_baselines3/ppo/ppo.py
@@ -172,6 +172,8 @@ class PPO(OnPolicyAlgorithm):
         pg_losses, value_losses = [], []
         clip_fractions = []
 
+        continue_training = True
+
         # train for n_epochs epochs
         for epoch in range(self.n_epochs):
             approx_kl_divs = []
@@ -239,6 +241,10 @@ class PPO(OnPolicyAlgorithm):
                     approx_kl_div = th.mean((th.exp(r) - 1) - r).cpu().numpy()
                     approx_kl_divs.append(approx_kl_div)
 
+                if self.target_kl is not None and approx_kl_div > 1.5 * self.target_kl:
+                    print(f"Early stopping at step {epoch} due to reaching max kl: {approx_kl_div:.2f}")
+                    break
+
                 # Optimization step
                 self.policy.optimizer.zero_grad()
                 loss.backward()
@@ -246,9 +252,7 @@ class PPO(OnPolicyAlgorithm):
                 th.nn.utils.clip_grad_norm_(self.policy.parameters(), self.max_grad_norm)
                 self.policy.optimizer.step()
 
-
-            if self.target_kl is not None and np.mean(approx_kl_divs) > 1.5 * self.target_kl:
-                print(f"Early stopping at step {epoch} due to reaching max kl: {np.mean(approx_kl_divs):.2f}")
+            if not continue_training:
                 break
 
         self._n_updates += self.n_epochs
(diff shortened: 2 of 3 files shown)


stable_baselines3/ppo/ppo.py
```diff
@@ -231,18 +233,27 @@ def train(self) -> None:
...
+                # see issue #417: https://github.com/DLR-RM/stable-baselines3/issues/417
+                # and Schulman blog: http://joschu.net/blog/kl-approx.html 
+                with th.no_grad():
+                    r = rollout_data.old_log_prob - log_prob
+                    approx_kl_div = th.mean((th.exp(r) - 1) - r).cpu().numpy()
+                    approx_kl_divs.append(approx_kl_div)
+
+                if self.target_kl is not None and approx_kl_div > 1.5 * self.target_kl:
+                    continue_training = False
+                    print(f"Early stopping at step {epoch} due to reaching max kl: {approx_kl_div:.2f}")
```
Miffyli, line 246: Off-PR, but this should be inside `if self.verbose >= 1:` (a debug-print-ish thing)

docs/misc/changelog.rst
```diff
@@ -10,6 +10,7 @@ Release 1.1.0a5 (WIP)
 Breaking Changes:
 ^^^^^^^^^^^^^^^^^
 - Renamed ``_last_dones`` and ``dones`` to ``_last_episode_starts`` and ``episode_starts`` in ``RolloutBuffer``.
+- Updated the KL Divergence estimator in the PPO algorithm to be positive definite and have lower variance (@09tangriro)
```
Miffyli, line 14: Add update on the changed behaviour of `target_kl` (when the termination is checked)

Miffyli:
Thanks for the fast PR! Left some comments. Could you possibly also show the difference between new and old KL over CartPole training session (you can also upload the training logs of the printed out stats, and I can make the plots)? :)

09tangriro:
Hey, sure I can do the plots as well tbh :)

I added the other things to commits, but before I push I want to sort out the commit checks. Sorry if this is a bit of nuisance, but for some reason when I run 'make check-codestyle' it suggests reformatting 22 files rather than just the 1 in the Git CI pipeline:
image
I don't suppose you know why this is? For now, I've just run the same command over the ppo file directly but just in case it something goes wrong hopefully this goes some way to explain it 😅

09tangriro:
These are the graphs over 10k timesteps of cartpole, I've plotted the old and new KL divergence estimate over each update as well as over each epoch:
image
image

stable_baselines3/ppo/ppo.py
```diff
@@ -231,18 +233,28 @@ def train(self) -> None:
 
                 loss = policy_loss + self.ent_coef * entropy_loss + self.vf_coef * value_loss
 
+                # Calculate approximate form of reverse KL Divergence for early stopping
+                # see issue #417: https://github.com/DLR-RM/stable-baselines3/issues/417
+                # and Schulman blog: http://joschu.net/blog/kl-approx.html
+                with th.no_grad():
+                    r = rollout_data.old_log_prob - log_prob
```
araffin, line 240: maybe a better variable name would be `prob_ratio` or just `ratio` (but this name is used before)

09tangriro: Sure, I think `log_ratio` might be more appropriate?

araffin: yep =)

araffin:
> These are the graphs over 10k timesteps of cartpole, I've plotted the old and new KL divergence estimate over each update as well as over each epoch:

thanks for doing the experiment =)
I assume you use the same random seed ?

araffin:
>I added the other things to commits, but I feel I should bring up a problem I'm having with the commit checks. Sorry if this is a bit of nuisance, but for some reason when I run 'make check-codestyle' it suggests reformatting 22 files rather than just the 1 in the Git CI pipeline:

Make sure to have latest black and isort versions...

09tangriro:
> 
> 
> > These are the graphs over 10k timesteps of cartpole, I've plotted the old and new KL divergence estimate over each update as well as over each epoch:
> 
> thanks for doing the experiment =)
> I assume you use the same random seed ?

No worries :)
And indeed I did

09tangriro:
Sorry, let me try fix my make commands again 😅

stable_baselines3/ppo/ppo.py
Miffyli: Should this not also be the other way around `log_ratio = log_prob - rollout_data.old_log_prob`? In the material `r = log p(x)/q(x)`, where data is sampled from `q(x)`, which here corresponds to the old log probs.

09tangriro: I don't think so, let me elaborate below:

`p(x)` represents a 'true' distribution and `q(x)` is the distribution estimating `p(x)`. You're right, we sample data from `q(x)`, in this case that means sampled from the neural network with respect to the network parameters, θ. That is, `q(x) = current_dist`, which is estimating the old policy such that `p(x) = old_dist`

araffin:
The build is failing but I'm not sure why, I tried to reproduce pytype error locally but couldn't ...

stable_baselines3/ppo/ppo.py
Miffyli: Hmm I am having hard time wrapping my head around that. If we sample from `q`, it should then be `q(x) = old_dist`, because it was that `old_dist` we sampled the experiences from. Also looking at the replaced KL estimate, which is `log q(x)/p(x)` (or `rollout_data.old_log_prob - log_prob`), seems to indicate we should treat `old_log_prob` as `log q(x)`.

09tangriro: From the article I linked in the issue on Forward and Reverse KL: 
image

In this case, I'd argue we control the current policy through the network parameters. I think the 'true distribution' is the old policy since this is the fixed benchmark we want to measure the current policy against.

09tangriro:
> 
> 
> The build is failing but I'm not sure why, I tried to reproduce pytype error locally but couldn't ...

Yeah it's certainly suspicious that a single variable name change would break everything, I'll fiddle around in the coming days :)

stable_baselines3/ppo/ppo.py
09tangriro: Another little blurb:
image

09tangriro: Although I see your point, the neural network is used to predict both the old and new policy, so changing the network parameters would in fact change the old policy as well, but we have to take one of these as fixed, and I think it makes more sense to take the old policy for this role.

09tangriro:
Going back to working build briefly to try again :)

09tangriro:
Well this is strange the same code passed the build last time 🤔

stable_baselines3/ppo/ppo.py
Miffyli: Gah I am still having trouble digesting all this (been a long time since I had to play with KLs in this way) and still not convinced with the setup. My main problem is with how Schulman's blog sets `x ~ q(x)`, which in our case is the old policy (which is also fixed with stored `log_prob`), but the code now treats as samples are from `x ~ p(x)`, changing things around.

Could you do a quick comparison what happens in the CartPole KL-plots if you flip the ratio around? Maybe this could shed bit of light in to the matter.

@araffin can you throw a comment here?

09tangriro: Yes actually I do see your point, the way sampling is being described certainly complicates things.

However, I think a good thought experiment is to zoom out briefly and discuss the KL divergence without the approximation, since this setting removes the idea of samples altogether 😁:

`KL(x) = E[log(q(x) / p(x))]` with the expectation over samples from q, so `KL(x) = sum/integral_x q(x) * q(x)/p(x)`. 

I'd argue the way to calculate this in this setting would be to get all the possible actions, `x`, and put them in that equation where `q(x)` is the current policy and `p(x)` is the old policy, again since we control the current policy. If this is the case for the full form equation, I'd propose it should also be the case for the approximation.

Still, I tried reversing the ratio to see what happens but as expected the ratio has the same range, and you get a similar looking graph:
image

Miffyli: Thanks for being so patient with slow-me! I do agree with your explanation and that old policy should be `p(x)`, so lets go with this current code you have. The empirical results are still desirable, with a major change in value scale. While I do not normally agree with "results look good enough" approach, I think we need to call this done for now before we get too stuck into this (and I do not have the skill-set to fully argue against this).

Lets leave this discussion open ("not resolved") so anybody wondering this change can see the discussion :).

09tangriro: Not at all to be honest I'm less sure now myself 😅, I agree with the way sampling is phrased it does appear that it should be the other way round; however, I still can't shake the feeling that it breaks intuition (for me at least). 

I agree, let's leave it open. Hopefully Schulman one day stumbles onto this and maybe sheds some light on the issue lool.

Miffyli: I'll the brave move and poke @joschu if he would be so kind to help us lost souls on this matter ^^

09tangriro:
See #424, same code but build passes somehow :)

araffin:
Build passes again now =) (I restarted it)

stable_baselines3/ppo/ppo.py
joschu: Glad to see the blog post is useful! Using the blog's notation, I'd let q be the old distribution (=the sampling distribution). Then we can approximate KL[new,old] or KL[old,new] by using the two expressions I listed.
I assume you guys know this, so maybe I'm missing something.

09tangriro: Ah ok, I thought it was the other way round (q = new/current distribution) because I think it fits the intuition of KL Divergence better (we control the current distribution and the old distribution is seen as fixed). 

Just one more question if you'd be willing, in the case of the full KL Divergence (i.e. away from sampling since we sum/integrate over all actions), would you still say q = old distribution? If so, why?

Miffyli (approved):
LGTM! Big thanks for being patient with me and the CI system! :)

PR #431, opened 2021-05-13
KL Divergence Helper Function

### LINKED ISSUE #430: [Feature Request] Full KL Divergence
### 🚀 Feature

Add a full form KL Divergence function that is compatible with the `Distribution` wrappers and can easily be called.

### Motivation

When implementing the PPG algorithm in the Contrib library, it would be useful to have a full form KL Divergence function in the `distributions.py` file which is compatible with the `Distribution` wrappers around the PyTorch distributions. OpenAI also use the full form equation for KL Divergence in their PPG implementation.

### Pitch

I want to add a `kl_divergence(dist1: Distribution, dist2: Distribution) -> th.Tensor` method to the `distributions.py` file which would essentially be a wrapper around the PyTorch `kl_divergence` function. As well as PPG, I think this would be a useful feature in general.

### Alternatives

N/A

### Additional context

N/A

### Checklist

09tangriro:
## Description
Add a KL Divergence wrapper function for `Distribution` classes.
Closes #430 

## Motivation and Context

## Types of changes

## Checklist:


tests/test_distributions.py
```diff
@@ -111,3 +112,23 @@ def test_categorical(dist, CAT_ACTIONS):
...
+        StateDependentNoiseDistribution(N_ACTIONS).proba_distribution(
+            th.rand(N_ACTIONS), th.rand([N_ACTIONS, N_ACTIONS]), th.rand([N_ACTIONS, N_ACTIONS])
+        ),
+    ],
+)
+def test_kl_divergence(dist_type):
+    dist1 = dist_type
+    dist2 = dist_type
+
+    assert kl_divergence(dist1, dist2) == 0
```
araffin, line 134: I would also add a check that compare the kl div here with the one you can compute by sampling (see what is done for the entropy)

stable_baselines3/common/distributions.py
```diff
@@ -325,23 +323,23 @@ def proba_distribution_net(self, latent_dim: int) -> nn.Module:
         return action_logits
 
     def proba_distribution(self, action_logits: th.Tensor) -> "MultiCategoricalDistribution":
-        self.distributions = [Categorical(logits=split) for split in th.split(action_logits, tuple(self.action_dims), dim=1)]
```
araffin, line 328: why do you need the rename here? (as you are using a different implementation below)

stable_baselines3/common/distributions.py
```diff
@@ -676,3 +672,50 @@ def make_proba_distribution(
...
+    :param dist_true: the p distribution
+    :param dist_pred: the q distribution
+    :return: KL(dist_true||dist_pred)
+    """
+    # KL Divergence for different distribution types is out of scope
+    assert dist_true.__class__ == dist_pred.__class__, "Error: input distributions should be the same type"
+
+    # MultiCategoricalDistribution is not a PyTorch Distribution subclass
+    # so we need to implement it ourselves!
+    if isinstance(dist_pred, MultiCategoricalDistribution):
```
araffin, line 690: do you have a reference for this implementation or are you just using the definition?

docs/misc/changelog.rst
```diff
@@ -40,9 +40,10 @@ New Features:
   to handle gym3-style vectorized environments (@vwxyzjn)
 - Ignored the terminal observation if the it is not provided by the environment
   such as the gym3-style vectorized environments. (@vwxyzjn)
-- Add policy_base as input to the OnPolicyAlgorithm for more flexibility (@09tangriro)
+- Added policy_base as input to the OnPolicyAlgorithm for more flexibility (@09tangriro)
 - Added support for image observation when using ``HER``
 - Added ``replay_buffer_class`` and ``replay_buffer_kwargs`` arguments to off-policy algorithms
+- Added ``kl_divergence`` wrapper for ``Distribution`` classes (@09tangriro)
```
araffin, line 46:
```suggestion
- Added ``kl_divergence`` helper for ``Distribution`` classes (@09tangriro)
```

tests/test_distributions.py
araffin: I would also do some sanity checks when we know in advance the value of the kl div (for which the kl is not zero, for instance comparing two 1D Gaussian dist)

09tangriro: sure I can add these :)

stable_baselines3/common/distributions.py
09tangriro: Mainly for conformity, there's no real upside to having it be named differently than all the other distributions I'd argue it just makes things confusing when using it, especially since `self.distribution` would also be callable since it's in the base class now.

stable_baselines3/common/distributions.py
09tangriro: I'm just using the definition here, just a simple sum over the action space :)

araffin (commented):
Apart from the additional tests and from the MultiCategorical I need to double check, already look good to me, thanks for the PR =)

09tangriro:
Is there a way to temporarily upgrade the PyTorch version to include `th.tile()` for the kl divergence tests? Trying to replicate the behaviour in Python native results in huge run times. I haven't just replicated the entropy test because I'm not convinced it's been done correctly. As I understand it, you're testing that `E(unbiased_estimator) = entropy` via Monte Carlo sampling. I would have thought to do this you would:
1. Get the actual entropy of a **single** distribution
2.  Sample `N_SAMPLES` actions from that distribution
3. Calculate the negative log prob of the actions via the distribution
4. Calculate the mean of the negative log probs and compare to the actual entropy calculated in step 1

As far as I can tell, what actually happens is:
1. `N_SAMPLES` **different** distributions are created
2. The entropy of all the different distributions are calculated
3. An action is sampled from each distribution and it's negative log prob calculated according to the distribution it was sampled from
4. Average the entropies and log probs and compare

I think I'm missing or have misunderstood because clearly the test passes, I'm just not sure why. Any help on this would be massively appreciated, I'm curious 😊

Miffyli:
> Is there a way to temporarily upgrade the PyTorch version to include th.tile() for the kl divergence tests? Trying to replicate the behaviour in Python native results in huge run times.

A quick passing-by comment: you should be able to get same results with `.repeat` function of tensors, e.g:
```python
x = th.tile(th.rand(N_ACTIONS), (N_SAMPLES, 1))
# Becomes
x = th.rand(N_ACTIONS).repeat(N_SAMPLES, 1)
```

09tangriro:
> 
> 
> > Is there a way to temporarily upgrade the PyTorch version to include th.tile() for the kl divergence tests? Trying to replicate the behaviour in Python native results in huge run times.
> 
> A quick passing-by comment: you should be able to get same results with `.repeat` function of tensors, e.g:
> 
> ```python
> x = th.tile(th.rand(N_ACTIONS), (N_SAMPLES, 1))
> # Becomes
> x = th.rand(N_ACTIONS).repeat(N_SAMPLES, 1)
> ```

Ah lifesaver! Thanks :)

araffin:
Taking a look at SB2 code, I think the `MultiCategorical` kl div can be simplified to:
```python
def kl(dist_true, dist_pred):
    return th.stack(
        [th.distributions.kl_divergence(p, q) for p, q in zip(dist_true.distribution, dist_pred.distribution)],
        dim=1,
    ).sum(dim=1)
```
(I quickly tested it and it yields the same result as your implementation)

From https://github.com/hill-a/stable-baselines/blob/master/stable_baselines/common/distributions.py#L362

and the testing (entropy and kl) was done here: https://github.com/hill-a/stable-baselines/blob/master/tests/test_distri.py

09tangriro:
> 
> 
> Taking a look at SB2 code, I think the `MultiCategorical` kl div can be simplified to:
> 
> ```python
> def kl(dist_true, dist_pred):
>     return th.stack(
>         [th.distributions.kl_divergence(p, q) for p, q in zip(dist_true.distribution, dist_pred.distribution)],
>         dim=1,
>     ).sum(dim=1)
> ```
> 
> (I quickly tested it and it yields the same result as your implementation)
> 
> From https://github.com/hill-a/stable-baselines/blob/master/stable_baselines/common/distributions.py#L362
> 
> and the testing (entropy and kl) was done here: https://github.com/hill-a/stable-baselines/blob/master/tests/test_distri.py

Ah yeah I had a suspicion that the additive property held for KL Divergence, but it was taking me a while to derive the proof, should've just done a quick google search 😅: https://statproofbook.github.io/P/kl-add.html

stable_baselines3/common/distributions.py
```diff
@@ -676,3 +672,29 @@ def make_proba_distribution(
...
+        assert dist_pred.action_dims == dist_true.action_dims, "Error: distributions must have the same input space"
+        return th.stack(
+            [th.distributions.kl_divergence(p, q) for p, q in zip(dist_true.distribution, dist_pred.distribution)],
+            dim=1,
+        ).sum(dim=1)
+
+    # Use the PyTorch kl_divergence implementation
+    else:
+        # KL Divergence should not be a vector
```
araffin, line 699: why this comment?

09tangriro: Sorry you're right, that's outdated, I'll remove it :)

tests/test_distributions.py
```diff
@@ -111,3 +114,76 @@ def test_categorical(dist, CAT_ACTIONS):
...
+        ),
+    ],
+)
+def test_kl_divergence(dist_type):
+    # Test 1: same distribution should have KL Div = 0
+    dist1 = dist_type
+    dist2 = dist_type
+    # PyTorch implementation of kl_divergence doesn't sum across dimensions
+    # so we need to check each one
+    assert th.all(kl_divergence(dist1, dist2).bool()) == False
```
araffin, line 138: it's a bit confusing... why not:
```suggestion
    assert th.allclose(kl_divergence(dist1, dist2).sum(), th.tensor(0.0))
```
?

tests/test_distributions.py
```diff
@@ -111,3 +114,76 @@ def test_categorical(dist, CAT_ACTIONS):
...
+        mean_actions2 = th.rand(1).repeat(N_SAMPLES, 1)
+        _, log_std = dist1.proba_distribution_net(N_FEATURES, log_std_init=th.log(th.tensor(0.2)))
+        dist1.sample_weights(log_std, batch_size=N_SAMPLES)
+        dist2.sample_weights(log_std, batch_size=N_SAMPLES)
+        dist1 = dist1.proba_distribution(mean_actions1, log_std, state)
+        dist2 = dist2.proba_distribution(mean_actions2, log_std, state)
+
+    full_kl_div = kl_divergence(dist1, dist2).mean(dim=0)
+    actions = dist1.get_actions()
+    approx_kl_div = (dist1.log_prob(actions) - dist2.log_prob(actions)).mean(dim=0)
```
araffin, line 172: why not using the estimator from https://github.com/DLR-RM/stable-baselines3/pull/419 ?

tests/test_distributions.py
```diff
@@ -111,3 +114,76 @@ def test_categorical(dist, CAT_ACTIONS):
...
+    # so we need to check each one
+    assert th.all(kl_divergence(dist1, dist2).bool()) == False
+
+    # Test 2: KL Div = E(Unbiased approx KL Div)
+    if isinstance(dist_type, CategoricalDistribution):
+        dist1 = dist_type.proba_distribution(th.rand(N_ACTIONS).repeat(N_SAMPLES, 1))
+        # deepcopy needed to assign new memory to new distribution instance
+        dist2 = deepcopy(dist_type).proba_distribution(th.rand(N_ACTIONS).repeat(N_SAMPLES, 1))
+    elif isinstance(dist_type, DiagGaussianDistribution) or isinstance(dist_type, SquashedDiagGaussianDistribution):
+        mean_actions1 = th.rand(1).repeat(N_SAMPLES, 1)
```
araffin, line 146: why only 1D Gaussian?

tests/test_distributions.py
09tangriro: Sure!

tests/test_distributions.py
09tangriro: There's not much point in using the more complicated estimator since in this case we just want to assert the expected value of an unbiased estimator is equal to the original metric. For this, the variance of the metric doesn't matter so we may as well use the simpler form, using the other estimator would yield the same result just it would take longer to do the extra calculations -> no upside.

tests/test_distributions.py
09tangriro: Mainly for simplicity, the tests for everything other than `MultiCategoricalDistribution` are a little redundant since they coincide with the PyTorch distribution tests; I was viewing these tests as more a form of documentation.

araffin:
> Is there a way to temporarily upgrade the PyTorch version to include `th.tile()` for the kl divergence tests? Trying to replicate the behaviour in Python native results in huge run times. I haven't just replicated the entropy test because I'm not convinced it's been done correctly. As I understand it, you're testing that `E(unbiased_estimator) = entropy` via Monte Carlo sampling. I would have thought to do this you would:
> 
>     1. Get the actual entropy of a **single** distribution
> 
>     2. Sample `N_SAMPLES` actions from that distribution
> 
>     3. Calculate the negative log prob of the actions via the distribution
> 
>     4. Calculate the mean of the negative log probs and compare to the actual entropy calculated in step 1
> 
> 
> As far as I can tell, what actually happens is:
> 
>     1. `N_SAMPLES` **different** distributions are created
> 
>     2. The entropy of all the different distributions are calculated
> 
>     3. An action is sampled from each distribution and it's negative log prob calculated according to the distribution it was sampled from
> 
>     4. Average the entropies and log probs and compare
> 
> 
> I think I'm missing or have misunderstood because clearly the test passes, I'm just not sure why. Any help on this would be massively appreciated, I'm curious blush

>1. `N_SAMPLES` **different** distributions are created

what makes you think that?
the `state = th.rand(N_SAMPLES, N_FEATURES)` for the state dependent noise?
yes, I think you are right. I updated the test, does it look better now?

09tangriro:
> 
> 
> > Is there a way to temporarily upgrade the PyTorch version to include `th.tile()` for the kl divergence tests? Trying to replicate the behaviour in Python native results in huge run times. I haven't just replicated the entropy test because I'm not convinced it's been done correctly. As I understand it, you're testing that `E(unbiased_estimator) = entropy` via Monte Carlo sampling. I would have thought to do this you would:
> > ```
> > 1. Get the actual entropy of a **single** distribution
> > 
> > 2. Sample `N_SAMPLES` actions from that distribution
> > 
> > 3. Calculate the negative log prob of the actions via the distribution
> > 
> > 4. Calculate the mean of the negative log probs and compare to the actual entropy calculated in step 1
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > As far as I can tell, what actually happens is:
> > ```
> > 1. `N_SAMPLES` **different** distributions are created
> > 
> > 2. The entropy of all the different distributions are calculated
> > 
> > 3. An action is sampled from each distribution and it's negative log prob calculated according to the distribution it was sampled from
> > 
> > 4. Average the entropies and log probs and compare
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think I'm missing or have misunderstood because clearly the test passes, I'm just not sure why. Any help on this would be massively appreciated, I'm curious blush
> 
> > 1. `N_SAMPLES` **different** distributions are created
> 
> what makes you think that?
> the `state = th.rand(N_SAMPLES, N_FEATURES)` for the state dependent noise?
> yes, I think you are right. I updated the test, does it look better now?

Partly, but also with `deterministic_actions = th.rand(N_SAMPLES, N_ACTIONS)` which goes on to be the means of the distributions. I think this should be something like `th.rand(N_ACTIONS).repeat(N_SAMPLES, 1)`.

araffin:
>Partly, but also with deterministic_actions = th.rand(N_SAMPLES, N_ACTIONS) which goes on to be the means of the distributions. I think this should be something like th.rand(N_ACTIONS).repeat(N_SAMPLES, 1).

I see, well the entropy of a Gaussian only depends on its variance, but I get your point. I will change that.

09tangriro:
> 
> 
> > Partly, but also with deterministic_actions = th.rand(N_SAMPLES, N_ACTIONS) which goes on to be the means of the distributions. I think this should be something like th.rand(N_ACTIONS).repeat(N_SAMPLES, 1).
> 
> I see, well the entropy of a Gaussian only depends on its variance, but I get your point. I will change that.

Ah yeah that makes sense actually, nice though 🙌

PR #469, opened 2021-06-07
Fix logger setup

### LINKED ISSUE #109: Allow for logger setup before calling BaseAlgorithm.learn()

### LINKED ISSUE #367: [Bug] Logger verbosity change between each call
### 🐛 Bug

After calling a model with `verbose=0`, the logger does not print anything, even though `verbose=1` is passed.
Probably should be addressed with #109 .

### To Reproduce

```python
from stable_baselines3 import PPO, A2C

model = PPO("MlpPolicy", "CartPole-v1", verbose=0).learn(3000)
model = PPO("MlpPolicy", "CartPole-v1", verbose=1).learn(3000)


```

araffin:
## Motivation and Context

closes #109
closes #367

## Types of changes

## Checklist:


stable_baselines3/common/base_class.py
```diff
@@ -228,6 +233,17 @@ def _wrap_env(env: GymEnv, verbose: int = 0, monitor_wrapper: bool = True) -> Ve
     def _setup_model(self) -> None:
         """Create networks, buffer and optimizers."""
 
+    def set_logger(self, logger: Logger) -> None:
+        """Setter for for logger object."""
```
AdamGleave, line 237: Maybe we should document that this will override the settings `tensorboard_log` and (many of the effects of) `verbose` set in `__init__`, what do you think?

stable_baselines3/common/logger.py
```diff
@@ -689,11 +690,14 @@ def configure(folder: Optional[str] = None, format_strings: Optional[List[str]]
     if format_strings is None:
         format_strings = os.getenv("SB3_LOG_FORMAT", "stdout,log,csv").split(",")
 
-    format_strings = filter(None, format_strings)
+    format_strings = list(filter(None, format_strings))
     output_formats = [make_output_format(f, folder, log_suffix) for f in format_strings]
 
     Logger.CURRENT = Logger(folder=folder, output_formats=output_formats)
```
AdamGleave, line 696: Do we still need Logger.CURRENT? It seems confusing to me that we both return a Logger instance but calling `configure` still has side-effects. (Perhaps we can keep Logger.CURRENT but have a deprecation warning when it's used if needed for backward compatibility?)

stable_baselines3/common/utils.py
```diff
@@ -172,26 +172,29 @@ def configure_logger(
...
     tb_log_name: str = "",
     reset_num_timesteps: bool = True,
-) -> None:
+) -> Logger:
     """
     Configure the logger's outputs.
 
     :param verbose: the verbosity level: 0 no output, 1 info, 2 debug
     :param tensorboard_log: the log location for tensorboard (if None, no logging)
     :param tb_log_name: tensorboard log
```
AdamGleave, line 181: Not new but this seems to be missing documentation for the reset_num_timesteps parameter?

(I like [darglint](https://github.com/terrencepreilly/darglint) for catching these errors.)

stable_baselines3/common/vec_env/vec_video_recorder.py
```diff
@@ -93,7 +92,7 @@ def step_wait(self) -> VecEnvStepReturn:
             self.video_recorder.capture_frame()
             self.recorded_frames += 1
             if self.recorded_frames > self.video_length:
-                logger.info("Saving video to ", self.video_recorder.path)
+                print(f"Saving video to {self.video_recorder.path}")
```
AdamGleave, line 95: Hm, `logger` did seem more natural here, I guess we're doing this to try and avoid using the global logger? Perhaps we can make the `common.logger` module only used for key-value logging (like recording loss over time), but still have some global logger for the Stable Baselines code. Or even just logger = logging.getLogger(__name__) in each module.

tests/test_logger.py
```diff
@@ -103,6 +108,48 @@ def read_fn(_format):
...
+    assert model.logger.dir == new_tmp_path
+
+    # Default outputs with verbose=1
+    model = A2C("MlpPolicy", "CartPole-v1", verbose=1).learn(4)
+    assert isinstance(model.logger.output_formats[0], HumanOutputFormat)
+    # with tensorboard
+    model = A2C("MlpPolicy", "CartPole-v1", verbose=1, tensorboard_log=str(tmp_path)).learn(4)
+    assert isinstance(model.logger.output_formats[0], HumanOutputFormat)
+    assert isinstance(model.logger.output_formats[1], TensorBoardOutputFormat)
+    model.learn(32)
```
AdamGleave, line 135: Consider adding `assert len(model.logger.output_formats) == 2`?

tests/test_logger.py
```diff
@@ -103,6 +108,48 @@ def read_fn(_format):
...
+
+    model = A2C("MlpPolicy", "CartPole-v1", verbose=1)
+    model.set_logger(new_logger)
+    model.learn(32)
+    # Check that the new logger is not overwritten
+    assert isinstance(model.logger.output_formats[0], HumanOutputFormat)
+    assert isinstance(model.logger.output_formats[1], CSVOutputFormat)
+    assert isinstance(model.logger.output_formats[2], TensorBoardOutputFormat)
+
```
AdamGleave, line 151: Consider adding `assert len(model.logger.output_formats) == 3`?

tests/test_logger.py
```diff
@@ -103,6 +108,48 @@ def read_fn(_format):
...
+    model = A2C("MlpPolicy", "CartPole-v1", verbose=1, tensorboard_log=str(tmp_path)).learn(4)
+    assert isinstance(model.logger.output_formats[0], HumanOutputFormat)
+    assert isinstance(model.logger.output_formats[1], TensorBoardOutputFormat)
+    model.learn(32)
+    # set new logger
+    model.set_logger(new_logger)
+    # Check that the new logger is correctly setup
+    assert isinstance(model.logger.output_formats[0], HumanOutputFormat)
+    assert isinstance(model.logger.output_formats[1], CSVOutputFormat)
+    assert isinstance(model.logger.output_formats[2], TensorBoardOutputFormat)
```
AdamGleave, line 141: Consider adding assert len(model.logger.output_formats) == 3?

tests/test_logger.py
```diff
@@ -103,6 +108,48 @@ def read_fn(_format):
...
+    model.learn(32)
```
AdamGleave, line 142: What does this line do? We overwrite `model` in the next line. Is this just checking that training happens (including logging) without raising an exception?

AdamGleave (commented):
Thanks for implementing this, seems like a significant improvement!

Right now I'm a bit confused what the intended use of `common.logger` is. It seems mostly (entirely?) used with local instances of the Logger object. But we still have the global `Logger.CURRENT` around associated methods `logger.{log,info,warn}` around.

I'm wondering if we should get rid of the global entirely, replacing it e.g. with `logger = logging.getLogger(__name__)` in modules that need logging that is not tied to a specific algorithm. Or have a separate global and algorithm-specific logger, with the global one perhaps not implementing the key-value methods like `record` and `dump` (which I think we always want tied to an algorithm)?

Interested to hear your thoughts on this, not sure what the best architecture is here, this is definitely an improvement on the status quo.

tests/test_logger.py
araffin: > Is this just checking that training happens (including logging) without raising an exception?

Exactly (because we changed the logger, it is just a "running" test)

stable_baselines3/common/utils.py
araffin: nice catch

stable_baselines3/common/logger.py
araffin: Good point, I removed it as it was not use anymore.

stable_baselines3/common/base_class.py
araffin: Very good point. Added a warning

stable_baselines3/common/vec_env/vec_video_recorder.py
araffin: I'm a bit puzzled with this one as it is the only place where logger is called from outside an algorithm.

>Or even just logger = logging.getLogger(name) in each module.

this seems a bit overkill compared to current usage, no?

araffin:
Thanks for the review =) (very good remarks)

>Right now I'm a bit confused what the intended use of common.logger is. It seems mostly (entirely?) used with local instances of the Logger object. But we still have the global Logger.CURRENT around associated methods logger.{log,info,warn} around.

the `Logger` object is solely used (there was one exception in the vec video recorder) in the algorithms to log stats about training.
I removed the global one to avoid confusion (I had issues in the past because of this global logger).

>I'm wondering if we should get rid of the global entirely, replacing it e.g. with logger = logging.getLogger(__name__) in modules that need logging that is not tied to a specific algorithm. Or have a separate global and algorithm-specific logger, with the global one perhaps not implementing the key-value methods like record and dump (which I think we always want tied to an algorithm)?

As the usage is limited to algorithms, I would just keep the local one.
Regarding a logger per module, we had an issue and PR for that (https://github.com/DLR-RM/stable-baselines3/pull/407) but it turned out to be a bit overkill for our usage (where just using `verbose` argument works quite well).

stable_baselines3/common/vec_env/vec_video_recorder.py
AdamGleave: Yeah it is a bit overkill for a single usage. Nice since it allows users of the library to e.g. choose what log level they want for Stable Baselines as a whole. Given this is a one-off I think I'm fine with sticking to print, or just `import logging` and `logging.info`

AdamGleave (approved):
LGTM. I think I still slightly prefer using Python global logger (rather than print) for non-algorithm-specific logging for flexibility (e.g. stdout is not always saved), but it's a pretty minor preference, and probably belongs in another PR anyway given there are existing uses of `print` in the codebase.

PR #507, opened 2021-07-07
Show how to log the dump for N steps

### LINKED ISSUE #506: Not getting any TensorBoard logs for DQN model
Hi Guys,
 
I am having a problem with the DQN logging to TensorBoard, I don't get when the log is getting called. At the moment, to me it seems the log is only made when the "done" parameter returns True from step(), and this is only after the second iteration, done needs to return true twice?!

My code doesn't return a done = True unless it completely fails. 

The code works fine in PPO method without the done returning true, and logs my values just fine at each iteration.

My data CSV is around 160k steps long. Am I missing something? Here's my model call. I also get the max score at every step when I print here, just my tensorboard is not working.... How do we get it working like PPO? I.E on every iteration after X steps? Thanks!

```
class TensorboardCallback(BaseCallback):
    Custom callback for plotting additional values in tensorboard.
 
    def __init__(self, verbose=0):
        super(TensorboardCallback, self).__init__(verbose)

    def _on_step(self) -> bool:
        # Log scalar value (here a random variable)
        # print(self.locals["rewards"])
        # print(self.locals)
        global top_score
        score = self.locals["infos"][0]["score"]
        self.logger.record('a_score', score)
        if (top_score == None):
            top_score = score
            model.save("./model_DQN")
        elif (score > top_score):
            top_score = score
            model.save("./model_DQN")
        return True

model = DQN("MlpPolicy", env, verbose=1, tensorboard_log="./a2c_cartpole_tensorboard/", device="cpu")
model.learn(total_timesteps=int(1e10), callback=TensorboardCallback())
```

saulable:
closes #506

## Motivation and Context

## Types of changes

## Checklist:

### DIFF
--- a/docs/guide/tensorboard.rst
+++ b/docs/guide/tensorboard.rst
@@ -270,6 +270,8 @@ Dumping DQN stats for TensorBoard
 
 If you would like to log the reward for DQN every N steps, follow the following formula. You will need to check the self.num_timesteps val and change the 10000 for how many steps you would like an update for. This is needed if "done" in step() does not return True when you are training the model. 
 
+If you want to track the reward, you will pass this from "info" as a dict in the step return. 
+
 .. code-block:: python
     class TensorboardCallback(BaseCallback):
         """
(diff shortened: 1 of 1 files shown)


Miffyli:
Hmm this documentation seems very specific to DQN at the current form. Looking at the original issue #506 , I think this PR should add an example on how to dump stats more often than at the end of the episodes (as instructed in the issue).

@araffin  comments?

araffin:
i would rather add a note and link to the two relevant issues.
The current code is indeed too specific as it only works for dqn with one env.

saulable:
Yeah, it's not correct sorry. I ran this code and discovered issues with
converging. It looks like it needs the done return after all. Complicated
stuff :)

PR #537, opened 2021-08-12
Training and evaluation: call model.train() and model.eval() 

davidblom603:
…e and disable dropout and batchnorm

Fixes #535

## Motivation and Context

- [x] I have raised an issue to propose this change ([required](https://github.com/DLR-RM/stable-baselines3/blob/master/CONTRIBUTING.md) for new features and bug fixes)

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation (update in the documentation)

## Checklist:

- [x] I've read the [CONTRIBUTION](https://github.com/DLR-RM/stable-baselines3/blob/master/CONTRIBUTING.md) guide (**required**)
- [x] I have updated the changelog accordingly (**required**).
- [ ] My change requires a change to the documentation.
- [x] I have updated the tests accordingly (*required for a bug fix or a new feature*).
- [ ] I have updated the documentation accordingly.
- [x] I have reformatted the code using `make format` (**required**)
- [x] I have checked the codestyle using `make check-codestyle` and `make lint` (**required**)
- [x] I have ensured `make pytest` and `make type` both pass. (**required**)
- [x] I have checked that the documentation builds using `make doc` (**required**)

Note: You can run most of the checks using `make commit-checks`.

Note: we are using a maximum length of 127 characters per line

Issue #535, opened 2021-08-11
Unexpected behaviour when predicting with batch norm and dropout


ayeright:
**Important Note: We do not do technical support, nor consulting** and don't answer personal questions per email.
Please post your question on the [RL Discord](https://discord.com/invite/xhfNqQv), [Reddit](https://www.reddit.com/r/reinforcementlearning/) or [Stack Overflow](https://stackoverflow.com/) in that case.

If your issue is related to a **custom gym environment**, please use the custom gym env template.

### 🐛 Bug

A neural network with dropout or batch normalisation layers should have different behaviour when training and predicting. When using PyTorch, we can switch been train and evaluation mode by using the methods `model.train()` and `model.eval()`, where `model` is an instance of `torch.nn.Module`. 

Stable-Baselines3 does not appear to switch to evaluation model when predicting. For example, if you call the [predict](https://github.com/DLR-RM/stable-baselines3/blob/master/stable_baselines3/dqn/dqn.py#L196) method of the `DQN` class and follow it through, there is nowhere in the code that switches the policy network to evaluation mode. 

This causes problems if you are using a custom policy network with either dropout or batch normalisation. If you are using batch normalisation and you try to run predict on a single observation, it will throw an error because the model is in train mode and in this case batch normalisation requires at least two data points to update the layer's statistics. If you are using dropout or batch normalisation and you run predict on multiple observations, it won't give an error but it will result in unexpected behaviour. For dropout, the network activations will be dropped randomly instead of being scaled by the dropout rate. For batch normalisation, the observations will be used to update the layer's statistics instead of simply being scaled by the existing statistics.

### To Reproduce

Steps to reproduce the behavior when predicting with a single observation (note that when predicting with multiple observations there is no error, but the behaviour will not be correct):

```python
from stable_baselines3 import DQN
from stable_baselines3.common.torch_layers import BaseFeaturesExtractor
from torch.nn import BatchNorm1d


class BatchNorm(BaseFeaturesExtractor):
    
    def __init__(self, observation_space):
        features_dim = observation_space.shape[0]
        super(BatchNorm, self).__init__(observation_space, features_dim)
        self.batch_norm = BatchNorm1d(num_features=features_dim)

    def forward(self, observations):
        return self.batch_norm(observations)


dqn = DQN(
    policy='MlpPolicy',
    env='LunarLander-v2',
    policy_kwargs=dict(features_extractor_class=BatchNorm),
)


obs = dqn.env.reset()
[...]

PR #553, opened 2021-08-27
Avoid putting target networks into training mode

ayeright:
Avoid putting target networks into training mode

## Description
Each policy is currently put into training mode by calling `self.policy.train()`. This affect policy networks with batch norm and dropout. For DQN, SAC and TD3, this call will also put the target networks into training mode. This is incorrect behaviour, as the target networks should always remain in evaluation mode. This PR aims to ensure that this is always the case.  

## Motivation and Context

It was raised in #535 that the policies are never switched between train and eval mode. This was apparently fixed in #537. However, the problem with the target networks remains.  

## Types of changes

## Checklist:


stable_baselines3/common/policies.py
```diff
@@ -194,6 +194,16 @@ def parameters_to_vector(self) -> np.ndarray:
         """
         return th.nn.utils.parameters_to_vector(self.parameters()).detach().cpu().numpy()
 
+    @abstractmethod
```
araffin, line 197: why not having a default behavior?
this would avoid duplicate code later, no?

stable_baselines3/dqn/policies.py
```diff
@@ -190,6 +203,17 @@ def _get_constructor_parameters(self) -> Dict[str, Any]:
...
+    def set_training_mode(self, mode: bool) -> None:
+        """
+        Put the policy in either training or evaluation mode.
+
+        This affects certain modules, such as batch normalisation and dropout.
+
+        :param mode: if true, set to training mode, else set to evaluation mode
+        """
+        self.q_net.set_training_mode(mode)
+        self.training = mode
```
araffin, line 215: this variable is not used, no? (nor defined in the constructor)

stable_baselines3/sac/sac.py
```diff
@@ -202,6 +202,7 @@ def train(self, gradient_steps: int, batch_size: int = 64) -> None:
                 self.actor.reset_noise()
 
             # Action by the current actor for the sampled state
+            self.actor.set_training_mode(True)
```
araffin, line 205: `self.actor` is just an alias of `self.policy.actor` so this (and the other calls to `critic`/`actor`) should not be needed, no?

stable_baselines3/td3/td3.py
```diff
@@ -158,6 +158,7 @@ def train(self, gradient_steps: int, batch_size: int = 100) -> None:
                 target_q_values = replay_data.rewards + (1 - replay_data.dones) * self.gamma * next_q_values
 
             # Get current Q-values estimates for each critic network
+            self.critic.set_training_mode(True)
```
araffin, line 161: dito

tests/test_run.py
```diff
@@ -139,3 +147,312 @@ def test_train_freq_fail(train_freq):
             train_freq=train_freq,
         )
         model.learn(total_timesteps=250)
+
+
+def test_dqn_train_with_batch_norm():
```
araffin, line 152: `test_run` are mostly for integration test (test that it runs, for the feature test, please keep it inside the `test_predict` or in another file)

tests/test_run.py
```diff
@@ -139,3 +147,312 @@ def test_train_freq_fail(train_freq):
...
+    )
+
+    (
+        q_net_bias_before,
+        q_net_running_mean_before,
+        q_net_target_bias_before,
+        q_net_target_running_mean_before,
+    ) = clone_dqn_batch_norm_stats(model)
+
+    total_timesteps, callback = model._setup_learn(total_timesteps=100, eval_env=model.get_env())
```
araffin, line 200: why not calling `.learn()` directly?

tests/test_run.py
araffin: oh I see, probably this is time to create a `test_eval_mode.py`, no?

araffin (commented):
Thanks a lot for the PR =)
Only minor comments / organization remarks

stable_baselines3/common/policies.py
ayeright: Yes, you're right. I was unsure whether to add default behaviour or not. I thought it might be safer to leave it as an abstract method, as it would force the person to consider this when implementing a new policy.

But I can add the default behaviour if you prefer?

araffin: yes, please

stable_baselines3/sac/sac.py
ayeright: The reason for this is that on lines 229-232 we have 

```
with th.no_grad():
    # Select action according to policy
    self.actor.set_training_mode(False)
    next_actions, next_log_prob = self.actor.action_log_prob(replay_data.next_observations)
```

What's happening is that, when `self.actor` is being used to get the next actions without gradients, I'm putting it into eval mode, then on line 205 I put it back into train mode before calling

```
actions_pi, log_prob = self.actor.action_log_prob(replay_data.observations)
```

I _think_ that this is the correct behaviour. What do you think?

stable_baselines3/td3/td3.py
ayeright: The reason for this is similar to the case above. 

On lines 175-177 we have 

```
# Compute actor loss
self.critic.set_training_mode(False)
actor_loss = -self.critic.q1_forward(replay_data.observations, self.actor(replay_data.observations)).mean()
```

So what I'm doing is putting `self.critic` into eval mode when updating `self.actor`, as `self.critic` is not updated at this point. Then I have to put `self.critic` back into train mode on line 161 before it's updated on the next iteration.

Again, I _think_ that this is the correct behaviour, but it would be good to hear your opinion.

stable_baselines3/common/policies.py
ayeright: No problem. I've made that change and removed the default behaviour from the other classes.

stable_baselines3/td3/td3.py
araffin: I see, that was mostly to keep the `train()` method clean

tests/test_run.py
ayeright: Yes, fair point. I have moved all the tests for this to `test_train_eval_mode.py`, including the new ones that were in `test_predict.py`.

Is that ok?

tests/test_run.py
ayeright: Do you mean call `.learn()` with `learning_starts >= total_timesteps`? I thought about this, but the on-policy algorithms don't have a `learning_starts` parameter, so I don't think it would work for them. 

I can change the tests for the off-policy algorithms if you prefer?

stable_baselines3/td3/td3.py
ayeright: To be honest, I'm not really sure it will make much difference to the training process. I can remove these calls if you prefer to keep the code clean?

tests/test_run.py
araffin: perfect =)

tests/test_run.py
araffin: you can do as we do for other tests, where we change `learning_starts` to zero if it is an off-policy algorithm, for instance https://github.com/DLR-RM/stable-baselines3/blob/master/tests/test_save_load.py#L178

ayeright: Sorry, I'm not sure I understand. The reason for this test is to check that the batch norm stats are _not_ updated when `.collect_rollouts()` is called. If I set `learning_starts = 0` before calling `.learn()`, then `.train()` will be called after each call to `.collect_rollouts()`, which will cause the batch norm stats to be updated and the test will fail.

araffin: > The reason for this test is to check that the batch norm stats are not updated when .collect_rollouts() is called. 


I see, then please do a first call to `learn()` with `num_timesteps < learning_starts` and that should do the trick, no?

stable_baselines3/td3/td3.py
araffin: Yes, please do. (if we remove it, it would only slightly update the batch norm stats which should be fine)
Also, it will allow the dropout to be active there.

tests/test_run.py
ayeright: Yes, that works. I've made that change for the off-policy algorithms.

stable_baselines3/td3/td3.py
ayeright: Cool. I've removed the extra calls to `set_training_mode`.

stable_baselines3/sac/sac.py
ayeright: As below, I've removed the extra calls to `set_training_mode`.

araffin:
I've refactored the tests and made them run faster, could you have a quick look?

I also took advantage of this PR to fix a small issue when `gradient_steps=0`.

Note: the collect rollout test for A2C/PPO is in theory not needed as it would crash if train mode was activated (as we only have one env)

ayeright:
> I've refactored the tests and made them run faster, could you have a quick look?
> 
> I also took advantage of this PR to fix a small issue when `gradient_steps=0`.
> 
> Note: the collect rollout test for A2C/PPO is in theory not needed as it would crash if train mode was activated (as we only have one env)

Thanks! All looks good

araffin (approved):
LGTM, thanks =)

araffin:
in case you still have some time, could you open a similar PR in the contrib repo?

repo: https://github.com/Stable-Baselines-Team/stable-baselines3-contrib
(should be fast, mostly copy pasting and adapting quickly to QR-DQN/TQC which are like DQN/SAC)

Otherwise, I will do it myself (but I've quick a lot of stuff on my stack already...)

ayeright:
> in case you still have some time, could you open a similar PR in the contrib repo?
> 
> repo: https://github.com/Stable-Baselines-Team/stable-baselines3-contrib
> (should be fast, mostly copy pasting and adapting quickly to QR-DQN/TQC which are like DQN/SAC)
> 
> Otherwise, I will do it myself (but I've quick a lot of stuff on my stack already...)

Yes, I should have some time this week to do that.

araffin:
>Yes, I should have some time this week to do that.

Merci =)

smanolloff:
Hello, sorry to bring back a 3-year old discussion 😬

I am very interested in the reasons behind this PR. I've noticed there are other PPO implementations which keep the network in training mode even during the experience collection phase. For me it is a "gray area" as I was unable to find a good reason for one or the other. I would be grateful if you could provide a more detailed explanation or references on the topic. Cheers 🍻

PR #584, opened 2021-09-27
Remove `sde_net_arch` + Simplify policy

araffin:
## Description

Remove `sde_net_arch` param (not very useful after all and complexify the code) + simplify code for policies
The idea is to go towards properly separated actor/critic as discussed in https://github.com/Stable-Baselines-Team/stable-baselines3-contrib/pull/40.

~Note: I still need to check that no results are changes~
Results before and after this change are identicals =)

## Motivation and Context

## Types of changes

## Checklist:


stable_baselines3/common/policies.py
```diff
@@ -171,6 +172,10 @@ def load(cls, path: str, device: Union[th.device, str] = "auto") -> "BaseModel":
         """
         device = get_device(device)
         saved_variables = th.load(path, map_location=device)
+
+        if "sde_net_arch" in saved_variables["data"]:
```
Miffyli, line 176: Maybe a warning here that the parameter is now removed and to use it one has to downgrade SB3?

stable_baselines3/common/policies.py
```diff
@@ -583,39 +577,21 @@ def forward(self, obs: th.Tensor, deterministic: bool = False) -> Tuple[th.Tenso
...
+        latent_pi, latent_vf = self.mlp_extractor(features)
         # Evaluate the values for the given observations
         values = self.value_net(latent_vf)
-        distribution = self._get_action_dist_from_latent(latent_pi, latent_sde=latent_sde)
+        distribution = self._get_action_dist_from_latent(latent_pi)
         actions = distribution.get_actions(deterministic=deterministic)
         log_prob = distribution.log_prob(actions)
         return actions, values, log_prob
 
-    def _get_latent(self, obs: th.Tensor) -> Tuple[th.Tensor, th.Tensor, th.Tensor]:
```
Miffyli, line 594: I see the point of removing this part (only few lines of code), but I wonder if some people rely on this for debugging/modification purposes. Might be better idea to keep it for now for backwards compat (yes, not an "official API" thing, but just a thought :) )

Edit: This function was also used in at least four places, so it might be worth keeping?

Miffyli (commented):
Overall LGTM, just some minor comments regarding simplifying code :)

stable_baselines3/common/policies.py
araffin: yes, I do not consider it as part of the API so I'm ok with breaking it.

The main argument against was that it was always computing latent variables for policy and value but most of the time you just need one of the two, hence the removal (it saves computation and makes things clearer by having a better separation).

Miffyli: Gotcha!

Miffyli (approved):
LGTM!

PR #605, opened 2021-10-11
Use consistent logging keys

timokau:
## Description

This PR proposed to use underscores instead of spaces in all logging keys. Most keys already follow this policy. 

## Motivation and Context

My main motivation is to make sure that all algorithms and the eval callback use the same key to log the total timesteps. That simplifies processing of the logged values. I had to decide on one of the previously used versions (underscore or space). I decided to use the underscored version, since it seems less surprising to me and is also more common in the codebase.

I noticed that only two other keys used spaces, so I changed those as well to improve consistency. This may avoid some confusion, especially since its easy to overlook the space/underscore distinction.

It would be "less breaking" to use `time/total timesteps` everywhere. That would also be fine by me, let me know if you prefer that.

I have not opened an issue first, since after reading `CONTRIBUTING.md` I thought that this was only required for new features. When I read about this requirement in the PR template I already made the change, so we might as well discuss the proposal with the added context of the PR. This is a fairly low-effort change, so I haven't lost much if it gets rejected. Treat this as an issue for now :)

## Types of changes

This may or may not be considered a bug fix.

## Checklist:

**Notice**: I have not hard-wrapped my additions to the changelog, since most other lines are not hard-wrapped either. Let me know if you want me to change that.

  - I have checked that no tests depend on the previous keys. Let me know if you want me to add any additional tests.
    - **Notice:** Most tests pass, but 5 tests fail because I have not installed `atari_py`.  I do not expect that my change would make those tests fail, but let me know if you want me to run them anyway.


timokau:
I agree that it's good to think twice about unnecessary breaking changes. I think the cost-benefit consideration is not entirely clear for the general "space replacement", but I do think that it is worth it for the steps. Logging the same thing under two different names can make processing more difficult.

Maybe releasing such a change should be postponed to the next major version update, depending on sb3's versioning policy.

Thanks for the review and the rst syntax fix :)

PR #636, opened 2021-10-29
Adjust FPS calculation to accommodate for reset_num_timesteps=False

### LINKED ISSUE #518: [Bug] FPS calculation is incorrect
**Important Note: We do not do technical support, nor consulting** and don't answer personal questions per email.
Please post your question on the [RL Discord](https://discord.com/invite/xhfNqQv), [Reddit](https://www.reddit.com/r/reinforcementlearning/) or [Stack Overflow](https://stackoverflow.com/) in that case.

If your issue is related to a **custom gym environment**, please use the custom gym env template.

### 🐛 Bug

The training FPS appears to be calculated by dividing the total number of time steps by the number of seconds that passed since the start of call to the `learn()` method. The problem is that when the `reset_num_timesteps=False` is used, the total number of time steps continues from previous learning, so we're dividing the number of time steps over all `learn()` calls by the time passed in only the current `learn()` call. This results in a very high FPS value at the beginning of each call to `learn()`. Additionally, the current time is taken between the rollout collection and the network training, so the number of rollout collections when the time is taken is always one greater than the number of calls to `train()`. Since the number of time steps is increased during rollout collection, this results in the calculated FPS value being higher than the real average value at the beginning of each call to `learn()`.

### To Reproduce

Steps to reproduce the behavior.

Please try to provide a minimal example to reproduce the bug. Error messages and stack traces are also helpful.

Please use the [markdown code blocks](https://help.github.com/en/articles/creating-and-highlighting-code-blocks)
for both code and stack traces.

```python
import gym

env = make_vec_env("CartPole-v1", n_envs=4)

model = PPO("MlpPolicy", env, verbose=1)
model.learn(total_timesteps=25000)
print("Start of second call")
model.learn(total_timesteps=25000, reset_num_timesteps=False)
```
Notice that the FPS value starts off high and then decreases at the beginning, before jumping to an extremely large value at the beginning of the second call to `learn()`. At the beginning of the second call, the FPS value is approximately equal to the total number of time steps (including those from the first call) divided by the number of seconds that passed since the start of the second call.

### Expected behavior

The FPS should stay roughly constant over time, assuming the computational power available doesn't change.

### System Info

Describe the characteristic of your environment:
 * Describe how the library was installed (pip, docker, source, ...): pip
 * GPU models and configuration: none
 * Python version: 3.9.2
 * Stable Baselines 3 version: 1.1.0
 * PyTorch version: 1.9.0
 * Gym version: 0.18.3

### Checklist

kachayev:
## Motivation and Context

closes #518 

The issue is described [here](https://github.com/DLR-RM/stable-baselines3/issues/518). 

As the solution, I updated `_setup_learn` from `BaseAlgorithm` to make the change easier both for on-policy and off-policy implementation. The idea is to track where we started when `learn` was called and compute FPS based on number of timesteps during the latest `learn` cycle. It is a breaking change, in a way, but I don't think that previous behavior was expected.

I didn't update tests as I was not able to find any test cases regarding what was recorded through the logger (there are tests for the logger itself but they only verify that the resulting log is not empty). Please, let me know if I just missed them. If there are none, I think I can try to create a new suite to cover content of the log (would be more appropriate as independent PR).

## Types of changes

## Checklist:


araffin:
Hello
thanks for the PR =)

> If there are none, I think I can try to create a new suite to cover content of the log

can't we just have a small test checking that fps outputs reasonable values? (we can hack it by manually setting `num_timesteps` or make use of `time.sleep()` to limit fps)

stable_baselines3/common/base_class.py
```diff
@@ -420,6 +420,7 @@ def _setup_learn(
             # Make sure training timesteps are ahead of the internal counter
             total_timesteps += self.num_timesteps
         self._total_timesteps = total_timesteps
+        self._num_timesteps_at_start = self.num_timesteps
```
araffin, line 423: Please also add `self._num_timesteps_at_start = 0` in the constructor

kachayev:
> can't we just have a small test checking that fps outputs reasonable values? 

Yeah, absolutely. Is there an existing test suite that I should update (don't want to create a new suite for the sake of a single test)? Maybe `test_run`?

stable_baselines3/common/base_class.py
kachayev: Sure. Would `None` be more appropriate as a value in the constructor?

kachayev: Ah, okay. `_total_timesteps` is already set to `0` in constructor. Will follow the same convention.

araffin:
> Is there an existing test suite that I should update (don't want to create a new suite for the sake of a single test)? 

The test should probably be added in  `test_logger.py`, `test_run.py` is more meant to check that all basic usage of algorithms run without throwing errors.

kachayev:
@araffin I added a very simple test for FPS. I created in-memory logger for this purpose, and a dummy environment that does nothing but sleeps each step. Let me know what do you think.

araffin (approved):
LGTM, thanks =) (I added a comment and made the test fail more obviously in case there is a regression)

PR #658, opened 2021-11-10
Add timeout handling for on-policy algorithms

### LINKED ISSUE #633: [Feature Request] Proper TimeLimit/Infinite Horizon Handling for On-Policy algorithm
### 🚀 Feature

Same as #284 but for on-policy algorithms.
The current workaround is to use a `TimeFeatureWrapper` (cf. zoo).

### Checklist

araffin:
## Motivation and Context

closes #633 

## Types of changes

## Checklist:


araffin:
@zhihanyang2022 could you have a look at that one?

zhihanyang2022:
@araffin Sorry, which part do you want me to take a look? It seems like the changes are passing tests correctly?

araffin:
> @araffin Sorry, which part do you want me to take a look? It seems like the changes are passing tests correctly?

could you review the code? (both logic and style/naming)

tests/test_gae.py
```diff
@@ -112,3 +129,43 @@ def test_gae_computation(model_class, gae_lambda, gamma, num_episodes):
...
+    if model_class == A2C:
+        value = model.policy.predict_values(obs_tensor).item()
+    else:
+        value = model.critic(obs_tensor, model.actor(obs_tensor))[0].item()
+    # True value (geometric serie with a reward of one at each step)
+    infinite_horizon_value = 1 / (1 - gamma)
+
+    if handle_timeout_termination:
+        # true value +/- 1
+        assert abs(infinite_horizon_value - value) < 1.0
```
Miffyli, line 168: What is the choice of "1.0" here? Shouldn't `infinite_horizon_value` and `value` be close to each other (e.g. `np.isclose` check?). Sounds noisy with 1.0, even when `infinite_horizon_value` is 50.

Miffyli (commented):
LGTM with few comments :). I would like second opinion if @zhihanyang2022 has time to look over this (about ~20 lines of relevant code changes).

tests/test_gae.py
araffin: there are close (infinite_horizon_value ~ 50 vs value ~ 10 without proper handling ) but because the test must run fast, I reduced the number of trainings steps a lot, so the actual precision is only below 1 (in fact, for A2C, it is really close, but for SAC it depends on the seed and there is entropy... oh wait, le me set the entropy to zero)

araffin: for reference, true value vs value when infinite horizon is properly handled or not:
```
tests/test_gae.py::test_infinite_horizon[False-SAC] 49.99999999999996 6.762823581695557
PASSED
tests/test_gae.py::test_infinite_horizon[True-A2C] 49.99999999999996 49.99942398071289
PASSED
tests/test_gae.py::test_infinite_horizon[True-SAC] 49.99999999999996 50.144901275634766
```

zhihanyang2022 (approved):
I'm not very familiar with how testing works here, but I've checked the changes made to `on_policy_algorithm.py` and I think they are correct. Adding `gamma * terminal_value` to the final reward is logically identical to what's done in OpenAI SpinUp:

https://github.com/openai/spinningup/blob/038665d62d569055401d91856abb287263096178/spinup/algos/pytorch/ppo/ppo.py#L59

araffin:
> I'm still trying to understand how testing works here, 

to give a bit more context, I created an environment where I know in advance the true value of each state (because the reward does not depends on the policy) but that is different if you look at it as an infinite or finite horizon issue.
The env is composed of 4 states (0, 1, 2, 3) that are repeated over time, depending on the max number of steps.
For a max episode length of 8, the agent will go into states 0, 1, 2, 3, 0, 1, 2, 3 and receive a reward of one for each state.
Because the remaining time is not included, we break Markov assumption and the value of the first state V(s=0) is not really well defined, it is something between 8 (discounted sum of 8 steps) and 3... (discounted sum of 4 steps).

On the other hand, if we treat the problem as infinite horizon, the true value is a geometric series `1 + 1**gamma + 1**gamma^2 + ... = 1 / (1 - gamma)` (we get a reward of one at each step and the value is the discounted sum of it on an infinite horizon)

Issue #692, opened 2021-12-14
Newline problem of Monitor's ResultsWriter in Windows environments


hsuehch:
https://github.com/DLR-RM/stable-baselines3/blob/e24147390d2ce3b39cafc954e079d693a1971330/stable_baselines3/common/monitor.py#L181
When running the codes in Windows environments, the logged csv files end each line with `\r\r\n`, making empty lines between data lines.
From Python 3 documentation, adding `newline=''` as an argument of the `open` function seems able to solve the problem.
https://docs.python.org/3/library/csv.html#examples

Miffyli:
Hey, thanks for reporting this! Can you provide a minimal code example I can copy/paste and run to replicate the (unwanted) behaviour for better clarity on what goes wrong? I think I get the idea, but I might fail replicating the issue (it will be also easier to test on linux) :)

hsuehch:
Thank you for your quick reply!
The followings are codes that can replicate the behavior, which generates a log file as attached.
The problem seems specific in Windows environments. When I ran the same codes on Ubuntu, each line ends with `\r\n` and is no problem.
```
import gym
from stable_baselines3.common.monitor import Monitor
from stable_baselines3 import DQN

env = gym.make('CartPole-v1')
env = Monitor(env, 'Train')

model = DQN('MlpPolicy', env)
model.learn(total_timesteps=1000)
``
[Train.monitor.csv](https://github.com/DLR-RM/stable-baselines3/files/7711827/Train.monitor.csv)
`

**Comment by araffin** [comment]
> The problem seems specific in Windows environments. 

But does it creates any error?
I have been using SB3 on Windows recently and I didn't experience any issue with monitor files (the plot scripts in the RL Zoo still work fine).

**Comment by Miffyli** [comment]
> But does it create any error?

Nope, but that depends on the CSV parser you use. I am not sure if CSV has formal grammar/rules for handling empty lines, but "default" Windows text editors etc will show the contents as

```
#{"t_start": 1639486796.3546653, "env_id": "CartPole-v1"}
r,l,t

23.0,23,1.257611

35.0,35,1.259605

37.0,37,1.2616

73.0,73,1.26559

...
```

This might be a headache, depending on what you want to do with the file afterwards.

hsuehch:
> (the plot scripts in the RL Zoo still work fine)

I have not tried RL Zoo yet.

> Nope, but that depends on the CSV parser you use.

[...]

PR #695, opened 2021-12-17
Eliminate extra empty lines in CSV monitor files on Windows

hsuehch:
## Description

Added ``newline="\n"`` when opening CSV monitor files so that each line ends with ``\r\n`` instead of ``\r\r\n`` on Windows while Linux environments are not affected

## Motivation and Context

Related to #692

## Types of changes

## Checklist:

### DIFF
--- a/docs/misc/changelog.rst
+++ b/docs/misc/changelog.rst
@@ -37,6 +37,7 @@ Others:
 ^^^^^^^
 - Added a warning in the env checker when not using ``np.float32`` for continuous actions
 - Improved test coverage and error message when checking shape of observation
+- Added ``newline="\n"`` when opening CSV monitor files so that each line ends with ``\r\n`` instead of ``\r\r\n`` on Windows while Linux environments are not affected (@hsuehch)
 
 Documentation:
 ^^^^^^^^^^^^^^
@@ -852,4 +853,4 @@ And all the contributors:
 @ShangqunYu @PierreExeter @JacopoPan @ltbd78 @tom-doerr @Atlis @liusida @09tangriro @amy12xx @juancroldan
 @benblack769 @bstee615 @c-rizz @skandermoalla @MihaiAnca13 @davidblom603 @ayeright @cyprienc
 @wkirgsn @AechPro @CUN-bjy @batu @IljaAvadiev @timokau @kachayev @cleversonahum
-@eleurent @ac-93 @cove9988 @theDebugger811
+@eleurent @ac-93 @cove9988 @theDebugger811 @hsuehch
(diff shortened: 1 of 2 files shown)


Miffyli:
Thanks for the PR! Have you been able to run the tests on Windows for this (e.g. `pytest tests/test_logger.py` should do the trick) to confirm that this works? I can try to remember to do this later.

araffin:
Hello,

thanks for the PR. 

before merging, I would like to try it on my Windows pc (but I can only do that next week).
I'm also thinking if we could just do `if os.name == 'nt':` so we are sure the behavior is the same for Linux/Mac OS.

hsuehch:
> Thanks for the PR! Have you been able to run the tests on Windows for this (e.g. `pytest tests/test_logger.py` should do the trick) to confirm that this works?

@Miffyli I tried to run `pytest tests/test_logger.py`, but it failed at `test_fps_logger`.
Even without the change (no `newline="\n"`), the test still failed. I have no ideas what happened.
The followings are some output messages.
```
platform win32 -- Python 3.8.10, pytest-6.2.5, py-1.11.0, pluggy-1.0.0
...
algo = <class 'stable_baselines3.a2c.a2c.A2C'>
...
        # fps should be at most max_fps
        model.learn(100, log_interval=1)
>       assert max_fps / 10 <= logger.name_to_value["time/fps"] <= max_fps
E       assert (1000 / 10) <= 57

tests\test_logger.py:340: AssertionError
```
Another failure happened when testing DQN at the same place.
About newlines, when testing using the simple codes I posted in #692, `\r\r\n` has indeed become `\r\n`.

> I'm also thinking if we could just do `if os.name == 'nt':` so we are sure the behavior is the same for Linux/Mac OS.

@araffin I have also tested the codes on an Ubuntu machine. `pytest tests/test_logger.py` passed. I also confirmed that each line ends with `\r\n` no matter whether the change (`newline="\n"`) is added or not.

Miffyli:
LGTM (tested on Windows and Linux)! Approving and merging.

@araffin For context, it seems like this error [stems from the csv DictWriter](https://docs.python.org/3/library/csv.html#csv.writer), which instructs to open with `newline=''`. Using `newline='\n'` [has the same effect](https://docs.python.org/3/library/functions.html#open) (scroll down on `newline` and see text about writing mode):

> When writing output to the stream, if newline is None, any '\n' characters written are translated to the system default line separator, os.linesep. If newline is '' or '\n', no translation takes place. If newline is any of the other legal values, any '\n' characters written are translated to the given string.

Miffyli (approved):

PR #654, opened 2021-11-06
VecHerReplayBuffer

araffin:
## Motivation and Context

- [x] I have raised an issue to propose this change ([required](https://github.com/DLR-RM/stable-baselines3/blob/master/CONTRIBUTING.md) for new features and bug fixes)

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation (update in the documentation)

## Checklist:

- [x] I've read the [CONTRIBUTION](https://github.com/DLR-RM/stable-baselines3/blob/master/CONTRIBUTING.md) guide (**required**)
- [ ] I have updated the changelog accordingly (**required**).
- [ ] My change requires a change to the documentation.
- [ ] I have updated the tests accordingly (*required for a bug fix or a new feature*).
- [ ] I have updated the documentation accordingly.
- [ ] I have reformatted the code using `make format` (**required**)
- [ ] I have checked the codestyle using `make check-codestyle` and `make lint` (**required**)
- [ ] I have ensured `make pytest` and `make type` both pass. (**required**)
- [ ] I have checked that the documentation builds using `make doc` (**required**)

Note: You can run most of the checks using `make commit-checks`.

Note: we are using a maximum length of 127 characters per line

Issue #699, opened 2021-12-21
[Bug] Large drop in mean reward using multiprocessing with make_vec_env and SAC


thiakx:
### 🐛 Bug

Large drop in mean reward using multiprocessing with make_vec_env and SAC

### To Reproduce

See code below (modified from Colab example, using latest SB3)

```python
# Install stable-baselines latest version
!pip install git+https://github.com/DLR-RM/stable-baselines3#egg=stable-baselines3[extra]

import gym
from stable_baselines3.common.env_util import make_vec_env
from stable_baselines3 import SAC, PPO
from stable_baselines3.common.evaluation import evaluate_policy
import time

def compare_multi_process(model_choice, env_id, n_timesteps, num_cpu=6, policy_type='MlpPolicy'):

    eval_env = gym.make(env_id) # environment for evaluation
    vec_env = make_vec_env(env_id, n_envs=num_cpu) # num_cpu = number of processes to use

    # Multiprocessed RL Training
    multi_model = model_choice(policy_type, vec_env, verbose=0)
    model_name = type(multi_model).__name__ # save model name as string for print 
    start_time = time.time()
    multi_model.learn(n_timesteps)
    total_time_multi = time.time() - start_time
    print(f"\nTook {total_time_multi:.2f}s for {model_name}_multi - {n_timesteps / total_time_multi:.2f} FPS")

    # Single Process RL Training
    non_multi_model = model_choice(policy_type, env_id, verbose=0)
    start_time = time.time()
    non_multi_model.learn(n_timesteps)
    total_time_single = time.time() - start_time

    print(f"Took {total_time_single:.2f}s for {model_name}_non_multi - {n_timesteps / total_time_single:.2f} FPS")
    print("Multiprocessed training is {:.2f}x faster!".format(total_time_single / total_time_multi))

    # Evaluate the trained agents
    mean_reward, std_reward = evaluate_policy(multi_model, eval_env, n_eval_episodes=10)
    print(f'Mean reward: {model_name}_multi, {mean_reward} +/- {std_reward:.2f}')

    mean_reward, std_reward = evaluate_policy(non_multi_model, eval_env, n_eval_episodes=10)
    print(f'Mean reward: {model_name}_non_multi, {mean_reward} +/- {std_reward:.2f}')

n_timesteps = 10000
env_id = "Pendulum-v0"
# env_id = "MountainCarContinuous-v0"
compare_multi_process(SAC, env_id, n_timesteps)
[...]

PR #704, opened 2021-12-22
Multiprocessing support for HerReplayBuffer

### LINKED ISSUE #720: Allow different data types in her replay buffer
It seems that HER replay buffer is always created with np.float32, while other buffers are created with same dtype as observation space. Shall we change it here? People use different dtypes to save memory. I can create a PR if this is not intentional

https://github.com/DLR-RM/stable-baselines3/blob/e9a8979022d7005560d43b7a9c1dc1ba85f7989a/stable_baselines3/her/her_replay_buffer.py#L146

### LINKED ISSUE #1031: [Feature Request] Reducing memory consumption when using HerReplayBuffer
### 🚀 Feature
- In `HerReplayBuffer`, initialize `self.self._buffer` considering dtype of each inputs

### Motivation
In the implementation of `HerReplayBuffer`, `self.self._buffer` is initialized with zeros of `np.float32`, which may lead to huge memory allocation.
If the buffer for each input is initialized with the respective dtypes, we can save memory and increase the buffer size especially when using image observations (saving images as uint8 instead of float32).

https://github.com/DLR-RM/stable-baselines3/blob/2cc1477fa28e654caa4b7bf9b367febe726513e8/stable_baselines3/her/her_replay_buffer.py#L133-L148

A similar one is implemented in `DictReplayBuffer`

https://github.com/DLR-RM/stable-baselines3/blob/2cc1477fa28e654caa4b7bf9b367febe726513e8/stable_baselines3/common/buffers.py#L530-L537

### Pitch
- This enables increasing the buffer size when using HERReplayBuffer

### Checklist

Thanks developers!

### LINKED ISSUE #1335: HER with Dict Observation Space
### ❓ Question

Hello,

I am looking to use my custom environment with HER. Currently, it works perfectly with the normal environment and it passes all the checks. The observation space is, however, a dict, as it contains multiple images, joint pos, joint vel and so on. In order to wrap the environment, I added the required method (`compute_reward`) and I modified the observation space by creating a new `Dict` observation space with the keys `observation`, `achieved_goal` and `desired_goal`. In doing so, by `observation` value is basically another `Dict`. 

Is there any way to use this `MultiInput` kind of policy or is it something not currently supported now? I was also considering making my own custom policy in order to go around the issue. Thank you!

### Checklist

- [X] If code there is, it is minimal and working
- [X] If code there is, it is formatted using the [markdown code blocks](https://help.github.com/en/articles/creating-and-highlighting-code-blocks) for both code and stack traces.

qgallouedec:
## Description

Make `HerReplayBuffer` compatible with Multiprocessing.

Closes #720 
Closes #1031 
Closes #1335

## Motivation and Context

Alternative to PR #654 with a different implementation of Multiprocessing compatibility for `HerReplayBuffer`

(Motivation: https://github.com/DLR-RM/stable-baselines3/pull/654#issuecomment-999470139)

## Types of changes

## Checklist:


araffin:
Thanks for creating the PR =) 
This is in fact the cleaner way to do it that I had in mind but no time to invest in...
What is missing? More testing? Where can I help?

buoyancy99:
Hello, 
Thank you for the PR! 

I write a simple config to test this on this [branch](https://github.com/buoyancy99/rl-baselines3-zoo/tree/vecher) of my fork of stable-baselines-zoo3. 

To reproduce, simply
```
git clone https://github.com/buoyancy99/rl-baselines3-zoo
cd rl-baselines3-zoo
git checkout vecher

# assume you installed your fork of sb3
python train.py --algo sac --env FetchPush-v1
```

One minor issue I discovered: if you change the `learning_starts` parameter to 100 in this [line](https://github.com/buoyancy99/rl-baselines3-zoo/blob/d1639c38757d855e4aa8c5b0c4cf42006a7f6d4f/hyperparams/sac.yml#L240), or anything too small, you will get the following error:

```
  File "train.py", line 195, in <module>                                                                                                                                                                                                                                          
    exp_manager.learn(model)                                                                                                             
  File "/home/boyuan/Projects/buoyancy99/rl-baselines3-zoo/utils/exp_manager.py", line 202, in learn                                     
    model.learn(self.n_timesteps, **kwargs)    
  File "/home/boyuan/anaconda3/envs/sbp/lib/python3.7/site-packages/stable_baselines3/sac/sac.py", line 301, in learn
    reset_num_timesteps=reset_num_timesteps,  
  File "/home/boyuan/anaconda3/envs/sbp/lib/python3.7/site-packages/stable_baselines3/common/off_policy_algorithm.py", line 374, in learn                                                                                                                                         
    self.train(batch_size=self.batch_size, gradient_steps=gradient_steps)                                                                
  File "/home/boyuan/anaconda3/envs/sbp/lib/python3.7/site-packages/stable_baselines3/sac/sac.py", line 199, in train                    
    replay_data = self.replay_buffer.sample(batch_size, env=self._vec_normalize_env)                                                                                                                                                                                              
  File "/home/boyuan/anaconda3/envs/sbp/lib/python3.7/site-packages/stable_baselines3/her/her_replay_buffer.py", line 217, in sample                                                                                                                                              
    return self._sample_transitions(batch_size, maybe_vec_env=env, online_sampling=True)  # pytype: disable=bad-return-type              
  File "/home/boyuan/anaconda3/envs/sbp/lib/python3.7/site-packages/stable_baselines3/her/her_replay_buffer.py", line 304, in _sample_transitions                                                                                                                                 
    episode_indices = np.random.choice(all_episode_indices, batch_size)                                                                  
  File "mtrand.pyx", line 915, in numpy.random.mtrand.RandomState.choice                                                                 
ValueError: 'a' cannot be empty unless no samples are taken
```

The same parameter works fine without error for non vectorized environment. Setting `learning_starts` to something bigger (400 with 8 envs in my case) will avoid this issue. However, I believe we should have a better error message or fix here.

I will test the training curve of vectorized her on openai environments to check if the performance is fine.

qgallouedec:
> Thanks for creating the PR =)
> This is in fact the cleaner way to do it that I had in mind but no time to invest in...
> What is missing? More testing? Where can I help?

There are two major issues:
- the offline swapping causes an error when the number of workers is greater than or equal to 2, because of a design problem (not just implementation). I have to rethink this, I'm working on it.

-the results obtained are different depending on the number of workers, which is not desirable (see curve).

Screenshot 2022-01-03 at 10 01 32

Environment: BitFlipping 8x8; algorithm: DQN, 10 random seeds; online sampling; goal selection strategy : future; evaluation frequency: 1000; evaluation episodes: 50

araffin:
> the results obtained are different depending on the number of workers, which is not desirable (see curve).

you need to update `gradient_steps` (cf. doc, same as https://github.com/DLR-RM/stable-baselines3/issues/699) 


> I have to rethink this, I'm working on it.

tell me if you have any questions.

qgallouedec:
> you need to update `gradient_steps` (cf. doc, same as #699)

Thanks, working much better:

Screenshot 2022-01-03 at 11 02 08

qgallouedec:
Currently working on a new implementation of the `HerReplayBuffer`. 

- The new implementation is more consistent with the others buffers. The goal is to no longer have to deal with the particular case of this buffer during its construction.
-  I no longer use an internal buffer (`self._buffer`) : observations, actions etc. are stored in `self.observations`, `self.actions` etc. 
- It is no longer necessary to specify a maximum episode size

There is still some work to do:

- Implement offline sampling (I don't know how to do it yet)
- Check that the new implementation gives the same results as the previous one for `n_envs == 1`.
- Check that the vectorization does not change the results (`n_envs > 1`)
- Check that the new implementation is not slower

And other checks I haven't thought of yet.

araffin:
> Currently working on a new implementation of the HerReplayBuffer.

You mean for both offline and online sampling?
The main reason for this implementation is efficiency.

btw, in the worst case, we could just say that offline sampling is not supported for now too.

qgallouedec:
> You mean for both offline and online sampling? The main reason for this implementation is efficiency.
> 
> btw, in the worst case, we could just say that offline sampling is not supported for now too.

That's good to know! If it's just a matter of efficiency for the old (master branch) implementation, I won't spend time on it. The argument will still exist but will raise an error.

> I also agree with that change ;) (but would need to implement backward compatible loading)

You mean some kind of method that allows to load transitions from the old implementation of `Her ReplayBuffer`? It would be implemented to allow to load models trained with an old version of SB3?

araffin:
> It would be implemented to allow to load models trained with an old version of SB3?

Yes, but probably for a separate PR.

> The argument will still exist but will raise an error.

but we need to keep it working with single env, does it?

qgallouedec:
> but we need to keep it working with single env, does it?

Making it work for a single environment should not be a problem. I'll do that.

qgallouedec:
Why do you need to remove termination signal due to timeout in HerReplayBuffer

https://github.com/DLR-RM/stable-baselines3/blob/e9a8979022d7005560d43b7a9c1dc1ba85f7989a/stable_baselines3/her/her_replay_buffer.py#L409

and not in other buffers ?

https://github.com/DLR-RM/stable-baselines3/blob/e9a8979022d7005560d43b7a9c1dc1ba85f7989a/stable_baselines3/common/buffers.py#L586

qgallouedec:
Things are going well. The code seems to work in all configurations.
I have one major concern though: the code is too slow!

```python
import cProfile

from stable_baselines3.common.vec_env import DummyVecEnv
from stable_baselines3 import DQN, HerReplayBuffer

with cProfile.Profile() as pr:
    env = DummyVecEnv(4 * [lambda: BitFlippingEnv(n_bits=10)])
    model = DQN("MultiInputPolicy", env, replay_buffer_class=HerReplayBuffer, learning_starts=100)
    model.learn(20000)

pr.print_stats(sort="cumtime")
```

```log
         8102837 function calls (7431115 primitive calls) in 23.612 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000   22.943   22.943 dqn.py:245(learn)
        1    0.032    0.032   22.943   22.943 off_policy_algorithm.py:306(learn)
     1244    0.109    0.000   13.899    0.011 dqn.py:171(train)
     1244    0.286    0.000   11.509    0.009 her_replay_buffer.py:133(sample)
     1244    0.075    0.000   10.845    0.009 her_replay_buffer.py:262(_create_virual_trans)
     1244    0.223    0.000   10.393    0.008 her_replay_buffer.py:188(sample_goals)
     1244    0.091    0.000    9.851    0.008 her_replay_buffer.py:197(<listcomp>)
    31100    7.226    0.000    9.760    0.000 her_replay_buffer.py:219(_get_episode_from_uid)
     1250    0.082    0.000    9.004    0.007 off_policy_algorithm.py:507(collect_rollouts)
     5000    0.047    0.000    5.229    0.001 off_policy_algorithm.py:442(_store_transition)
     5000    4.824    0.001    5.042    0.001 her_replay_buffer.py:92(add)
    32344    1.208    0.000    2.623    0.000 index_tricks.py:148(__getitem__)
     5000    0.013    0.000    1.903    0.000 off_policy_algorithm.py:365(_sample_action)
     4975    0.026    0.000    1.888    0.000 dqn.py:215(predict)
     4505    0.041    0.000    1.821    0.000 policies.py:307(predict)
     5000    0.011    0.000    1.491    0.000 base_vec_env.py:154(step)
```

Almost half of the learning time is spent on executing this [list by comprehension](https://github.com/qgallouedec/stable-baselines3/blob/423d0579c2e01037c246adc593f71ed7b5811181/stable_baselines3/her/her_replay_buffer.py#L197).
I am working on a solution.

stable_baselines3/common/off_policy_algorithm.py
```diff
@@ -266,7 +249,8 @@ def load_replay_buffer(
 
         if isinstance(self.replay_buffer, HerReplayBuffer):
             assert self.env is not None, "You must pass an environment at load time when using `HerReplayBuffer`"
-            self.replay_buffer.set_env(self.get_env())
+            compute_reward = self.env.get_attr("compute_reward", indices=[0])[0]
```
araffin, line 252: this won't probably work with `SubprocVecEnv`, no?

araffin:
> 

Because the timeout is handled at sampling time for the classic replay buffer (normally).
But we can probably do the same for the online sampling.

qgallouedec:
> But we can probably do the same for the online sampling.

Done

araffin:
>  I have one major concern though: the code is too slow!

That's because you removed the max episode length, right?

Btw, was that needed?
Now the implementation seems really complex...

qgallouedec:
> Now the implementation seems really complex...

Since my last post, there have been many changes, and the code is considerably simpler and faster.

> That's because you removed the max episode length, right?

No, that's not the main reason. 
In my opinion, if this new implementation is slower, it is because the interactions are no longer stored according to the structure `(episode_idx, trans_idx, env_idx)`, which is quite efficient. My motivation tu use a different structure was not to remove `max_episode_length`, even if it is a convenient thing that I do not need it anymore. Here is my thoughts:
When interacting with multiple environments, there is no guarantee that all environments will reach a terminal state at the same time. So:
1. If you want to keep the structure `(episode_idx, trans_idx, env_idx)`, then you would have to manage a list of indexes of transitions and episode for each environment. It seems pretty complicated. I didn't even try.
2. Alternatively, you could use a structure like `(episode_idx, trans_idx)`. When an environment finishes an episode, it starts writing the next one on the next available `episode_idx`. You just need to keep track of the index of the episode being written for each environment. This is what I did in my first commits of this PR:

https://github.com/qgallouedec/stable-baselines3/blob/7942417881ed46a739565c6617676d466e3287db/stable_baselines3/her/her_replay_buffer.py#L87

This solution works quite well but does not seem completely satisfactory in the sense that it is not consistent with other replay buffers: `_buffer` dictionary, or internal buffer for offline sampling, different arguments when building it, different storage for `done` etc.
These inconsistencies often require to distinguish the `HerReplayBuffer` case outside the buffer, which can make its use difficult.

3. So I thought of a third solution: make an effort on the consistency of `HerReplayBuffer` to make it easier to use. Of course, this required to abandon the above-mentioned efficient storage structure. Transitions are no longer stored per episode like in all other buffers. 
In the current state of the code, I only keep track of the episode start indexes, and the episode size.

Drawback: 
- it's a bit slower (2x empirically).
Advantages:
- use is much easier, because consistent with other buffers
- the code is more concise (545 lines versus 340, for what it's worth)
- it is no longer necessary to specify a maximum episode size
- it is memory efficient (no need to store 0 for episodes that don't reach the size limit)

araffin:
>  (episode_idx, trans_idx, env_idx), then you would have to manage a list of indexes of transitions and episode for each environment. It seems pretty complicated. I didn't even try.

Actually, this structure was never used... I agree we should move to `(episode_idx, trans_idx)`

> This is what I did in my first commits of this PR:

Your first implementation was working, right?

>  These inconsistencies often require to distinguish the HerReplayBuffer case outside the buffer, which can make its use difficult.

I agree that inconsistencies should be avoided when possible.

> it's a bit slower (2x empirically).

2x is not a bit...

> use is much easier, 

use is easier but it is harder to read, right?

Let me check the current code, we might still reference the more consistent implementation for people interested but I'm not sure we will keep it mainly for the two reasons above (computation time and readability).

araffin:
> Let me check the current code, we might still reference the more consistent implementation for people interested but I'm not sure we will keep it mainly for the two reasons above (computation time and readability).

so at the end, readability is fine but I think current implementation is not right (see my note for the "future" sampling strategy + comment on subproc env), the budget in the performance check should not be changed, it is usually an indicator that something is wrong.

> Agree; I will give a more precise value.

Please do =)

stable_baselines3/her/her_replay_buffer.py
```diff
@@ -325,22 +325,24 @@ def _sample_goals(self, batch_inds: np.ndarray, env_indices: np.ndarray) -> np.n
...
             # replay with final state of current episode
-            goal_ep_inds = batch_ep_length - 1
+            transition_indices_in_episode = batch_ep_length - 1
 
         elif self.goal_selection_strategy == GoalSelectionStrategy.FUTURE:
             # replay with random state which comes from the same episode and was observed after current transition
-            batch_idx_within_ep = batch_inds - batch_ep_start
-            goal_ep_inds = np.random.randint(batch_idx_within_ep, batch_ep_length)
+            current_indices_in_episode = batch_inds - batch_ep_start
+            # Note(antonin): current implementation does not seem correct, a +1
```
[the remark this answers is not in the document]
araffin: > Should we strictly stick to the strictly future strategy?

Yes, otherwise we are not following the paper/original implementation (I expect it to have some impact on performance even if it is a pain to handle special cases)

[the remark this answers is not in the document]
qgallouedec: 👍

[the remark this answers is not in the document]
buoyancy99: I like the current version without max_episode_length from a research perspective, if performance trade off isn't too bad. Enforcing max_episode_length can make the env no longer a MDP and cause issues for off-policy algorithms, especially when `max_episode_length` is set to something small. A detailed explanation can be seen [here](https://github.com/rlworkgroup/metaworld/issues/192#issuecomment-752284550).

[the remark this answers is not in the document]
araffin: see https://github.com/DLR-RM/stable-baselines3/issues/284
for how we handle infinite horizon problem with off policy algos.

tests/test_her.py
```diff
@@ -73,11 +80,16 @@ def test_her(model_class, online_sampling, image_obs_space):
...
-def test_goal_selection_strategy(goal_selection_strategy, online_sampling):
+def test_goal_selection_strategy(n_envs, goal_selection_strategy, online_sampling):
     """
     Test different goal strategies.
     """
-    env = BitFlippingEnv(continuous=True)
+    if n_envs > 1 and not online_sampling:
+        pytest.skip("Offline sampling is not compatible with multiprocessing")
+
+    env_fn = lambda: BitFlippingEnv(continuous=True)
```
araffin, line 90: is there a reason you reverted to lamndas? (see linter warnnigs)

qgallouedec: None, I did not see your commit. I am reverting this change.

araffin:
Looking at the test, the +1 for the future strategy does actually make a big difference? (the performance test was failing before even with almost twice more budget)

qgallouedec:
> Looking at the test, the +1 for the future strategy does actually make a big difference? (the performance test was failing before even with almost twice more budget)

Indeed! I didn't expect it.

Screenshot 2022-01-13 at 18 41 38

Btw, there is no reason that online sampling gives better results. I only used 10 seeds, it might explain the gap.

araffin:
> Btw, there is no reason that online sampling gives better results.

There is. In fact, in my experience, the two are not equivalent (and the online sampling usually but not always give better results):
- with online sampling, you choose random episodes and then transitions inside those episodes
- with offline sampling, you choose random transitions in the entire buffer, so you are more likely to sample longer episodes I would say

The storage of virtual transitions is also different and I'm not sure if it is equivalent.

qgallouedec:
In the old implementation, the environment is an attribute of `HerReplayBuffer` only to compute the reward. I don't find this ideal because some people might want a reward function other than the environment one.

So I replaced `env` by `compute_reward` in the constructor. If works perfectly when the environment is a  `DummyVecEnv`. But, as @araffin suspected, it does not work when it is a `SubprocVecEnv`.. See the code below which raises a `TypeError` that I can not explain.


```python
from stable_baselines3.common.envs.bit_flipping_env import BitFlippingEnv
from stable_baselines3.common.vec_env import SubprocVecEnv
from stable_baselines3 import HerReplayBuffer, DQN


def test_her(n_envs, image_obs_space, vec_env_cls):
    env_fn = lambda: BitFlippingEnv(n_bits=4, image_obs_space=image_obs_space)
    env = make_vec_env(env_fn, n_envs, vec_env_cls=vec_env_cls)
    model = DQN("MultiInputPolicy", env, replay_buffer_class=HerReplayBuffer, learning_starts=100)
    model.learn(total_timesteps=1000)


if __name__ == "__main__":
    test_her(n_envs=1, image_obs_space=True, vec_env_cls=SubprocVecEnv)
```

As I have very little experience with the multiprocessing library, I ask for help. 
Should we still want to pass `compute_reward` as an argument?
If so, is it reasonable to handle this in this PR?

qgallouedec:
> yes, because the `compute_reward` function is in another process, not accessible directly from the main one. Btw, please add a separate test, not multiply the number of test by 2 (as done currently) to save time.

You are right, I will reduce the number of tests. However, if you look at the test logs, you'll see that the failures occur only for `image_obs_space=True` and `vec_env_cls=SubprocVecEnv` which is rather unexpected. So I will at least keep these for the tests.

> Sounds like an uncommon case, I would fork SB3 in that case if needed. Or do you have additional examples where it is needed?

Indeed. I find it more intuitive to pass only `compute_reward` as an argument. But I'll stick to the core of the PR.

qgallouedec:
Here we compare old (current master branch) and new implementation (of this fork).

Screenshot 2022-01-18 at 13 33 15

Results are obtained using 10 random seeds on the BitFlipping environment with 6 bits. The following hyperparameters were used:

| Hyperparameters  | Value |
|-|-|
| `target_update_interval` | 500 (DQN and SAC only) |
| `exploration_final_eps`| 0.02 (DQN only) |
| `n_sampled_goal` | 2 |
| `learning_rate` | 5e-4 |
| `train_freq` | 1 |
| `gradient_steps` | 1 |
| `learning_starts` | 100 |
| `batch_size` | 32 |
| `buffer_size` | 1e5 |

| Evaluation  | Value |
|-|-|
| `total_timesteps` | 8000 |
| `n_eval_episodes` | 50 |
| `eval_freq` | 400 |

The curves do not overlap exactly. The new implementation behaves sometimes better, sometimes worse. Now we have to determine the reason. Maybe it is just a statistical effect. I will try to increase the number of tests for the most different configurations.

araffin:
> Maybe it is just a statistical effect.

Most probable yes.
I would also do some testing with harder env (cf. the RL Zoo with highway env and then other like pick and place).

But most important right now is to quantify how much slower it is.

qgallouedec:
> Most probable yes.

I am currently performing the same experiments with more seeds to definitively answer this question.

> But most important right now is to quantify how much slower it is.

Then I'll do that. 
If it's too slow and you don't want to merge, I'll continue to maintain this branch until a faster version is available.

> I would also do some testing with harder env (cf. the RL Zoo with highway env and then other like pick and place).

Finally I'll do that.

qgallouedec:
With 30 seeds. All settings seem correct except for online sampling with the future goal selection strategy. I will investigate.

Screenshot 2022-01-19 at 20 59 34

qgallouedec:
New implementation is slower. Running trainings with RL Zoo with 20000 timesteps:

| Algorithm | Environment             | Old  | New  | Realtive duration |
|-----------|-------------------------|------|------|-------------------|
| TQC       | `parking-v0`            | 250  | 253  | + 1.2%            |
| TQC       | `FetchReach-v1`         | 274  | 378  | + 37%             |
| TQC       | `FetchPush-v1`          | 411  | 879  | + 113%            |
| TQC       | `FetchPickAndPlace-v1`  | 364  | 876  | + 140%            |
| TQC       | `FetchSlide-v1`         | 366  | 869  | + 137%            |

qgallouedec:
Finally some good news about performance. After a lot of testing and profiling, I finally found an easy and fast way to vectorize HER. I also fixed the normalization ([84ed129](https://github.com/DLR-RM/stable-baselines3/pull/704/commits/84ed129311d0d306d28c1139fce1700eac131fb5)). 
The new results show preformances extremely close to the non-vectorized version (master), and fps also very close (for 1 env obviously). 

In summary:

| Environment\FPS | master | Vec HER |
|-|-|-|
| FetchReach-v1 | 229 | 225 |

(there is no comparison between the environments, the tests were executed on different settings)

Experiment tracking can be found [here](https://wandb.ai/qgallouedec/vec-her-sb3/workspace?workspace=user-qgallouedec).

This comment and the wandb project will continue to be updated with new results.

araffin:
> . After a lot of testing and profiling, I finally found an easy and fast way to vectorize HER.

Good to hear =)

> Experiment tracking can be found [here](https://wandb.ai/qgallouedec/vec-her-sb3/workspace?workspace=user-qgallouedec).

404... I guess it is private

btw, looking at the code, do you think it would make sense and simplify things if we drop offline sampling?

qgallouedec:
> 404... I guess it is private

Now it's public

> do you think it would make sense and simplify things if we drop offline sampling?

Yes, it would simplify the code, and I don't think this option is used anyway.

qgallouedec:
Report: https://api.wandb.ai/links/qgallouedec/9qo886eq

## TODO

### Have at least 5 reference (master) runs for

- [ ] PandaReach-v1
- [ ] PandaPush-v1
- [ ] PandaSlide-v1
- [ ] PandaPickAndPlace-v1
- [ ] FetchReach-v1
- [ ] FetchPush-v1
- [ ] FetchSlide-v1
- [ ] FetchPickAndPlace-v1
- [ ] parking-v0 

### Have at least 5 runs for the new implementation

- [ ] PandaReach-v1
- [ ] PandaPush-v1
- [ ] PandaSlide-v1
- [ ] PandaPickAndPlace-v1
- [ ] FetchReach-v1
- [ ] FetchPush-v1
- [ ] FetchSlide-v1
- [ ] FetchPickAndPlace-v1
- [ ] parking-v0

qgallouedec:
There are still 3 reference run running but it won't change anything. This PR is now ready to review.

docs/misc/changelog.rst
```diff
@@ -13,12 +13,15 @@ Breaking Changes:
 - Removed shared layers in ``mlp_extractor`` (@AlexPasqua)
 - Refactored ``StackedObservations`` (it now handles dict obs, ``StackedDictObservations`` was removed)
 - You must now explicitely pass a ``features_extractor`` parameter when calling ``extract_features()``
+- Dropped offline sampling for ``HerReplayBuffer``
```
araffin, line 16: new her replay buffer is only not backward compatible, no?

stable_baselines3/her/her_replay_buffer.py
```diff
@@ -185,347 +116,235 @@ def set_env(self, env: VecEnv) -> None:
...
+    def _sample_goals(self, batch_inds: np.ndarray, env_indices: np.ndarray) -> np.ndarray:
         """
         Sample goals based on goal_selection_strategy.
-        This is a vectorized (fast) version.
 
-        :param episode_indices: Episode indices to use.
-        :param her_indices: HER indices.
-        :param transitions_indices: Transition indices to use.
-        :return: Return sampled goals.
+        :param trans_coord: Coordinates of the transistions within the buffer
```
araffin, line 310: outdated docstring?

stable_baselines3/her/her_replay_buffer.py
```diff
@@ -185,347 +116,235 @@ def set_env(self, env: VecEnv) -> None:
...
             # replay with final state of current episode
-            transitions_indices = self.episode_lengths[her_episode_indices] - 1
+            transition_indices_in_episode = batch_ep_length - 1
 
         elif self.goal_selection_strategy == GoalSelectionStrategy.FUTURE:
             # replay with random state which comes from the same episode and was observed after current transition
-            transitions_indices = np.random.randint(
-                transitions_indices[her_indices], self.episode_lengths[her_episode_indices]
-            )
+            current_indices_in_episode = batch_inds - batch_ep_start
```
araffin, line 322: does that work also when the episode is split in the buffer?

araffin: oh i see, you do the wrapping later but I still have a doubt because of `np.random.randint(current_indices_in_episode, batch_ep_length)`

tests/test_her.py
```diff
@@ -26,19 +25,23 @@ def test_import_error():
     assert "documentation" in str(excinfo.value)
 
 
+@pytest.mark.parametrize("n_envs", [1, 2])
 @pytest.mark.parametrize("model_class", [SAC, TD3, DDPG, DQN])
-@pytest.mark.parametrize("online_sampling", [True, False])
 @pytest.mark.parametrize("image_obs_space", [True, False])
-def test_her(model_class, online_sampling, image_obs_space):
+def test_her(n_envs, model_class, image_obs_space):
```
araffin, line 31: why test multiple envs here if they are tested in `test_multiprocessing`?

tests/test_her.py
```diff
@@ -314,64 +309,31 @@ def test_full_replay_buffer():
...
             goal_selection_strategy="future",
-            online_sampling=True,
-            max_episode_length=n_bits,
         ),
         gradient_steps=1,
         train_freq=4,
         policy_kwargs=dict(net_arch=[64]),
-        learning_starts=1,
-        buffer_size=20,
+        learning_starts=n_bits * n_envs,
```
araffin, line 316: maybe we should add a warning on the her doc page? (if it's not there already)

araffin (commented):
Overall looks like a good improvement over the previous implementation =)
I'm also curious of how fast we can make things if we combine it with SBX.

docs/misc/changelog.rst
qgallouedec: Indeed. I don't know how much effort it would take to make it backward compatible.

stable_baselines3/common/off_policy_algorithm.py
```diff
@@ -177,34 +177,16 @@ def _setup_model(self) -> None:
...
-            self.replay_buffer = HerReplayBuffer(
-                self.env,
-                self.buffer_size,
-                device=self.device,
-                replay_buffer=replay_buffer,
-                **self.replay_buffer_kwargs,
-            )
-
         if self.replay_buffer is None:
+            args = ()
```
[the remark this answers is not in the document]
qgallouedec: Indeed, we could use `self.replay_buffer_kwargs`. I don't mind modifying this, tell me if you prefer this option

stable_baselines3/her/her_replay_buffer.py
```diff
@@ -1,87 +1,64 @@
...
         self,
-        env: VecEnv,
         buffer_size: int,
-        device: Union[th.device, str] = "auto",
-        replay_buffer: Optional[DictReplayBuffer] = None,
-        max_episode_length: Optional[int] = None,
+        observation_space: spaces.Space,
+        action_space: spaces.Space,
+        env: VecEnv,
+        device: Union[th.device, str] = "cpu",
```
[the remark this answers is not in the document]
qgallouedec: It should be auto

stable_baselines3/her/her_replay_buffer.py
```diff
@@ -1,87 +1,64 @@
...
+        n_envs: int = 1,
+        optimize_memory_usage: bool = False,
```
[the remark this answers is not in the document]
qgallouedec: yes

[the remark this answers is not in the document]
qgallouedec: Actually, it's already done here.

https://github.com/qgallouedec/stable-baselines3/blob/38d7c871f388af982c96ac1dac13cf7cbcd6b5b5/stable_baselines3/common/buffers.py#L525

It's enough, do you agree?

stable_baselines3/her/her_replay_buffer.py
```diff
@@ -185,347 +116,235 @@ def set_env(self, env: VecEnv) -> None:
...
+            observations/rewards when sampling, defaults to None
+        :return: Samples, with new desired goals and new rewards
+        """
+        # Get infos and obs
+        obs = {key: obs[batch_inds, env_indices, :] for key, obs in self.observations.items()}
+        next_obs = {key: obs[batch_inds, env_indices, :] for key, obs in self.next_observations.items()}
+        # For speed up we replace this
+        # infos = copy.deepcopy(self.infos[batch_inds, env_indices])
+        # by
+        infos = [{} for _ in range(len(batch_inds))]
```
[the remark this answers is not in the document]
qgallouedec: This is an omission. I noticed that the info dict deepcopy was taking a non-negligible amount of time so I removed it because the info is rarely (never to my knowledge) used to compute the reward. I'll put it back and see what it implies in terms of time.

stable_baselines3/her/her_replay_buffer.py
qgallouedec: indeed

stable_baselines3/her/her_replay_buffer.py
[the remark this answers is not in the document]
araffin: then, maybe add an option to activate it?
(I used it in the past to compute continuity cost for instance)

stable_baselines3/common/off_policy_algorithm.py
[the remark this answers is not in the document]
araffin: if self.replay_buffer_kwargs is already defined, yes.
That's mostly to prevent breaking things if we change the order of the arguments.

stable_baselines3/her/her_replay_buffer.py
qgallouedec: Yes, it is handled here:

```python
transition_indices = (transition_indices_in_episode + batch_ep_start) % self.buffer_size
```

I also have a doubt for FUTURE when episode is split. I changed it. If it worked before, it was a fluke.

tests/test_her.py
qgallouedec: I've just added a error message instead, do you think it's enough?

stable_baselines3/common/off_policy_algorithm.py
```diff
@@ -125,10 +125,14 @@ def __init__(
...
-            replay_buffer_kwargs = {}
-        self.replay_buffer_kwargs = replay_buffer_kwargs
+        if replay_buffer_class is None:
+            if isinstance(self.observation_space, spaces.Dict):
+                self.replay_buffer_class = DictReplayBuffer
+            else:
+                self.replay_buffer_class = ReplayBuffer
+        else:
+            self.replay_buffer_class = replay_buffer_class
+        self.replay_buffer_kwargs = replay_buffer_kwargs if replay_buffer_kwargs is not None else {}
```
araffin, line 135:
```suggestion
        self.replay_buffer_kwargs = replay_buffer_kwargs or {}
```

tests/test_her.py
qgallouedec: Since I set `n_envs=2` for almost all tests, I think I'd better remove `test_multiprocessing`. Do you agree?

stable_baselines3/her/her_replay_buffer.py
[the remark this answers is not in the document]
qgallouedec:
```python
import panda_gym

from stable_baselines3 import SAC, HerReplayBuffer

env = make_vec_env("PandaPush-v1", n_envs=4)

SAC("MultiInputPolicy", env, replay_buffer_class=HerReplayBuffer, verbose=1, learning_starts=1_000).learn(20_000)
```

| info deepcopy | with | without |
|-|-|-|
| FPS | 249 | 257 |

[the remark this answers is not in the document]
araffin: I added an argument for it, disabled by default

tests/test_her.py
araffin: It does test the sub processes with it, I removed the `n_envs` from that test though.
For the rest I'm ok, even though I find it a bit weird that `n_envs=2` in all the other tests.

araffin:
@qgallouedec The implementation is still using inclusive future, right? (see https://github.com/DLR-RM/stable-baselines3/pull/704#issuecomment-1012363777)
If so, we should add a note.

Also, since we fixed https://github.com/DLR-RM/stable-baselines3/pull/704#discussion_r1128648127, could you do one quick run on the same machine with the same seed, with and without the fix?

Once it's check that nothing changed, I'll be happy to merge, thanks for the work =)

Side remark: it looks like the ideas from that implementation could be combined with https://github.com/DLR-RM/stable-baselines3/pull/81 to provide a n-step replay buffer (that won't be fully correct with SAC though).

qgallouedec:
> @qgallouedec The implementation is still using inclusive future, right? (see [#704 (comment)](https://github.com/DLR-RM/stable-baselines3/pull/704#issuecomment-1012363777)) If so, we should add a note.

Yes.
You mean, a note in the HER documentation to indicate that we use the future inclusive?

> Also, since we fixed [#704 (comment)](https://github.com/DLR-RM/stable-baselines3/pull/704#discussion_r1128648127), could you do one quick run on the same machine with the same seed, with and without the fix?

I'll do that.

araffin:
> You mean, a note in the HER documentation to indicate that we use the future inclusive?

yes and in the class docstring + maybe a one line comment above the sampling.

qgallouedec:
> could you do one quick run on the same machine with the same seed, with and without the fix?

There was indeed a silent error. In this case, another transition of the buffer was sampled. However, it happened very rarely. For a buffer of 1M and an episode length of 50, it affected one sample in 40k.
In openrlbenchmark, by a happy coincidence, it never happened because the episode length divides the buffer size (no cut episodes).

araffin:
Note: I prepared a PR to update the RL Zoo too, but it needs a new release of SB3.

araffin:
@qgallouedec I could give that PR a try with SBX (https://github.com/araffin/sbx/pull/11)
and got good results in 1e5 timesteps in 30 minutes using DroQ configuration =) (SBX is 2-3x faster than SB3)

Unfortunately, I couldn't see the usefulness of multiple env for now :/ (the bottleneck is not in the data collection, I need to see if it's the gradient update or the sampling when using SBX).

My command (green curve below):

`python train.py --algo tqc --env PandaPush-v1 -P -c hyperparams/droq.yml -params policy_delay:5 gradient_steps:5 learning_starts:1000 batch_size:1024 gamma:0.98 --log-interval 100 --n-eval-envs 5 --eval-episodes 20`
(DroQ: `dict(net_arch=[512, 512, 512], n_critics=2, layer_norm=True, dropout_rate=0.01)` and `policy_delay=gradient_steps`)

Training_Success_Rate_time

Training_Success_Rate

araffin (approved):
LGTM, thanks =)

qgallouedec:
The runs used to motivate this PR are going to be moved (copying is currently not possible) to openrlbenchmark. The report is going to be deleted. So here is the pdf version.

[SB3 - Vectorized HER | vec-her-sb3 – Weights & Biases.pdf](https://github.com/DLR-RM/stable-baselines3/files/11087388/SB3.-.Vectorized.HER.vec-her-sb3.Weights.Biases.pdf)

PR #756, opened 2022-02-03
HumanOutputFormat: make length configurable, throw error if keys alias

### LINKED ISSUE #750: [Bug] HumanOutputFormat aliases keys with same truncated name
### 🐛 Bug

In `HumanOutputFormat.write`, the output to write is constructed via `key2str[self._truncate(key)]`. Different keys with the same prefix can be mapped to the same truncated key, in which case they will silently overwrite each others value. The user will only see one output for all the keys where multiple were expected truncated to the same, and which key's value "wins" could even be non-deterministic if the call to `write()` happens in different orders.

Happy to work on a PR to fix this once we decide what the right behavior should be. 

I lean towards making `max_size` for `self._truncate` configurable, and perhaps just bumping the default `max_size` which is quite small, especially when you start logging things hierarchically: e.g. we ran into this bug with the keys `agent/rollout/ep_{len,rew}_mean` getting aliased.

In the case that there is still aliasing after `max_size`, I lean towards just printing out both -- we could swap out a dict for a multidict to make this work. Or we could throw an error, but that seems a bit aggressive, especially as other output formats will be fine.

Thoughts?

### To Reproduce

Minimal example of error:

```python

def main():
    log = logger.configure(format_strings=["stdout"])
    log.record("foo", 42) # correct
    long_prefix = "foobar" * 4
    for i in range(3):
        log.record(long_prefix + str(i), i)
    log.dump()

if __name__ == "__main__":
    main()
```

Produces:
```
$ python sb_logger.py 
--------------------------------
| foo                     | 42 |
| foobarfoobarfoobarfo... | 2  |
--------------------------------
```

### Expected behavior

```
$ python sb_logger.py 
--------------------------------
| foo                     | 42 |
| foobarfoobarfoobarfo... | 1  |
| foobarfoobarfoobarfo... | 2  |
| foobarfoobarfoobarfo... | 3  |
--------------------------------
```

### System Info

Bug is present on master, 54bcfa4544315fc920be0944fc380fd75e2f7c4a

### Checklist

- [X] I have read the [documentation](https://stable-baselines3.readthedocs.io/en/master/) (**required**)
- [X] I have provided a minimal working example to reproduce the bug (**required**)

AdamGleave:
## Description

Adds an attribute and `__init__` keyword argument `max_length` to `HumanOutputFormat`, controlling the maximum length before truncation of keys and values output. Increases the default from 23 to 36, which is chosen to keep the maximum output width <= 79 character width (should be safe everywhere). Raises a `ValueError` if multiple keys are truncated to the same value as this would otherwise cause information loss in logging.

**NOTE**: Merge https://github.com/DLR-RM/stable-baselines3/pull/757 first as this PR is based on top of that (so CI passes and we do not have messy commit history).

## Motivation and Context
Closes #750.

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist:


stable_baselines3/common/logger.py
```diff
@@ -114,12 +114,21 @@ def write_sequence(self, sequence: List) -> None:
 
 
 class HumanOutputFormat(KVWriter, SeqWriter):
-    def __init__(self, filename_or_file: Union[str, TextIO]):
+    max_length: int
```
Miffyli, line 117: Hmm any reasoning to add this here? For the docstring? But the __init__ also has docstring. I lean towards not using this unique syntax for one class in the whole codebase, if possible :)

stable_baselines3/common/logger.py
```diff
@@ -159,7 +168,10 @@ def write(self, key_values: Dict, key_excluded: Dict, step: int = 0) -> None:
             if tag is not None and tag in key:
                 key = str("   " + key[len(tag) :])
 
-            key2str[self._truncate(key)] = self._truncate(value_str)
+            truncated_key = self._truncate(key)
+            if truncated_key in key2str:
+                raise ValueError(f"Key '{key}' truncated to " f"'{truncated_key}' that already exists.")
```
Miffyli, line 173:
```suggestion
                raise ValueError(f"Key '{key}' truncated to " f"'{truncated_key}' that already exists. Consider increasing `max_length`.")
```

Miffyli (commented):
LGTM, just minor comments :)

stable_baselines3/common/logger.py
AdamGleave: I just added this to document that `max_length` is part of the public API and it's OK for users to modify the attribute. Since it can be awkward to set the `max_length` keyword argument in `__init__` if using `build_output_format`. I guess that's implicit by it not being `_max_length`.

Good point about this type annotation style not being used elsewhere in the codebase, though. I could just make a class docstring and document it there, or add a `@property` and `@property.setter`. What do you think?

Miffyli: Good to learn about new tricks like this :D.

Hmmm I would personally opt for adding a docstring/class-docstring that instructs to modify the parameter in place. `@property` stuff would be cleaner, yes, but again something that is not much used elsewhere in the code.

AdamGleave: OK, added a class docstring now (and updated Docker so GitLab passes again.)

Miffyli (approved):
LGTM :)

PR #780, opened 2022-02-19
Add Gym 0.26 support

### LINKED ISSUE #271: [Bug] Allow noop_max to be 0 and make ClipReward actually clip rewards
In [atari_wrappers.py](https://github.com/DLR-RM/stable-baselines3/blob/b8c72a53489c6d80196a1dc168835a2f375b868d/stable_baselines3/common/atari_wrappers.py), allow `noop_max` to be `0` to make it possible to have a completely deterministic Atari environment. 

Also in the same file, having `ClipRewardEnv` bin, and not clip rewards, causes problems when there's a small amount of reward noise in the environments, e.g., actual reward of `0.0` on injecting noise changes to `0.01` and this gets "clipped" to `1.0`. I know that many libraries are implementing it like done here (i.e. binning), but is there a reason we can't change it here?

I have a patched fork with the fix [here](https://github.com/RaghuSpaceRajan/stable-baselines3/commit/17ff9b872f85ff12e916c4eaaea3128578866aba). Shall I create a pull request?

### LINKED ISSUE #840: Support Gym 0.23.1
When installing Stable Baselines 3, Pip uninstalled my existing Gym version 0.23.1 and instead installed Gym 0.21.0, probably because that's what Stable Baselines 3 specifies. It would be nice if I could use the latest version of Gym with Stable Baselines 3.

### Checklist

### LINKED ISSUE #1156: [Feature Request] Pass all gym Environments in this repo through the env checkers during testing
### 🚀 Feature

Both the Gym repo and this repo have an `env_checker` but currently the environments in this repo (most of them are for testing purposes) are not checked against these checkers. This would help catch faulty implementations of changes in the gym API (e.g. the new render API).

I realise this is actually a test on 'test code' but I think it would help maintaining the tests. 

### Motivation

_No response_

### Pitch

_No response_

### Alternatives

_No response_

### Additional context

_No response_

### Checklist

carlosluis:
**EDIT: Please use the Gymnasium branch instead of this one: https://github.com/DLR-RM/stable-baselines3/pull/1327**
(the new branch is compatible with gym 0.21, gym 0.26 and gymnasium)

**See comment https://github.com/DLR-RM/stable-baselines3/pull/780#issuecomment-1164774401 to use this PR**

To install SB3 with gym 0.26+ support:

```
pip install git+https://github.com/carlosluis/stable-baselines3@fix_tests
```
(or as a requirement: `git+https://github.com/carlosluis/stable-baselines3@fix_tests#egg=stable_baselines3[extra,tests,docs]`  and `"sb3_contrib @ git+https://github.com/Stable-Baselines-Team/stable-baselines3-contrib@feat/new-gym-version"` in a `setup.py` see https://github.com/DLR-RM/rl-baselines3-zoo/pull/256)

Note: if you want to use gymnasium, you can do:
```python
import sys
import gymnasium
sys.modules["gym"] = gymnasium
```
before any stable-baselines3 import.

**for native gymnasium support**, you can take a look at https://github.com/DLR-RM/stable-baselines3/pull/1327 or install it with:
```
pip install git+https://github.com/DLR-RM/stable-baselines3@feat/gymnasium-support
```

## Description
Gym 0.26 has been released and with it breaking changes. The objective of this PR is to fix all the failing tests.

Moving from Gym 0.26.2 to gymanisum 0.26.2 (which are the same) is part of this PR.

Missing:
- [ ] update changelog
- [x] update notebooks and check it works with gym 0.26 (https://github.com/Stable-Baselines-Team/rl-colab-notebooks/pull/9)
- [x] update RL Zoo and check it works with gym 0.26 (https://github.com/DLR-RM/rl-baselines3-zoo/pull/256, not ready, gym needs to be fixed first)
- [x] update tutorials (https://github.com/araffin/rl-tutorial-jnrr19/pull/13)
- [ ] prepare docker image
- [x] ~push docker images~
- [x] update SB3 contrib (https://github.com/Stable-Baselines-Team/stable-baselines3-contrib/pull/71)
- [x] Check test environments #1156
- [x] Use shimmy to automatically convert gym envs, see https://github.com/thu-ml/tianshou/pull/789
(note: waiting for https://github.com/Farama-Foundation/Shimmy/pull/29 and proper decoupling with gymnasium)

## Motivation and Context
[Gym release notes](https://github.com/openai/gym/releases/tag/0.22.0)

## Types of changes

closes #840 #871 
closes #271 
closes #1156 

## Checklist:

### Status
- There's still 28 failing tests when running locally, need to investigate further


araffin:
it seems that the failures come from `pygame` not being installed somehow, probably an issue from gym...

carlosluis:
> it seems that the failures come from `pygame` not being installed somehow, probably an issue from gym...

Only a subset of the failures, those are fixed now.

carlosluis:
@RedTachyon I'm currently investigating a bug which may be related to https://github.com/openai/gym/pull/2422.

Here's a stack trace of the type of errors:
```
tests/test_save_load.py:216: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
stable_baselines3/dqn/dqn.py:258: in learn
    return super(DQN, self).learn(
stable_baselines3/common/off_policy_algorithm.py:354: in learn
    rollout = self.collect_rollouts(
stable_baselines3/common/off_policy_algorithm.py:584: in collect_rollouts
    actions, buffer_actions = self._sample_action(learning_starts, action_noise, env.num_envs)
stable_baselines3/common/off_policy_algorithm.py:415: in _sample_action
    unscaled_action, _ = self.predict(self._last_obs, deterministic=False)
stable_baselines3/dqn/dqn.py:238: in predict
    action = np.array([self.action_space.sample() for _ in range(n_batch)])
stable_baselines3/dqn/dqn.py:238: in <listcomp>
    action = np.array([self.action_space.sample() for _ in range(n_batch)])
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = Discrete(10)

    def sample(self) -> int:
>       return self.start + self.np_random.randint(self.n)
E       AttributeError: 'numpy.random._generator.Generator' object has no attribute 'randint'
```
I've been digging into the code and the problem seems to be that [this line](https://github.com/openai/gym/blob/3d2146d54cebe062ea38744697a968bf31a1b162/gym/spaces/discrete.py#L28) 
```python
def sample(self) -> int:
        return self.start + self.np_random.randint(self.n)
```
is being called with `self.np_random` of type `numpy.random._generator.Generator`, instead of `gym.utils.seeding.RandomNumberGenerator`. The former doesn't have a method `randint` so the code fails. 

If it's any help, this is only happening after loading a model, so the problem may be related to it (I haven't checked much on that front yet, wanted to rule out something obvious with the changes on seeding first). Here's an example test were it fails https://github.com/DLR-RM/stable-baselines3/blob/58a98060f9ec618a740f121c5904ee845fd02f15/tests/test_save_load.py#L216

It seems `discrete.py` in Gym assumes np_random will always be of some type but doesn't enforce it explicitly?  Or there's something funny happening with the loading...

RedTachyon:
I think I found the issue - the custom RNG class inherits from the numpy Generator for compatibility purposes, but when it gets pickled/unpickled, it defaults to the numpy behavior which creates a new Generator. 

I think I see two main ways to solve it - either in `__setstate__` of Space make sure that a RNG is set instead of a Generator; or in the RNG code do some custom code to pickle/unpickle the proper class

carlosluis:
> I think I found the issue - the custom RNG class inherits from the numpy Generator for compatibility purposes, but when it gets pickled/unpickled, it defaults to the numpy behavior which creates a new Generator.
> 
> I think I see two main ways to solve it - either in `__setstate__` of Space make sure that a RNG is set instead of a Generator; or in the RNG code do some custom code to pickle/unpickle the proper class

I see, thanks for checking. I wonder what this means for SB3 adopting Gym 0.22, since the problem seems to be on Gym's side. I'm guessing the fix (whenever it comes) won't be available for SB3 until the next release by Gym?

araffin:
>  won't be available for SB3 until the next release by Gym?

https://github.com/openai/gym/issues/2640 must be fixed first too

jkterry1:
We're going to do a 0.22.1 release in the new future that will fix all of these things, some of the relevant fixes here have already been merged

stable_baselines3/common/envs/bit_flipping_env.py
```diff
@@ -2,13 +2,13 @@
...
 import numpy as np
-from gym import GoalEnv, spaces
+from gym import Env, spaces
 from gym.envs.registration import EnvSpec
 
 from stable_baselines3.common.type_aliases import GymStepReturn
 
 
-class BitFlippingEnv(GoalEnv):
+class BitFlippingEnv(Env):
```
araffin, line 11: we should rather use `GoalEnv` from gym rotobics once https://github.com/Farama-Foundation/gym-robotics/issues/7 is solved, no?

jkterry1: My intention was/is to remove the GoalEnv class from Gym in favor of the regular base class. We just kept it in gym-robotics because we didn't want to bother removing it. I would discourage other projects from using it.

carlosluis:
Regarding the "mean reward below threshold" problem, after some further investigations the root cause is the change in seeding behaviour in gym.

SB3 sets seeds here:
https://github.com/DLR-RM/stable-baselines3/blob/e88eb1c9ca98650f802409e5845e952c39be9e76/stable_baselines3/common/base_class.py#L576

Before gym 0.22.0, the default `seed()` method was not actually setting any seed [[source](https://github.com/openai/gym/blob/v0.21.0/gym/core.py#L124)]. In gym 0.22 the default `seed()` does set the seed [[source](https://github.com/openai/gym/blob/2dddaf722acccfd0412d745890c40dcd972586d5/gym/core.py#L157)].

The consequence of it is, for instance, that the result of 
https://github.com/DLR-RM/stable-baselines3/blob/e88eb1c9ca98650f802409e5845e952c39be9e76/stable_baselines3/common/evaluation.py#L82
is not consistent between gym==0.21.0 and gym >= 0.22.0. The solution I say makes the most sense is to simply adjust the threshold value in the test.

araffin:
> Before gym 0.22.0, the default seed() method was not actually setting any seed [[source](https://github.com/openai/gym/blob/v0.21.0/gym/core.py#L124)]. In gym 0.22 the default seed() does set the seed [[source](https://github.com/openai/gym/blob/2dddaf722acccfd0412d745890c40dcd972586d5/gym/core.py#L157)].

I see...

> is not consistent between gym==0.21.0 and gym >= 0.22.0.

is there any way to make it consistent?

> The solution I say makes the most sense is to simply adjust the threshold value in the test.

The whole point of those performance tests is to detect any change that may injure performance/change results. (they fail easily as soon as any minor change is made) So we should not change the threshold but rather fix the underlying change/issue, or in the current case, I would prefer changing the seed if we cannot have consistent behavior with previous version.


btw, are all those warnings (see below) due to gym only?
```
gym/utils/seeding.py:47: 1 warning
tests/test_callbacks.py: 4153 warnings
tests/test_cnn.py: 377 warnings
tests/test_custom_policy.py: 106 warnings
tests/test_deterministic.py: 238 warnings
tests/test_dict_env.py: 14116 warnings
tests/test_envs.py: 2 warnings
tests/test_her.py: 1566 warnings
tests/test_identity.py: 18153 warnings
tests/test_logger.py: 300 warnings
tests/test_monitor.py: 3000 warnings
tests/test_predict.py: 109 warnings
tests/test_run.py: 222 warnings
tests/test_save_load.py: 6597 warnings
tests/test_spaces.py: 1110 warnings
tests/test_train_eval_mode.py: 31 warnings
tests/test_utils.py: 12 warnings
tests/test_vec_envs.py: 216 warnings
tests/test_vec_monitor.py: 4000 warnings
tests/test_vec_normalize.py: 6547 warnings
  /opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/gym/utils/seeding.py:47: DeprecationWarning: WARN: Function `rng.randint(low, [high, size, dtype])` is marked as deprecated and will be removed in the future. Please use `rng.integers(low, [high, size, dtype])` instead.
```

same for 

```
tests/test_envs.py::test_non_default_action_spaces[new_action_space4]
  /home/runner/work/stable-baselines3/stable-baselines3/stable_baselines3/common/env_checker.py:272: UserWarning: We recommend you to use a symmetric and normalized Box action space (range=[-1, 1]) cf https://stable-baselines3.readthedocs.io/en/master/guide/rl_tips.html
    warnings.warn(

tests/test_utils.py: 16 warnings
  /opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/gym/utils/seeding.py:138: DeprecationWarning: WARN: Function `hash_seed(seed, max_bytes)` is marked as deprecated and will be removed in the future. 
    deprecation(

tests/test_utils.py: 16 warnings
  /opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/gym/utils/seeding.py:175: DeprecationWarning: WARN: Function `_bigint_from_bytes(bytes)` is marked as deprecated and will be removed in the future. 
    deprecation(
```
?

carlosluis:
Similarly, the HER test failure was also caused by the change in RNG. The failing test required that after training for 200 steps, the env had to be left in the middle of a trajectory. After the change in gym, the RNG gods decided that after 200 steps the env had just finished an episode, i.e., `self.current_idx == 0` in

https://github.com/DLR-RM/stable-baselines3/blob/e88eb1c9ca98650f802409e5845e952c39be9e76/stable_baselines3/her/her_replay_buffer.py#L523

==> warning is never raised ==> test assertion fails.

A simple change in the seed fixes the test.

carlosluis:
@araffin I would have to check the warnings, but now every time env.seed is call will raise a warning.

I reverted the threshold back to 90 and changing the seed did the trick.

araffin:
>  but now every time env.seed is called will raise a warning.

Could we make the `VecEnv` seed method backward compatible? In the sense that if `seed` is accepted in `reset()` it calls `reset(seed=seed)`, otherwise it call `seed(seed)` when `venv.seed(seed)` is called.
This would prevent most warnings normally.

setup.py
```diff
@@ -73,7 +73,7 @@
     packages=[package for package in find_packages() if package.startswith("stable_baselines3")],
     package_data={"stable_baselines3": ["py.typed", "version.txt"]},
     install_requires=[
-        "gym==0.21",  # Fixed version due to breaking changes in 0.22
+        "gym>=0.21",
```
araffin, line 76: I would fix gym version again, unless we are sure no breaking changes is coming afterward...

RedTachyon:
@carlosluis 
>Regarding the "mean reward below threshold" problem, after some further investigations the root cause is the change in seeding behaviour in gym.

Can you elaborate on how you got to this conclusion? Something seems very weird here, since this would imply that the test used to be stochastic (and happened to always pass), and now is deterministic, but the seed happens to fail. What's more, the change you mention affects `env.np_random`, which from what I can tell isn't used in the actual [test](https://github.com/DLR-RM/stable-baselines3/blob/master/stable_baselines3/common/envs/identity_env.py#L49)

setup.py
jkterry1: There will be breaking API changes until 1.0, probably in a few more months. Any releases after will still likely result in some breaking changes, e.g. environment version bumps and so on.

RedTachyon:
The difference is probably that the underlying RNG mechanism changed somewhere between releases, so the effect of setting an old random seed won't be the same.

I'd be super uncomfy having a test that depends so heavily on the random seed, but if that's just to check for consistency, searching for a random seed that works again would probably be good enough.

carlosluis:
> @carlosluis i think there was a misunderstanding with my remark here: [#780 (comment)](https://github.com/DLR-RM/stable-baselines3/pull/780#issuecomment-1062235112)
> 
> please do not change `VecEnv` signature but make `VecEnv` `seed()` method forward compatible (so it supports all gym versions).

@araffin I see, yes I understood something different. AFAIK, to make `VecEnv` `seed()` forward compatible we should internally call `reset(seed=seed)` for every env, similar to how it's now done in `DummyVecEnv`.
```python
def seed(self, seed: Optional[int] = None) -> List[Union[None, int]]:
        seeds = list()
        for idx, env in enumerate(self.envs):
              seeds.append(env.reset(seed=seed + idx))
        return seeds
```
but I'm not sure how to do that from the base class `VecEnv`, it doesn't have access to the individual environments right?

araffin:
> AFAIK, to make VecEnv seed() forward compatible we should internally call reset(seed=seed) for every env, similar to how it's now done in DummyVecEnv.

yes

> but I'm not sure how to do that from the base class VecEnv

you don't do it in the base class, but rather in `DummyVecEnv` and in `SubprocVecEnv` (in the worker) and that should do the trick ;)

carlosluis:
> > AFAIK, to make VecEnv seed() forward compatible we should internally call reset(seed=seed) for every env, similar to how it's now done in DummyVecEnv.
> 
> yes
> 
> > but I'm not sure how to do that from the base class VecEnv
> 
> you don't do it in the base class, but rather in `DummyVecEnv` and in `SubprocVecEnv` (in the worker) and that should do the trick ;)

Got it! I already have that mostly, I will remove the signature change in the base class and also make the `SubprocVecEnv` worker change.

carlosluis:
@araffin As I was going through the commit diff to document the changes, I realized that maybe removing the `GoalEnv` class is incompatible with HER? If that's the case, then we would have to do as you said before and bring the class from https://github.com/Farama-Foundation/Gym-Robotics

In the docs for HER we have:

```
.. warning::

    HER requires the environment to inherits from `gym.GoalEnv <https://github.com/openai/gym/blob/3394e245727c1ae6851b504a50ba77c73cd4c65b/gym/core.py#L160>`_
```

I don't find where specifically such a warning is raised, but there's definitely a codepath that expects to find the `compute_reward()` method enforced by the `GoalEnv` interface.

https://github.com/DLR-RM/stable-baselines3/blob/30772aa9f53a4cf61571ee90046cdc454c1b11d7/stable_baselines3/her/her_replay_buffer.py#L346-L362

This also impacts the HER colab example in https://github.com/DLR-RM/stable-baselines3/blob/master/docs/guide/examples.rst

Lastly, the [highway env repo](https://github.com/eleurent/highway-env) went through a similar process to start using GoalEnv from gym robotics but ultimately they dropped the dependency and copied the class instead, see https://github.com/eleurent/highway-env/commit/f8a4c633012ee004786346908e3f9b841edd64aa. Maybe we need to keep that in mind when discussing what to do about this issue.

araffin:
> I don't find where specifically such a warning is raised, but there's definitely a codepath that expects to find the compute_reward() method enforced by the GoalEnv interface.

well, the two features that we use are:
- we expect a dict obs with three keys `observation`, `achieved_goal` and `desired_goal`
- we expect a vectorized implementation of `compute_reward()`

carlosluis:
> > I don't find where specifically such a warning is raised, but there's definitely a codepath that expects to find the compute_reward() method enforced by the GoalEnv interface.
> 
> well, the two features that we use are:
> 
>     * we expect a dict obs with three keys `observation`, `achieved_goal` and `desired_goal`
> 
>     * we expect a vectorized implementation of `compute_reward()`

I wanted to check if there's something to be done in SB3 regarding the removal of GoalEnv from gym, in which case we should probably address it in this PR.

araffin:
> I wanted to check if there's something to be done in SB3 regarding the removal of GoalEnv from gym, in which case we should probably address it in this PR.

Those checks are now missing:
https://github.com/openai/gym/blob/v0.21.0/gym/core.py#L182
and need to be added to the env checker at least (maybe somewhere else if it's easy to do).

And the doc must be updated to explain what is expected for `HER` replay buffer (because we cannot refer to `GoalEnv` interface anymore).

JesseFarebro:
@araffin the ALE related warnings about `seeding.hash_seed` and `rng.randint` are now fixed in `ale-py==0.7.5`.

arjun-kg:
@araffin I would like to help getting this PR merged. Are you or anyone else working on the missing points? If not, could I work on PRs to the connected repos (zoo, notebooks etc.)?

araffin:
> @araffin I would like to help getting this PR merged. Are you or anyone else working on the missing points? If not, could I work on PRs to the connected repos (zoo, notebooks etc.)?

thanks for the help =)
currently no one is working on the missing points, please go ahead!

arjun-kg:
@araffin regarding the checklist, 
1. Tutorials, Notebooks - Created PRs. 
2. Zoo - Tests are working with sb3 at this fork, and sb3-contrib at current master (however sb3 contrib tests fail due to using gym.GoalEnv, but that does not affect zoo, maybe I can separately work on fixing that). Only need to change gym version in requirements and changelog which can be done along with next SB3 release. 

I think that's all I can do outside this PR. Please let me know if this is okay.

stable_baselines3/common/env_checker.py
```diff
@@ -93,6 +93,27 @@ def _check_nan(env: gym.Env) -> None:
         _, _, _, _ = vec_env.step(action)
 
 
+def _is_goal_env(env: gym.Env) -> bool:
```
araffin, line 96: note for myself: add a check for the compute reward

araffin:
> 2\. Zoo - Tests are working with sb3 at this fork,

what test did you run for the zoo?

by running `make pytest` with this branch, I already get at least one test failing.

arjun-kg:
> what test did you run for the zoo?

I had run `make pytest` and everything had passed for me. I pulled recent commits and tried again now, everything was still passing. 

I had missed `make check-trained-agents`. I ran this now. This had errors due to - PandaEnvs (because of removal of `gym.GoalEnv` and panda-gym is not updated to reflect this). I'll create a PR there.

qgallouedec:
> `panda-gym` master has updated the changes to `GoalEnv`.

I'm the author of `panda-gym`, if I can help, let me know. 
FYI:
panda-gym v2.0.1 (current version) works with gym 0.23 (current version, without `GoalEnv`)
panda-gym v1.1.1 works with gym 0.21 (old version, with `GoalEnv`)

arjun-kg:
@qgallouedec thank you for the update :) 

@araffin I think if the PR mentioned above is merged, and `panda-gym` version is  updated to v2.0.1  (after SB3 is updated to gym 0.23.1), all the tests should pass.

araffin:
Note: tests are working with gym 0.24 (as expected, because this is only a bug fix release) but I had to fix `step:12:Definition list ends without a blank line; unexpected unindent.` when building the doc with sphinx (the default docstring of gym step is invalid rst )

araffin:
I'm not sure why the doc build fails... I tried to reproduce locally but couldn't, any help is welcomed ;)

cloderic:
Is there an ETA for this? >0.22 gym version migrated a bunch of environment to PyGame which is way better for headless rendering (rendering to a picture) which we are using in https://github.com/cogment/cogment-verse a ton.

Is there a way to help to finalize this PR ?

araffin:
> Is there an ETA for this

@cloderic as @RedTachyon mentioned, the ETA depends on when Gym will be stable again.

but you can already use this PR (which will get updates) if you want to use SB3 with latest Gym version:

```
pip install git+https://github.com/carlosluis/stable-baselines3@fix_tests
```

We wanted actually to add this information in our documentation.
(we also have a branch in SB3 contrib, see PR description at the top of the page)

> Is there a way to help to finalize this PR ?

Help is actually welcomed on the PR for the zoo: https://github.com/DLR-RM/rl-baselines3-zoo/pull/256
and this PR will anyway receive updates when new Gym versions are released.

cloderic:
Thanks!

> Help is actually welcomed on the PR for the zoo: [DLR-RM/rl-baselines3-zoo#256](https://github.com/DLR-RM/rl-baselines3-zoo/pull/256)
> and this PR will anyway receive updates when new Gym versions are released.

I'll look into it when I have a chance!

ChengYen-Tang:
Hi, I have a question.
I found in the gym code that reset must return observation and info.
However, stable-baselines seem to only accept observation.
https://github.com/openai/gym/blob/master/gym/core.py#L148

araffin:
Hello,

>  I found in the gym code that reset method must return observation and info.

yes, we are aware of it, but it will be mandatory only from gym > 0.26 (released soon).

Talking about that, @carlosluis could you give me rights to your fork (maintainer right)? it will make the update for that branch easier (if we receive external contributions for instance).

carlosluis:
> Hello,
> 
> > I found in the gym code that reset method must return observation and info.
> 
> yes, we are aware of it, but it will be mandatory only from gym >= 0.26 (released soon).
> 
> Talking about that, @carlosluis could you give me rights to your fork (maintainer right)? it will make the update for that branch easier (if we receive external contributions for instance).

Done, I sent an invite

pseudo-rnd-thoughts:
@carlosluis Hey, Gym version 26 has just been released and hopefully in PyPI in the next few days. 
Do you plan on making the updates here or in another PR? Im happy to help make some of the upgrades if you want

carlosluis:
> @carlosluis Hey, Gym version 26 has just been released and hopefully in PyPI in the next few days. Do you plan on making the updates here or in another PR? Im happy to help make some of the upgrades if you want

Hey, I'm not actively working on this. You may want to check with @araffin if there's anything you can help with to get this merged.

tlpss:
> > @carlosluis Hey, Gym version 26 has just been released and hopefully in PyPI in the next few days. Do you plan on making the updates here or in another PR? Im happy to help make some of the upgrades if you want
> 
> Hey, I'm not actively working on this. You may want to check with @araffin if there's anything you can help with to get this merged.

@araffin I'm also very interested in this PR and would be willing to assist if you need some help.

pseudo-rnd-thoughts:
@tlpss Would you be able to contact me on the gym discord server (https://discord.gg/nHg2JRN489) and we can coordinate updating together? My name is PseudoRnd

araffin:
fyi @carlosluis gave me the rights to his fork, so best would be to do PR to his fork so we keep the same PR for latest gym version support.

tlpss:
@araffin (cc @pseudo-rnd-thoughts) I was looking at this PR.

I switched to v26 on my fork, after which tests started failing ofc. 

I think this is a list of  changes that need to happen:
- All custom Gym Classes in the repo should be updated to the v26  API meaning they should 
1. return obs, info in reset
2. take seed as a reset argument
3. return a 5 tuple in the step function

- The Replay buffer needs to handle the new Truncated/terminated (done is mapped to terminated, truncated is new)
- The DummyVecEnv should handle the new API

- Later on, the algorithms should effectively start to use the terminated vs truncated signal, but that is beyond the scope of this PR i believe.

This list is probably not complete (my knowledge of SB3 is limited), but I believe it is a good start. 

One question that I have is: do we want to keep the `done` for readability throughout the repo? Or are we gonna map `done` to `terminated` everywhere?

ChengYen-Tang:
Hello everyone,
I have a question about Gym V0.26, are termination and truncation equivalent to the previous is_done? One of the options is true, it will reset the environment.

araffin:
@tlpss thanks for working on this, I made some comments directly on your PR.
I think it will be easier and better for the use to only add compatibility in the `VecEnv` (as everything is assumed to be a vec env in SB3).

@ChengYen-Tang gym issues is a better place for such questions, there will be a blog post explaining it (I agree the change is confusing): https://github.com/pseudo-rnd-thoughts/gym/issues/1

tlpss:
@araffin, I get how you would do this for the step method, but how do you see this for the reset that now includes an optional seed? 

Furthermore it seems to me that this is quite a small change and I would expect most gym environments to adapt quickly to the new API? I get the importance of backwards compatibility, but I feel like it would create more work in the future?

araffin:
> I get how you would do this for the step method, but how do you see this for the reset that now includes an optional seed?

Replying to you directly on https://github.com/carlosluis/stable-baselines3/pull/1

> Furthermore it seems to me that this is quite a small change and I would expect most gym environments to adapt quickly to the new API? I get the importance of backwards compatibility, but I feel like it would create more work in the future?

not so small if you have to adapt all wrappers and other glue code.

> I get the importance of backwards compatibility, but I feel like it would create more work in the future?

I prefer to ensure smooth transition for our users. And at the end, the plan would be to change only the `VecEnv` internal, not its API (I'm afraid we would have to change too much things internally otherwise).

stable_baselines3/common/utils.py
```diff
@@ -535,3 +545,66 @@ def compat_gym_seed(env: GymEnv, seed: int) -> None:
...
+) -> Tuple[GymObs, Dict]:
+    """
+    Transform gym reset return to new API (gym > 0.26)
+
+    :param reset_return: (obs) or (obs, info)
+    :param warn: Whether to warn or not the user
+    :return: (obs, info)
+    """
+    if isinstance(reset_return, tuple):
+        assert len(reset_return), (
```
tlpss: should be `assert len(reset_return) == 2`?

tests/test_vec_envs.py
```diff
@@ -31,14 +31,14 @@ def reset(self, seed: Optional[int] = None):
...
-        return self.state
+        return self.state, {}
 
     def step(self, action):
         reward = float(np.random.rand())
         self._choose_next_state()
         self.current_step += 1
-        done = self.current_step >= self.ep_length
-        return self.state, reward, done, {}
+        done = truncated = self.current_step >= self.ep_length
```
tlpss, line 40: If i'm not mistaken, in the gymv0.26 API done would be false here as this is not a terminal state, and truncation would be true. (setting only truncation to true would also allow for testing of correct handling of all cases of done and truncation)

tests/test_vec_envs.py
```diff
@@ -147,13 +147,13 @@ def __init__(self, max_steps):
...
         self.current_step = 0
-        return np.array([self.current_step], dtype="int")
+        return np.array([self.current_step], dtype="int"), {}
 
     def step(self, action):
         prev_step = self.current_step
         self.current_step += 1
-        done = self.current_step >= self.max_steps
-        return np.array([prev_step], dtype="int"), 0.0, done, {}
+        done = truncated = self.current_step >= self.max_steps
```
tlpss, line 155: same here

tests/test_vec_normalize.py
```diff
@@ -35,13 +35,14 @@ def step(self, action):
         self.t += 1
         index = (self.t + self.return_reward_idx) % len(self.returned_rewards)
         returned_value = self.returned_rewards[index]
-        return np.array([returned_value]), returned_value, self.t == len(self.returned_rewards), {}
+        done = truncated = self.t == len(self.returned_rewards)
```
tlpss, line 38: same, we might want to test the influence of truncations on the statistics separetely?

tests/test_gae.py
```diff
@@ -25,18 +25,22 @@ def reset(self, seed: Optional[int] = None):
...
     def step(self, action):
         self.n_steps += 1
 
-        done = False
+        done = truncated = False
         reward = 0.0
         if self.n_steps >= self.max_steps:
             reward = 1.0
             done = True
+            # To simplify GAE computation checks,
```
tlpss, line 38: but does the GAE calculation distinguish between truncation and termination? from here https://github.com/tlpss/stable-baselines3/blob/95bb4d6b880f21e8e3117d29d4391937f12fedbf/tests/test_gae.py#L96 it seems not? (fixing this might be not in the scope of this PR, but I feel the GAE tests should cover this now that the truncation is  added)

stable_baselines3/common/utils.py
araffin: it should be yes, but the code is not longer used.

tests/test_vec_envs.py
araffin: See my comments in https://github.com/openai/gym/issues/3102#issuecomment-1264648469

My current interpretation of it is made to match gym 0.21 semantic: 
- terminated=True -> reset
- truncated=True -> reset (included in terminated) but bootstrap is valid

> also allow for testing of correct handling of all cases of done and truncation)

we do that in test_gae

tests/test_vec_normalize.py
araffin: I need to think more about that one, but my guess is that it should not matter much, the scaling is meant to avoid large values and to evolve with the agent performance. Still, we should double check that.
Could you open an issue so we don't forget?

tests/test_gae.py
araffin: See test below that one, we handle truncations like that: #658 
It is a trick that allows to keep the same code when truncation happens.

docs/misc/changelog.rst
```diff
@@ -65,6 +65,7 @@ Release 1.6.0 (2022-07-11)
 
 Breaking Changes:
 ^^^^^^^^^^^^^^^^^
+- Switched minimum Gym version to 0.24 (@carlosluis)
```
qgallouedec, line 68: 0.26

docs/modules/her.rst
```diff
@@ -22,7 +22,10 @@ It creates "virtual" transitions by relabeling transitions (changing the desired
 
 .. warning::
 
-    HER requires the environment to inherits from `gym.GoalEnv <https://github.com/openai/gym/blob/3394e245727c1ae6851b504a50ba77c73cd4c65b/gym/core.py#L160>`_
+    HER requires the environment to follow the legacy `gym.GoalEnv interface <https://github.com/openai/gym/blob/3394e245727c1ae6851b504a50ba77c73cd4c65b/gym/core.py#L160>`_
```
qgallouedec, line 25: `GoalEnv` is now part of `gym-robotics`. I'd replace by 
```rst
[...] legacy `gym_robotics.GoalEnv interface <https://github.com/Farama-Foundation/Gymnasium-Robotics/blob/a35b1c1fa669428bf640a2c7101e66eb1627ac3a/gym_robotics/core.py#L8>`_
```

tlpss:
> > One question that I have is: do we want to keep the `done` for readability throughout the repo? Or do we want to refactor `done` to `terminated` everywhere?
> 
> I think we must rename `done` to `terminated`
> 
> > **edit** I started working on this [here](https://github.com/carlosluis/stable-baselines3/pull/1) and kept the `done` name for the `terminated` value for now to reduce the amount of changes at a time.
> 
> Ok, but IMO, at the end (before merging this PR), these changes should be included.

@araffin it seems like there are only some minor changes left before this PR is ready to merge? Would you be able to make those changes? I could offer some help if required.

stable_baselines3/common/atari_wrappers.py
```diff
@@ -103,12 +106,12 @@ def reset(self, **kwargs) -> np.ndarray:
         :return: the first observation of the environment
         """
         if self.was_real_done:
-            obs = self.env.reset(**kwargs)
+            obs, info = self.env.reset(**kwargs)
         else:
             # no-op step to advance from terminal/lost life state
-            obs, _, _, _ = self.env.step(0)
+            obs, _, _, info = self.env.step(0)
```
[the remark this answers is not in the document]
araffin: thanks, nice catch

araffin:
> @araffin it seems like there are only some minor changes left before this PR is ready to merge? Would you be able to make those changes? I could offer some help if required.

Well, the main changes have been made but there is quite some work to do and we still need to wait a bit to check everything is working as expected (for instance, I found out some inconsistency: https://github.com/openai/gym/issues/3102).

I would actually need help to upgrade the notebooks and tutorials, see checklist: https://github.com/DLR-RM/stable-baselines3/pull/780
(and then later on SB3 contrib and the RL Zoo)
There are already pull request by @arjun-kg  for gym 0.22+ and you can build upon it.

tlpss:
> I would actually need help to upgrade the notebooks and tutorials, see checklist: https://github.com/DLR-RM/stable-baselines3/pull/780

I can offer some help with these changes. It might be convenient to have have access to the repo of this PR though, or do you prefer me to create a PR to that repo whenever I make some changes?

araffin:
> It might be convenient to have have access to the repo of this PR though

@arjun-kg could you give access to @tlpss and me to your forks?

Rocamonde:
Let me know if there is something concrete that I can help with to bring this PR to the finish line.

araffin:
> Let me know if there is something concrete that I can help with to bring this PR to the finish line.

yes, I would happy to receive help for the RL Zoo and the tutorials (see checklist).
For the tutorials/notebooks, @tlpss is already working on that normally.

Rocamonde:
Also https://github.com/carlosluis/stable-baselines3@fix_tests seems to not really be compatible with Gym 0.26? In particular, https://github.com/carlosluis/stable-baselines3/blob/cf565d9f1e19efddc39a348f1a9ab7f7d2d1e0fb/stable_baselines3/common/on_policy_algorithm.py#L175 still uses the old `.step()` API. This is within the `.learn()` method for OnPolicyAlgorithm, which seems to me like a core element of sb3.

araffin:
> Also https://github.com/carlosluis/stable-baselines3@fix_tests seems to not really be compatible with Gym 0.26?

it is, SB3 rely on VecEnv, and to minimize potential bugs and breaking changes, I kept the same API (https://github.com/carlosluis/stable-baselines3/pull/1#issuecomment-1264295531) but we accept gym 0.26 only for the underlying env.

tlpss:
@araffin I noticed that gym has an [env checker](https://github.com/openai/gym/blob/a368cfaaed0b79836f9e5af3c0a9896ef7f7d6ea/gym/utils/passive_env_checker.py#L268) and SB3 has a [checker](https://github.com/carlosluis/stable-baselines3/blob/ed191b073ceaebd5b9b19bbf8030f12ff6058bfa/stable_baselines3/common/env_checker.py#L288). 

But for example the render_checker  of SB3 is more limited than that of Gym and also still includes the old `mode` argument for the render  method. Is this something that needs to be addressed in this PR?

araffin:
> I noticed that gym has an [env checker](https://github.com/openai/gym/blob/a368cfaaed0b79836f9e5af3c0a9896ef7f7d6ea/gym/utils/passive_env_checker.py#L268) and SB3 has a [checker](https://github.com/carlosluis/stable-baselines3/blob/ed191b073ceaebd5b9b19bbf8030f12ff6058bfa/stable_baselines3/common/env_checker.py#L288).

yes (see https://github.com/DLR-RM/stable-baselines3/issues/813), the gym env checker originated from SB3, but we keep our internal ones as we have specific warnings.

> But for example the render_checker of SB3 is more limited than that of Gym and also still includes the old mode argument for the render method. Is this something that needs to be addressed in this PR?

you mean that we check less things for render with our env checker?
yes, it should be at least updated to match the current API.

tlpss:
> yes (see #813), the gym env checker originated from SB3, but we keep our internal ones as we have specific warnings.

I noticed

> you mean that we check less things for render with our env checker? yes, it should be at least updated to match the current API.

Yes that's what I meant indeed, I noticed it while updating the tutorials. I forgot to update the rendering and the env still passed the checks. So we should probably update the environment checker to match the API (and especially give warnings w.r.t. the old API)

araffin:
> (and especially give warnings w.r.t. the old API)

I would prefer to have a clear migration guide in Gym doc (currently not the case, as I pointed out in https://github.com/openai/gym/issues/3102)

tlpss:
> I would prefer to have a clear migration guide in Gym doc (currently not the case, as I pointed out in [openai/gym#3102](https://github.com/openai/gym/issues/3102))

Yeah I'm missing that too. Now I find myself searching in the release notes, which is suboptimal imo.

araffin:
> Yet, the videorecorder uses the gym videorecorder, which uses the new API so atm I cannot record videos.

I'm not sure what and how things have changed (gym 0.26 was released first, doc comes later :'(), I would expect we could still use gym recorder internally with minimal changes?

tlpss:
> > Yet, the videorecorder uses the gym videorecorder, which uses the new API so atm I cannot record videos.
> 
> I'm not sure what and how things have changed (gym 0.26 was released first, doc comes later :'(), I would expect we could still use gym recorder internally with minimal changes?

Most notable change: the `mode` argument has disappeared and an atrribute `render_mode` has been introduced with the same purpose (but that is supposed to stay unchanged after instantiating the gym instance I believe). Furthermore the 'human' render mode (if enabled), should render in the step method.  See release notes or [blog post](https://younis.dev/blog/render-api/)

```
Rendering - It is normal to only use a single render mode and to help open and close the rendering window,
 we have changed Env.render to not take any arguments and so all render arguments
 can be part of the environment's constructor i.e., gym.make("CartPole-v1", render_mode="human").
 For more detail on the new API, see [blog post](https://younis.dev/blog/render-api/) @younik
```

Rocamonde:
Bug report on this experimental branch:

```
  File "/venv/lib/python3.9/site-packages/stable_baselines3/common/vec_env/dummy_vec_env.py", line 99, in render
    return super().render(mode=mode)
  File "/venv/lib/python3.9/site-packages/stable_baselines3/common/vec_env/base_vec_env.py", line 178, in render
    imgs = self.get_images()
  File "/venv/lib/python3.9/site-packages/stable_baselines3/common/vec_env/dummy_vec_env.py", line 82, in get_images
    return [env.render(mode="rgb_array") for env in self.envs]
  File "/venv/lib/python3.9/site-packages/stable_baselines3/common/vec_env/dummy_vec_env.py", line 82, in <listcomp>
    return [env.render(mode="rgb_array") for env in self.envs]
  File "/venv/lib/python3.9/site-packages/gym/core.py", line 329, in render
    return self.env.render(*args, **kwargs)
  [... suppressed]
TypeError: render() got an unexpected keyword argument 'mode'
```

Rocamonde:
Hmm actually the above requires more changes than I thought, even though that will probably remove a bunch of errors. `render()` on DummyVecEnv and `_check_render` need to be changed, as well as some other files:

Screen Shot 2022-11-02 at 2 59 15 PM

qgallouedec:
render mode is now defined when making the env: https://gymnasium.farama.org/content/migration-guide/#environment-render

Rocamonde:
I think I fixed most of the issues in https://github.com/AlignmentResearch/stable-baselines3. It requires modifying the API. The only solution that's a bit of a dirty trick is the subprocess venv, which I have not tested but probably works.

tlpss:
> Bug report on this experimental branch:
> 
> ```
>   File "/venv/lib/python3.9/site-packages/stable_baselines3/common/vec_env/dummy_vec_env.py", line 99, in render
>     return super().render(mode=mode)
>   File "/venv/lib/python3.9/site-packages/stable_baselines3/common/vec_env/base_vec_env.py", line 178, in render
>     imgs = self.get_images()
>   File "/venv/lib/python3.9/site-packages/stable_baselines3/common/vec_env/dummy_vec_env.py", line 82, in get_images
>     return [env.render(mode="rgb_array") for env in self.envs]
>   File "/venv/lib/python3.9/site-packages/stable_baselines3/common/vec_env/dummy_vec_env.py", line 82, in <listcomp>
>     return [env.render(mode="rgb_array") for env in self.envs]
>   File "/venv/lib/python3.9/site-packages/gym/core.py", line 329, in render
>     return self.env.render(*args, **kwargs)
>   [... suppressed]
> TypeError: render() got an unexpected keyword argument 'mode'
> ```
> 
> Gym no longer takes mode on the render method. You initialize the mode when you create the environment. This should be easy to fix,just remove the kwarg in line 82 on `dummy_vec_env.py`

Thanks for the report, we are aware of this issue. Basically we need to decide what render api the vecenv will expose and fix all the local gym environments to adhere to the gym 0.26 API (set render mode during init)

I should find some time in the next days to implement the new render API and fix the Videorecorder as well.

Rocamonde:
@tlpss if you take a look at the repo I linked you can access a draft implementation which I am currently using for my research. It's working fine for me.

tlpss:
> I think I fixed most of the issues in https://github.com/AlignmentResearch/stable-baselines3. It requires modifying the API. The only solution that's a bit of a dirty trick is the subprocess venv, which I have not tested but probably works.

Ah I just missed this one. Thanks!

Rocamonde:
Once we're re-writing the render part of the code, can I ask if people think it's reasonable to request support for NOT tiling images in vectorized environments and instead returning a 4-dimensional array? Untiling the image is not a trivial thing to do (the tiling code has ~15 lines)

tlpss:
@araffin. Do we want the VecEnv to maintain the old step/reset API but adhere to the new render API? Or would you like to maintain the old render API somehow? I think we should go for the new render API, although this is probably a little bit more confusing.

tlpss:
> Re: the render method in venvs, can I ask if people think it's reasonable to request support for NOT tiling images in vectorized environments and instead returning a 4-dimensional array? Untiling the image is not a trivial thing to do (the tiling code has ~15 lines).
> 
> I will add support for this in my fork for now and allow passing a keyword argument, but it would be nice if this were kept as I would otherwise have to write an untiler when this PR is merged.

To me this seems like a valid request, although backward compatibility would have to be guarantueed probably (a flag in the vecenv could do the job, as you suggested). 
@araffin, what do you think?

Rocamonde:
Some gym environments do not allow changing the render mode along the way. So if the user passes a render mode that is not what the environment was initialized with, you should probably raise an error, not a deprecation warning, since this is indicative of a bug. But then at that point the sensible thing to do is to drop the argument and change the function signature, since this is how gym works now.

tlpss:
> Some gym environments do not allow changing the render mode along the way. So if the user passes a render mode that is not what the environment was initialized with, you should probably raise an error, not a deprecation warning, since this is indicative of a bug. 

Yes you're right, but that's why I suggested only giving a warning in case the argument matches the render mode that was used during initialization. 

>  But then at that point the sensible thing to do is to drop the argument and change the function signature, since this is how gym works now.

Yeah I see what you mean, but the decision to keep the API of the vecenv to the old gym API (for now) has been made already I'm afraid, so the question is if we want to adhere to this strictly (and hence still accept the older render signature) or if we should create an API that is a mix of the new and old gym API by having the render signature follow the new API. 

And then my 2 cents would be to reduce confusion by allowing both the old and new signature, but enforcing the new behavior (render mode should match the mode specified during initialization) & to throw a deprecation warning to make people already shift to the new API by not specifying the render mode in the  `render` method.

tlpss:
@araffin, I created a PR for the changes to the render API: https://github.com/carlosluis/stable-baselines3/pull/3
most important things to note:
- there is a new render mode `rgb_array_list` that is currently not Implemented as I did not known how to deal with it (and if you want to implement it)
- I renamed the `get_images` method to `get_render_output` as we now can no longer force the gyms to return images (the render mode has already be decided), unless you want to enforce `rgb_array` as only allowed render method (but I would argue against this, since it could be very useful to use the `human` mode during training). 
- If the `render_mode` is `rgb_array` the images are tiled and returned, otherwise the render  method returns None (now the render method also returned images in human mode).

Let me know what you think or if I have made mistakes.


@Rocamonde thanks for the inspiration!

AdamGleave:
@carlosluis are you looking for any help on this PR? Some members of my team might have time to put 2-3 days of work into it next week. Experienced devs but not experienced with Gym/SB per se.

rademacher-p:
Looking to use this PR with some `gymnasium` environments, noticed that `gym==0.26.2` is still being used throughout. My understanding is that `gymnasium` hasn't modified the API, such that simply changing import statements to `import gymnasium as gym` will suffice. Is this accurate? If so, can we get this change added? Happy to help, if needed! 👍

setup.cfg
```diff
@@ -13,11 +13,7 @@ filterwarnings =
 		# Tensorboard warnings
 		ignore::DeprecationWarning:tensorboard
 		# Gym warnings
-		ignore:Parameters to load are deprecated.:DeprecationWarning
```
araffin, line 16: note: remove or update to gymnasium later

stable_baselines3/common/vec_env/vec_video_recorder.py
```diff
@@ -48,6 +48,7 @@ def __init__(
             metadata = temp_env.metadata
 
         self.env.metadata = metadata
+        self.env.render_mode = "rgb_array"
```
araffin, line 51: TODO: remove once we support the new render api

tests/test_vec_monitor.py
```diff
@@ -132,8 +132,8 @@ def test_vec_monitor_ppo(recwarn):
     """
     Test the `VecMonitor` with PPO
     """
-    env = DummyVecEnv([lambda: gym.make("CartPole-v1")])
-    env.seed(0)
+    env = DummyVecEnv([lambda: gym.make("CartPole-v1", disable_env_checker=True)])
```
araffin, line 135: TODO: try to remove should be ok with the new gym version passive checker

Makefile
```diff
@@ -34,7 +40,8 @@ check-codestyle:
 commit-checks: format type lint
 
 doc:
-	cd docs && make html
+	# Prevent weird error due to protobuf
+	cd docs && PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=cpp make html
```
araffin, line 44: TODO: try to remove, the bug should be fixed now

pseudo-rnd-thoughts:
We have just released Gymnasium v0.27.0, this should be backward compatible. Would it be possible to update this Pr to v0.27 and check that nothing new breaks

araffin:
> We have just released Gymnasium v0.27.0, this should be backward compatible. Would it be possible to update this Pr to v0.27 and check that nothing new breaks

You can already check that using the hack shown in the PR description (and running `make pytest`, same for the RL Zoo but it has still some issues: https://github.com/DLR-RM/rl-baselines3-zoo/pull/256/). 

We are still updating and checking the new render api:  https://github.com/carlosluis/stable-baselines3/pull/3
In the future, it would be better to check first before releasing.

stable_baselines3/common/vec_env/vec_video_recorder.py
araffin: done

aegonwolf:
What version of gym does the current sb3 support? I.e. which version should I install so that I may use the documentation since otherwise all of it is out of date?

PaulMest:
FWIW, I've been using this with the latest version of gymnasium on LunarLander-v2 and it's been working great. Hope this gets merged soon. 🍀

jkterry1:
If you want to bump he minimum version here to be 0.27.0, it shouldn't create any new bugs or breaking changes and fixes an issues with making Atari environments that you actually care about

araffin:
Writing it down here so it is not forgotten, switching to gymnasium might create a bunch of hard to debug issue, 
see https://www.reddit.com/r/reinforcementlearning/comments/10djcbl/need_help_creating_an_action_space_compatible/ which is probably the same as https://github.com/DLR-RM/stable-baselines3/issues/1274
or https://discord.com/channels/765294874832273419/794148905889693718/1063432715875131462 (https://stackoverflow.com/questions/75108957/assertionerror-the-algorithm-only-supports-class-gym-spaces-box-box-as-acti)
For short, we have checks for supported obs/action space (`isinstance(...)`) that will fail when someone is using gym instead of gymnasium or the other way around.
We will also have issues of saved models (we check that the obs space is the same after loading).
I don't know what is the best way to handle it yet, maybe @pseudo-rnd-thoughts has an idea?
(one thing I see is monkey patching, pretty ugly but as it least it works: https://github.com/DLR-RM/rl-baselines3-zoo/blob/feat/gym-0.24/rl_zoo3/gym_patches.py#L52)

jkterry1:
@araffin theres what may be a simpler alternative: we have really really good gym compatibility wrappers built into gymnasium. You could just tell users to use those for gym environments as needed instead of having to handle all of this logic yourself

pseudo-rnd-thoughts:
> Writing it down here so it is not forgotten, switching to gymnasium might create a bunch of confusing issues, see https://www.reddit.com/r/reinforcementlearning/comments/10djcbl/need_help_creating_an_action_space_compatible/ which is probably the same as #1274 or https://discord.com/channels/765294874832273419/794148905889693718/1063432715875131462 (https://stackoverflow.com/questions/75108957/assertionerror-the-algorithm-only-supports-class-gym-spaces-box-box-as-acti) For short, we have checks for supported obs/action space (`isinstance(...)`) that will fail when someone is using gym instead of gymnasium or the other way around. We will also have issues of saved models (we check that the obs space is the same after loading). I don't know what is the best way to handle it yet, maybe @pseudo-rnd-thoughts has an idea? (one thing I see is monkey patching, pretty ugly but at least it works: https://github.com/DLR-RM/rl-baselines3-zoo/blob/feat/gym-0.24/rl_zoo3/gym_patches.py#L52)

Looking at the `gym_patches.py` file, tianshou is attempting to take a similar approach with a global variance for `gym` determined by an OS variable. 
Gym and gymnasium should be mostly compatible until v0.30.0, there we will provide a rewrite of the `VectorEnv` and wrapper implementations. 

I see there being three solution
1. Only support gym with this PR solely being update to v0.26
2. Update to v0.26 with support for either gym or gymnasium (v0.27 for atari support)
3. Update to only support gymnasium

For now, I would do solution 2 using a OS variable to determine if to use gym or gymnasium. 

Then making another major update at gymnasium 1.0.0

Rocamonde:
Hey @pseudo-rnd-thoughts, thanks for weighing in. Do you have a link to your development roadmap for gymnasium?

pseudo-rnd-thoughts:
https://github.com/farama-Foundation/gymnasium/issues/12

pseudo-rnd-thoughts:
I should note a fourth option taken by Tianshou, (I was wrong on their implementation). 
4. Tianshou are updating to only Gymnasium but are adding functionality to automatically apply the gym compatibility wrapper if a gym env is detected

araffin:
> I should note a fourth option taken by Tianshou, (I was wrong on their implementation). 4. Tianshou are updating to only Gymnasium but are adding functionality to automatically apply the gym compatibility wrapper if a gym env is detected

yep, saw that yesterday and added it to the checklist ;)

qgallouedec:
Just for the ref: https://github.com/carlosluis/stable-baselines3/pull/5

wkoot:
Would moving to gymnasium >=0.27.0 be feasible? It adds Python 3.11 support, as does the current pytorch nightly

araffin:
> Would moving to gymnasium >=0.27.0 be feasible? It adds Python 3.11 support, as does the current pytorch nightly

@wkoot @juhannc as mentioned in the description of the PR and in some comments, this is already possible (I myself have SB3 running with python 3.11 and this PR locally).
I'll be updating soon to gymnasium as default once some fixes have been pushed to shimmy that will enable a smooth transition from gym 0.21 to 0.26+.

wkoot:
> I'll be updating soon to gymnasium as default once some fixes have been pushed to shimmy that will enable a smooth transition from gym 0.21 to 0.26+.

Awesome, thanks for the quick reply :) I guess the CI test matrix can include 3.11 in that case?
Torch is the only other dependency which does not entirely support it yet, but the 2.0.0 nightly is working.
However, according to https://pytorch.org/get-started/pytorch-2.0/, this will be shipped next month

araffin:
@pseudo-rnd-thoughts I created a branch where gymnasium is now default: https://github.com/DLR-RM/stable-baselines3/pull/1327
I also listed the additional issues caused by this switch.
I added compatibility wrappers so the transition should be smoother (once some bugs are fixed in shimmy), SB3 with this branch works out of the box with gym 0.21, 0.26 and gymnasium.

pseudo-rnd-thoughts:
@araffin Thanks for doing this, very impressive to have all of the different versions working together. 
I will response on the branch, we are planning on fixing all the error you list in the next release (hopefully very soon) except the wrapper issue which we are looking into to see if there is a solution.

araffin:
In term of features, I consider that branch to be ready.
Now, I would like to ask anyone who can help to review this PR (you don't have to review everything and you can do it over multiple days).
I don't expect major bugs, but mostly some polishing missing.
Once this is done, I will close this PR and do a 2.0.0 alpha release using the gymnasium branch (https://github.com/DLR-RM/stable-baselines3/pull/1327).

In the meantime, I will add a warning to say that the v1.8.0 will be the last one with gym 0.21 as main backend.

Tagging @carlosluis @tlpss @Rocamonde @pseudo-rnd-thoughts (am I missing someone?)

.github/ISSUE_TEMPLATE/custom_env.yml
```diff
@@ -49,15 +49,16 @@ body:
                 self.observation_space = spaces.Box(low=-np.inf, high=np.inf, shape=(14,))
                 self.action_space = spaces.Box(low=-1, high=1, shape=(6,))
 
-            def reset(self):
-                return self.observation_space.sample()
+            def reset(self, seed=None):
```
pseudo-rnd-thoughts, line 52: Add `options` keyword

setup.py
```diff
@@ -124,8 +122,9 @@
         "extra": [
             # For render
             "opencv-python",
+            "pygame",
             # For atari games,
-            "ale-py==0.7.4",
+            "ale-py~=0.8.0",
```
pseudo-rnd-thoughts, line 127: 0.8.1

Rocamonde (commented):
Added some comments on the first few files. Apologies if I raise obvious points that have been discussed extensively, have been out of the loop on this PR for a while. Let me know if there are any discussions/comments I should read before reviewing more.

Happy to see the beginning of the end for this PR! Congratulations to everyone involved.

setup.py
```diff
@@ -76,7 +76,7 @@
     packages=[package for package in find_packages() if package.startswith("stable_baselines3")],
     package_data={"stable_baselines3": ["py.typed", "version.txt"]},
     install_requires=[
-        "gym==0.21",  # Fixed version due to breaking changes in 0.22
+        "gym==0.26.2",
```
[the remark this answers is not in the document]
pseudo-rnd-thoughts: v0.26.2 is the last release of gym as development on gym has stopped in favor of gymnasium, we don't expect a new version of gym to be released without a critical reason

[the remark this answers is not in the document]
araffin: since v0.19, all gym/gymnasium introduced breaking changes (more or less documented), so until it is fully stable again, i prefer to pin the version.
It also doesn't prevent user from trying newer versions if they really want to.

setup.py
araffin: fixed in the Gymnasium branch (where we rely on shimmy)

stable_baselines3/common/atari_wrappers.py
```diff
@@ -168,9 +173,10 @@ def step(self, action: int) -> GymStepReturn:
         :return: observation, reward, done, information
```
juhannc, line 173: Not in line with the true return of the function

Dockerfile
```diff
@@ -3,6 +3,9 @@ FROM $PARENT_IMAGE
 ARG PYTORCH_DEPS=cpuonly
 ARG PYTHON_VERSION=3.7
 
+# for tzdata
```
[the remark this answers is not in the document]
araffin: I think I did that because it was somehow needed (the image wouldn't build otherwise).
But as Gitlab restricted their CI, I haven't build/tested the image for a while.
I'm also thinking about using [cog](https://github.com/replicate/cog) instead.

stable_baselines3/common/envs/identity_env.py
```diff
@@ -34,18 +34,21 @@ def __init__(self, dim: Optional[int] = None, space: Optional[spaces.Space] = No
...
-    def step(self, action: T) -> Tuple[T, float, bool, Dict[str, Any]]:
+    def step(self, action: T) -> Tuple[T, float, bool, bool, Dict[str, Any]]:
         reward = self._get_reward(action)
         self._choose_next_state()
         self.current_step += 1
-        done = self.current_step >= self.ep_length
-        return self.state, reward, done, {}
+        done = False
+        truncated = self.current_step >= self.ep_length
+        return self.state, reward, done, truncated, {}
```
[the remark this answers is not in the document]
juhannc: Still quite a few occurrences:

<details>
<summary> grep for `, done, truncated`</summary>

```shell
$  grep -rni ", done, truncated" .
./tests/test_monitor.py:25:        _, reward, done, truncated, _ = monitor_env.step(monitor_env.action_space.sample())
./tests/test_monitor.py:78:        _, _, done, truncated, _ = monitor_env1.step(monitor_env1.action_space.sample())
./tests/test_monitor.py:101:            _, _, done, truncated, _ = monitor_env2.step(monitor_env2.action_space.sample())
./tests/test_utils.py:231:        obs, reward, done, truncated, info = self.env.step(action)
./tests/test_vec_envs.py:46:        return self.state, reward, done, truncated, {}
./tests/test_vec_envs.py:182:        return np.array([prev_step], dtype="int"), 0.0, done, truncated, {}
./tests/test_gae.py:45:        return self.observation_space.sample(), reward, done, truncated, {}
./tests/test_dict_env.py:73:        return self.observation_space.sample(), reward, done, truncated, {}
./tests/test_env_checker.py:19:        return observation, reward, done, truncated, info
./tests/test_vec_normalize.py:38:        return np.array([returned_value]), returned_value, done, truncated, {}
./tests/test_buffers.py:39:        return obs, reward, done, truncated, {}
./tests/test_buffers.py:67:        return obs, reward, done, truncated, {}
./docs/guide/examples.rst:480:      obs, reward, done, truncated, info = env.step(action)
./stable_baselines3/common/envs/bit_flipping_env.py:187:        return obs, reward, done, truncated, info
./stable_baselines3/common/envs/identity_env.py:51:        return self.state, reward, done, truncated, {}
./stable_baselines3/common/envs/identity_env.py:83:        return self.state, reward, done, truncated, {}
./stable_baselines3/common/envs/identity_env.py:156:        return self.observation_space.sample(), reward, done, truncated, {}
./stable_baselines3/common/envs/multi_input_envs.py:161:        return self.get_state_mapping(), reward, done, truncated, {"got_to_end": got_to_end}
```

</details>

Dockerfile
```diff
@@ -20,7 +23,7 @@ RUN curl -o ~/miniconda.sh https://repo.anaconda.com/miniconda/Miniconda3-latest
      ~/miniconda.sh -b -p /opt/conda && \
      rm ~/miniconda.sh && \
      /opt/conda/bin/conda install -y python=$PYTHON_VERSION numpy pyyaml scipy ipython mkl mkl-include && \
-     /opt/conda/bin/conda install -y pytorch $PYTORCH_DEPS -c pytorch && \
+     /opt/conda/bin/conda install -y pytorch=1.11 $PYTORCH_DEPS -c pytorch && \
```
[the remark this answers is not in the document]
araffin: this choice was deliberate. Before Gitlab restricted their CI, the docker image represented the minimal config needed to use SB3 (so python 3.7 and minimum torch version), but yes, it would make sense to have an env variable or remove the pinned version (I must have added it to force the correct pytorch version).
Switching to cog would probably solve that issue too.

docs/guide/vec_envs.rst
```diff
@@ -44,6 +44,38 @@ SubprocVecEnv ✔️       ✔️           ✔️        ✔️         ✔️
...
+  Additional information is passed via the ``infos`` value which is a list of dictionaries.
+- at the end of an episode, ``infos[env_idx]["TimeLimit.truncated"] = truncated and not terminated``
+  tells the user if an episode was truncated or not:
+  you should bootstrap when ``infos[env_idx]["TimeLimit.truncated"] is True`` or ``dones[env_idx] is False``.
+  Note: compared to Gym 0.26+ ``infos[env_idx]["TimeLimit.truncated"]`` and ``terminated`` `are mutually exclusive <https://github.com/openai/gym/issues/3102>`_.
+- at the end of an episode, because the environment resets automatically,
+  we provide ``infos[env_idx]["terminal_observation"]`` which contains the last observation
+  of an episode (and can be used when bootstrapping, see note in the previous section)
+- if you pass ``render_mode="rgb_array"`` to your Gym env, a corresponding VecEnv can automatically show the rendered image
+  by calling ``vec_env.render(mode="human")``. This is different from Gym which currently `doesn't allow multiple render modes <https://github.com/Farama-Foundation/Gymnasium/issues/100>`_
```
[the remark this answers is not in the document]
araffin: I guess the sentence is not clear.
The idea is to overcome the current limitation (only one render mode allowed per env instance, see linked issue), we recommend using `rgb_array` as we can both have the image as numpy array and display it using OpenCV.
if no mode is passed or `mode="rgb_array"` is passed then we use the default mode, otherwise, we use the opencv display.

docs/guide/vec_envs.rst
```diff
@@ -44,6 +44,38 @@ SubprocVecEnv ✔️       ✔️           ✔️        ✔️         ✔️
...
+  and doesn't allow passing a ``mode`` parameter. Note that if ``render_mode != "rgb_array"``, you can only call ``vec_env.render()`` (without argument or with ``mode=env.render_mode``).
+- the ``reset()`` method doesn't take any parameter. If you want to seed the pseudo-random generator,
+  you should call ``vec_env.seed(seed=seed)`` and ``obs = vec_env.reset()`` afterward.
+  If your Gym env implements a ``seed()`` method then it will be called,
+  otherwise ``env.reset(seed=seed)`` will be called (in that case, you will need two resets to set the seed).
```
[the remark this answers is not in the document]
araffin: the idea is to keep the same behavior for gym 0.21 when using newer SB3 version.
All gym 0.26+ env must have a `seed` argument, so the reset will be called and not the seed method (I probably need to rephrase the sentence).

The code is that one: https://github.com/DLR-RM/stable-baselines3/blob/feat/gymnasium-support/stable_baselines3/common/utils.py#L547-L559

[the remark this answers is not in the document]
araffin: this might not be needed for the gymnasium branch which already features compatibility layer

stable_baselines3/common/type_aliases.py
```diff
@@ -17,7 +17,9 @@
 
 GymEnv = Union[gym.Env, vec_env.VecEnv]
 GymObs = Union[Tuple, Dict[str, Any], np.ndarray, int]
+Gym26ResetReturn = Tuple[GymObs, Dict]
 GymStepReturn = Tuple[GymObs, float, bool, Dict]
+Gym26StepReturn = Tuple[GymObs, float, bool, bool, Dict]
```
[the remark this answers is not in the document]
araffin: >  feel like naming the new types Gym26... suggests that they are incompatible with Gym... types,

It is incompatible, and it used to be used together with `GymStepReturn`.
We need to check the contrib repo and the RL Zoo to be sure it is not used anymore, in the case, we can keep the name yes.
Anyway, with gymnasium, we will have to rewrite those annotations (see gymnasium branch and the new annotations for atari wrappers).

docs/guide/vec_envs.rst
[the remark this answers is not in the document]
Rocamonde: That’s pretty clear, improving the phrasing a little would help, thanks!

stable_baselines3/common/envs/identity_env.py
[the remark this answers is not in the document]
araffin: fixed in https://github.com/DLR-RM/stable-baselines3/pull/780/commits/06ad5a88c713db46604c0e3b1127b6ae5a2be45b

stable_baselines3/common/type_aliases.py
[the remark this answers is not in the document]
araffin: It is indeed used in the RL Zoo (I will have to leave them until we transition to the new annotations), I will probably rename them though (need to check if it's fine for the other repos)

docs/guide/vec_envs.rst
```diff
@@ -44,6 +44,38 @@ SubprocVecEnv ✔️       ✔️           ✔️        ✔️         ✔️
...
+- only the initial call to ``vec_env.reset()`` is required, environments are reset automatically afterward (and ``reset_infos`` is updated automatically).
+- the ``vec_env.step(actions)`` method expects an array as input
+  (with a batch size corresponding to the number of environments) and returns a 4-tuple (and not a 5-tuple): ``obs, rewards, dones, infos`` instead of ``obs, reward, terminated, truncated, info``
+  where ``dones = terminated or truncated`` (for each env).
+  ``obs, rewards, dones`` are numpy arrays with shape ``(n_envs, shape_for_single_env)`` (so with a batch dimension).
+  Additional information is passed via the ``infos`` value which is a list of dictionaries.
+- at the end of an episode, ``infos[env_idx]["TimeLimit.truncated"] = truncated and not terminated``
+  tells the user if an episode was truncated or not:
+  you should bootstrap when ``infos[env_idx]["TimeLimit.truncated"] is True`` or ``dones[env_idx] is False``.
+  Note: compared to Gym 0.26+ ``infos[env_idx]["TimeLimit.truncated"]`` and ``terminated`` `are mutually exclusive <https://github.com/openai/gym/issues/3102>`_.
```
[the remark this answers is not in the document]
araffin: done

araffin:
Closing this one as I just released an alpha version of the Gymnasium branch on PyPi:
```
# sb3 contrib installs sb3 too
pip install "sb3_contrib>=2.0.0a1" --upgrade
```
Documentation is available here: https://stable-baselines3.readthedocs.io/en/feat-gymnasium-support/

rfali:
@araffin if I understand correctly, SB3 directly jumped from supporting `gym==0.21` to `gymnasium==0.28.1`?

araffin:
SB3 2.x (master version) supports gymnasium 0.28.1 and gym 0.21/0.26 via shimmy

rfali:
@araffin is there any documentation/reference on how to use SB3 with gym 0.21/0.26 via shimmy? I saw the[ install requirements ](https://github.com/DLR-RM/stable-baselines3/blob/fd0cd82339511b54cd3907df228a656f2a32f0b8/setup.py#L105)of master branch and it only mentioned gymnasium (does not include gym).

araffin:
>  is there any documentation/reference on how to use SB3 with gym 0.21/0.26 via shimmy? 

It is automatic if you use `make_vec_env` (and if you pass the env to the constructor), otherwise take a look at shimmy documentation.
https://github.com/DLR-RM/stable-baselines3/blob/fd0cd82339511b54cd3907df228a656f2a32f0b8/stable_baselines3/common/env_util.py#L99-L101

https://github.com/DLR-RM/stable-baselines3/blob/master/stable_baselines3/common/vec_env/dummy_vec_env.py#L30

Patch env is defined here:
https://github.com/DLR-RM/stable-baselines3/blob/master/stable_baselines3/common/vec_env/patch_gym.py#L15

and yes gym is an optional dependency.

Issue #807, opened 2022-03-07
[Bug] Unchecked None value in random seeding of SubprocVecEnv


ScheiklP:
### 🐛 Bug
`seed()` has optional `seed` kwarg, but does not check for `None` -> trying to increment with `seed + idx` crashes, because `+` is not defined for `None` and `int`.
https://github.com/ScheiklP/stable-baselines3/blob/stacking_scalar_boxes/stable_baselines3/common/vec_env/subproc_vec_env.py#L127

```python
  def seed(self, seed: Optional[int] = None) -> List[Union[None, int]]:
      for idx, remote in enumerate(self.remotes):
          remote.send(("seed", seed + idx))
      return [remote.recv() for remote in self.remotes]
```

### Expected behavior

If kwarg is not set, sample a first random seed through `np.random.randint(0, 2**32 - 1)`.

### To Reproduce
```python
import gym
from stable_baselines3.common.vec_env import SubprocVecEnv

if __name__ == '__main__':
    env_id = "CartPole-v1"
    num_cpu = 4  # Number of processes to use
    # Create the vectorized environment
    env = SubprocVecEnv([gym.make_env(env_id) for i in range(num_cpu)])
    env.seed()
```

### Checklist

- [x] I have checked that there is no similar [issue](https://github.com/DLR-RM/stable-baselines3/issues) in the repo (**required**)
- [x] I have read the [documentation](https://stable-baselines3.readthedocs.io/en/master/) (**required**)
- [x] I have provided a minimal working example to reproduce the bug (**required**)

Miffyli:
Good catch! A PR to add the suggestion you proposed would be warmly welcomed :).

PR #808, opened 2022-03-07
Fixed unchecked None value in SubprocVecEnv

ScheiklP:
## Description
Fix for bug described in https://github.com/DLR-RM/stable-baselines3/issues/807

closes https://github.com/DLR-RM/stable-baselines3/issues/807

## Types of changes

## Checklist:


stable_baselines3/common/vec_env/subproc_vec_env.py
```diff
@@ -123,6 +123,8 @@ def step_wait(self) -> VecEnvStepReturn:
         return _flatten_obs(obs, self.observation_space), np.stack(rews), np.stack(dones), infos
 
     def seed(self, seed: Optional[int] = None) -> List[Union[None, int]]:
+        if seed is None:
```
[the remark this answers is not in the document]
ScheiklP: Ah, no, I just did not check. Fixed in a new commit.

stable_baselines3/common/vec_env/dummy_vec_env.py
```diff
@@ -51,6 +51,8 @@ def step_wait(self) -> VecEnvStepReturn:
         return (self._obs_from_buf(), np.copy(self.buf_rews), np.copy(self.buf_dones), deepcopy(self.buf_infos))
 
     def seed(self, seed: Optional[int] = None) -> List[Union[None, int]]:
+        if seed is None:
```
araffin, line 54: why not just return in that case?

ScheiklP: From intuition I would assume that if I call `env.seed()` without parameters, that a seed would be chosen for me randomly.
I mostly want to ensure, that all envs of the vecenv have a different seed. If we just return, all of them will have the same seed, when the processes are forked.

araffin: I see your point, I thought that the seed would have been different per process but a quick test shows it's not the case when processes are forked (which is not the default method of SB3 luckily), so the following yield a different result per process:

```python
import multiprocessing as mp

import numpy as np

def f(x):
    return np.random.rand()

if __name__ == '__main__':
    ctx = mp.get_context('spawn')
    with ctx.Pool(3) as p:
        print(p.map(f, [1, 1, 1]))
```

ScheiklP: Huh, for me the output of your example is the same over all processes with fork.
```
import multiprocessing as mp

import numpy as np

def f(x):
    return np.random.rand()

if __name__ == '__main__':
    ctx = mp.get_context('fork')
    with ctx.Pool(3) as p:
        print(p.map(f, [1, 1, 1]))
```
prints
```
[0.5667669684783624, 0.5667669684783624, 0.5667669684783624]
```

Yeah, sure, the `seed = np.random.randint(0, 2**32 - 1)` will be deterministic, if you force a seed before, but the subprocesses will still have their own unique seed.

araffin: > Huh, for me the output of your example is the same over all processes with fork.

yes, but that's not the case for `SuprocVecEnv` with start method fork for some reasons... (see test below, you can comment out `seed = np.random.randint(0, 2**32 - 1)` part and return None to check what happens)

araffin: > with start method fork for some reasons...

nevermind, I think the example in the test is using a separate random seed generator. Let me try.

araffin: See https://github.com/DLR-RM/stable-baselines3/pull/808/commits/a64714b546b82841bed456c53c36c6e84fec6392

araffin (approved):
LGTM, thanks =)

PR #848, opened 2022-04-01
Fix HER goal selection

### LINKED ISSUE #847: [Bug] HER: final goal selection strategy does not sample the final achieved goal
### 🐛 Bug

When you choose the final goal selection strategy, HER samples the second to last goal instead of the last goal (see the code).

If the episode is _[s0, s1, s2, s3]_, it should sample _s3_, but it samples _s2_.

The problem is simple to understand in the "final" case. But I think that this problem concerns all three goal selection strategies.

### To Reproduce

```python
import numpy as np
from stable_baselines3 import HerReplayBuffer
import gym

# Env creation is not relevant, it is just to make the code work.
env = make_vec_env(
    BitFlippingEnv,
    env_kwargs=dict(n_bits=2),
    wrapper_class=gym.wrappers.TimeLimit,
    wrapper_kwargs=dict(max_episode_steps=10),
)

# A simple episode
achieved_goals = np.array([
        [[0, 0]],
        [[1, 1]],
        [[2, 2]],
])

# Actions, rewards, true goals and info are not relevant, they will be constant
action = np.array([[0.0]])
reward = np.array([0.0])
infos = [{}]
desired_goals = np.zeros_like(achieved_goals)
observations = np.zeros_like(achieved_goals)


# Create the buffer
buffer = HerReplayBuffer(
    env,
    buffer_size=100,
    n_sampled_goal=np.inf,  # all sampled goals are relabelled
    goal_selection_strategy="final",
)
# And store the transitions
buffer.add(
    obs={"observation": observations[0], "achieved_goal": achieved_goals[0], "desired_goal": desired_goals[0]},
    next_obs={"observation": observations[1], "achieved_goal": achieved_goals[1], "desired_goal": desired_goals[1]},
    action=action,
    reward=reward,
    done=np.array([False]),
    infos=infos,
)

buffer.add(
    obs={"observation": observations[1], "achieved_goal": achieved_goals[1], "desired_goal": desired_goals[1]},
    next_obs={"observation": observations[2], "achieved_goal": achieved_goals[2], "desired_goal": desired_goals[2]},
    action=action,
    reward=reward,
    done=np.array([True]),
    infos=infos,
)

samples = buffer.sample(1)
print(samples.observations["desired_goal"]) # expect [2, 2], got [1, 1]
```

### Could fix the issue

I think the cause is that the new sampled goal comes from `achieved_goal` and not from `next_achived_goal`. Potential fix is to replace

https://github.com/DLR-RM/stable-baselines3/blob/44e53ff8115e8f4bff1d5218f10c8c7d1a4cfc12/stable_baselines3/her/her_replay_buffer.py#L265

by

```python
return self._buffer["next_achieved_goal"][her_episode_indices, transitions_indices] 
```

### System Info

 * OS: macOS-12.2.1-x86_64-i386-64bit Darwin Kernel Version 21.3.0: Wed Jan  5 21:37:58 PST 2022; root:xnu-8019.80.24~20/RELEASE_X86_64
 * Python: 3.8.13
 * Stable-Baselines3: 1.4.1a3
 * PyTorch: 1.11.0
 * GPU Enabled: False
 * Numpy: 1.22.3
 * Gym: 0.21.0

### Checklist

qgallouedec:
## Description

closes #847 

## Motivation and Context

## Types of changes

## Checklist:


stable_baselines3/her/her_replay_buffer.py
```diff
@@ -304,14 +304,6 @@ def _sample_transitions(
 
         ep_lengths = self.episode_lengths[episode_indices]
 
-        # Special case when using the "future" goal sampling strategy
```
araffin, line 307: why did you remove that?
this is different from the issue you raised, no?

qgallouedec: I think it's not.

Transitions are stored that way: 
```
(s0, g0, s1, g1), 
(s1, g1, s2, g2), 
(s2, g2, s3, g3), 
(s3, g3, s4, g4), 
```

The issue I raised is that the new desired goal are sampled from `achieved_goal` (first column) instead of `next_achieved_goal` (third column). Since you want _strictly next_ achieved goal, you add a +1 here

 https://github.com/DLR-RM/stable-baselines3/blob/44e53ff8115e8f4bff1d5218f10c8c7d1a4cfc12/stable_baselines3/her/her_replay_buffer.py#L255

If `s1` is sampled, you want the new desired goal to be either `s2`, `s3` or `s4`.
But, with the current implementation, you can only sample `s2` and `s3` as a new goal (that's the issue). 

Similarly, if `s3` is sampled, you want the new desired goal to be `s4` (only one choice).
But, with the current implementation, you cannot sample any goal. That's why this special case has been tackled.

With a correct implementation, even when the episode is a single transition `[(s0, g0, s1, g1)]`, you can sample a strictly future `achieved_goal`, by taking `s1`. So you don't need that fix anymore.

araffin: I see, thanks for the clarification =)

araffin (approved):
LGTM =) (I updated the changelog too)

Issue #875, opened 2022-04-19
[Bug] atari_wrapper - numpy problem (AttributeError: 'numpy.random._generator.Generator' object has no attribute 'randint')


semihtasbas:
Hi, 

I try to use your Atari example in doc, but I receive an error about numpy. I couldn't find a solution about that. I even delete my anaconda and setup environment from scratch. I shared details below.

```python
from stable_baselines3.common.env_util import make_atari_env
from stable_baselines3.common.vec_env import VecFrameStack
from stable_baselines3 import A2C

# There already exists an environment generator
# that will make and wrap atari environments correctly.
# Here we are also multi-worker training (n_envs=4 => 4 environments)
env = make_atari_env('PongNoFrameskip-v4', n_envs=4, seed=0)
# Frame-stacking with 4 frames
env = VecFrameStack(env, n_stack=4)

model = A2C('CnnPolicy', env, verbose=1)
model.learn(total_timesteps=25_000)

obs = env.reset()
while True:
    action, _states = model.predict(obs)
    obs, rewards, dones, info = env.step(action)
    env.render()
```
Output: 

```bash
A.L.E: Arcade Learning Environment (version 0.7.5+db37282)
[Powered by Stella]
Using cuda device
Wrapping the env in a VecTransposeImage.
Traceback (most recent call last):
  File "/home/x/PycharmProjects/tries/main.py", line 207, in <module>
    model.learn(total_timesteps=25_000)
  File "/home/x/anaconda3/envs/rl/lib/python3.7/site-packages/stable_baselines3/a2c/a2c.py", line 200, in learn
    reset_num_timesteps=reset_num_timesteps,
  File "/home/x/anaconda3/envs/rl/lib/python3.7/site-packages/stable_baselines3/common/on_policy_algorithm.py", line 243, in learn
    total_timesteps, eval_env, callback, eval_freq, n_eval_episodes, eval_log_path, reset_num_timesteps, tb_log_name
  File "/home/x/anaconda3/envs/rl/lib/python3.7/site-packages/stable_baselines3/common/base_class.py", line 429, in _setup_learn
    self._last_obs = self.env.reset()  # pytype: disable=annotation-type-mismatch
  File "/home/x/anaconda3/envs/rl/lib/python3.7/site-packages/stable_baselines3/common/vec_env/vec_transpose.py", line 110, in reset
    return self.transpose_observations(self.venv.reset())
[...]

PR #888, opened 2022-04-26
Compatibility with new Numpy Random Generator

KaiOnGitHub:
np_random.randint calls a legacy method which causes an AttributeError because gym uses the new numpy Random Generator that does not have a randint method.
Legacy Generator: https://numpy.org/doc/stable/reference/random/generated/numpy.random.RandomState.randint.html
New Generator that is used by gym: https://numpy.org/doc/stable/reference/random/generated/numpy.random.Generator.integers.html
Quote: "New code should use the integers method of a default_rng() instance instead; please see the Quick Start."

Screenshot 2022-04-26 at 16 36 45

### DIFF
--- a/stable_baselines3/common/atari_wrappers.py
+++ b/stable_baselines3/common/atari_wrappers.py
@@ -33,7 +33,10 @@ class NoopResetEnv(gym.Wrapper):
         if self.override_num_noops is not None:
             noops = self.override_num_noops
         else:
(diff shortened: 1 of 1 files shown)


KaiOnGitHub:
Oh I found the duplicate issue that is closed already: https://github.com/DLR-RM/stable-baselines3/issues/875

PR #905, opened 2022-05-07
Removed redundant double-check for nested observation environments

TibiGG:
In the creation of the BaseAlgorithm class, when the environment from a given name is created, a check is being made to ensure the environment does not present a nested Dict/Tuple space.

This is done first by the `check_for_nested_spaces` function.

After the function was being called, similar logic was introduced immediately afterwards, to check the environment did not have nested Dict observation space. This seems to be a redundant operation, so I removed it.

If I did not take some side effect into consideration, please let me know.

### DIFF
--- a/docs/guide/integrations.rst
+++ b/docs/guide/integrations.rst
@@ -137,3 +137,56 @@ Then, in this example, we train a PPO agent to play CartPole-v1 and push it to a
       filename="ppo-CartPole-v1",
       commit_message="Added Cartpole-v1 model trained with PPO",
   )
+
+MLFLow
+======
+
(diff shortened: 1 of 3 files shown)


araffin:
Hello,
thanks for the proposal, please fill up the PR template completely.

The change was made by @Miffyli apparently (https://github.com/J-Travnik/stable-baselines3/commit/f770217a44365805a2b97023eeae1a8c1e30d024), @Miffyli , do you remember any reason for this double check?

The only difference I see is that you are checking for nested dict only in the second check (where the first check, checks for tuple and dict)

Issue #934, opened 2022-06-17
[Bug] optimize_memory_usage not compatible with handle_timeout_termination


MWeltevrede:
### 🐛 Bug

When using the [ReplayBuffer class](https://github.com/DLR-RM/stable-baselines3/blob/d68f0a2411766beb6da58ee0e989d1a6a72869bc/stable_baselines3/common/buffers.py#L153), setting both `optimize_memory_usage = True` and `handle_timeout_termination = True` will lead to incorrect behaviour.

This is because when `handle_timeout_termination = True`, the replay buffer will set the `done` value at the end of an episode to `False` if the end of an episode was due to a timeout:
https://github.com/DLR-RM/stable-baselines3/blob/d68f0a2411766beb6da58ee0e989d1a6a72869bc/stable_baselines3/common/buffers.py#L300-L302

In an algorithm like DQN, this means the Q value target will bootstrap using the `next_observation` variable
https://github.com/DLR-RM/stable-baselines3/blob/d68f0a2411766beb6da58ee0e989d1a6a72869bc/stable_baselines3/dqn/dqn.py#L196

However, this leads to incorrect behaviour if `optimize_memory_usage = True`, because the `next_observation` variable in that case is defined as:
https://github.com/DLR-RM/stable-baselines3/blob/d68f0a2411766beb6da58ee0e989d1a6a72869bc/stable_baselines3/common/buffers.py#L291-L292
Which will get overwritten by the first state of the new episode. So, when `optimize_memory_usage = True`,  the replay buffer will not return the last state of the episode (which would be correct behaviour).

As a result, if both `optimize_memory_usage = True` and `handle_timeout_termination = True`, a reinforcement learning algorithm will sometimes have target values that bootstrap from the wrong state (the first state of the next episode, rather than the last state of the current episode)

> Note that this behaviour does not happen if `optimize_memory_usage = False`, because in that case the `next_observation` variable is stored in a separate buffer and won't be overwritten by the first state of the new episode.

### To Reproduce

This bug only affects the Q value updates if there are end of episodes due to timeouts in the environment. In practice, the difference between the correct target and the incorrect target can be small. 
For this reason, I wasn't able to find a simple scenario in which this will lead to noticeable incorrect training results However, I added a simple code segment that will illustrate the problem and why if both `optimize_memory_usage = True` and `handle_timeout_termination = True` the behaviour will be incorrect.

```python
import gym 
import numpy as np

from stable_baselines3 import DQN

[...]

PR #948, opened 2022-07-01
Prohibit simultaneous use of optimize_memory_usage and handle_timeout_termination

MWeltevrede:
Prohibit the simultaneous use of `optimize_memory_usage = True` and `handle_timeout_termination = True` due to a bug discussed in  #934.

## Description
This implements the hotfix as discussed in #934 that will throw an exception if both `optimize_memory_usage = True` and `handle_timeout_termination = True`. 

It also modifies `test_warn_buffer()` that tests whether the model emits a warning if `optimize_memory_usage = True`. The default value of `handle_timeout_termination` is `True` and so this test would encounter the exception implemented as part of this hotfix. To avoid this issue the test is modified to set `handle_timeout_termination = False`, as this does not interfere with the emitting of the warning.  

## Motivation and Context

## Types of changes

## Checklist:

### DIFF
--- a/tests/test_save_load.py
+++ b/tests/test_save_load.py
@@ -375,6 +375,7 @@ def test_warn_buffer(recwarn, model_class, optimize_memory_usage):
         select_env(model_class),
         buffer_size=100,
         optimize_memory_usage=optimize_memory_usage,
+        replay_buffer_kwargs={"handle_timeout_termination": False},
         policy_kwargs=dict(net_arch=[64]),
         learning_starts=10,
     )
--- a/tests/test_save_load.py
+++ b/tests/test_save_load.py
@@ -375,7 +375,9 @@ def test_warn_buffer(recwarn, model_class, optimize_memory_usage):
         select_env(model_class),
         buffer_size=100,
         optimize_memory_usage=optimize_memory_usage,
-        replay_buffer_kwargs={"handle_timeout_termination": False},
+        # we cannot use optimize_memory_usage and handle_timeout_termination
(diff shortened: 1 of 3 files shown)


tests/test_save_load.py
```diff
@@ -375,6 +375,7 @@ def test_warn_buffer(recwarn, model_class, optimize_memory_usage):
         select_env(model_class),
         buffer_size=100,
         optimize_memory_usage=optimize_memory_usage,
+        replay_buffer_kwargs={"handle_timeout_termination": False},
```
Miffyli, line 378: Hmm why this change? If removed, we might not be testing `handle_timeout_termination` anymore. Ideally tests should use the default values.

Miffyli:
LGTM! Just a tiny comment + quick fix for consistent style :)

tests/test_save_load.py
MWeltevrede: This change is required because `test_warn_buffer()` runs with `optimize_memory_usage=False` and `optimize_memory_usage=True` (it is designed to test different behaviours depending on the different values of that parameter). In the latter case, the default value of `handle_timeout_termination=True` will cause the `ReplayBuffer` to throw an exception (which is the new behaviour this PR implements). 

I agree it's not ideal we have to change the default values to be able to run this test. However, I think that is inevitable unless we properly fix the bug in #934, or change the default value of `handle_timeout_termination` to `False` (which I think is also not ideal).

araffin (approved):

PR #979, opened 2022-07-24
Use higher resolution time_ns() and avoid division by zero

AdamGleave:
## Description
We compute the time elapsed in RL algorithms in order to log both the total time, and the FPS (reciprocal of time elapsed).

This PR changes this time elapsed computation to be a delta of `time.time_ns()`, which is around 2.3x more precise than `time.time()` (see [PEP 564](https://peps.python.org/pep-0564/#annex-clocks-resolution-in-python)). Additionally, to avoid the possibility of division by zero when computing FPS, it takes the maximum of the time delta and the float machine epsilon. This replaces a more ad-hoc approach in `OffPolicyAlgorithm` (not present in `OnPolicyAlgorithm`, strangely) of adding a fixed `1e-8` to the time elapsed.

## Motivation and Context

I found that on Windows (lower precision timer than Linux) the computed time elapsed on quick environments for smaller numbers of timesteps would sometimes be `0.0`, resulting in a division by zero error. This ended up producing a quite annoying flaky test. I doubt it'd crop up in real use cases where we tend to have `total_timesteps` larger than 10 (my unit test...), so this is a fairly minor fix, but shouldn't harm anything.

## Types of changes

## Checklist:

### DIFF
--- a/stable_baselines3/common/off_policy_algorithm.py
+++ b/stable_baselines3/common/off_policy_algorithm.py
@@ -428,7 +428,7 @@ class OffPolicyAlgorithm(BaseAlgorithm):
         """
         Write log.
         """
-        time_elapsed = max(time.time() - self.start_time, sys.float_info.epsilon)
+        time_elapsed = max(time.time_ns() - self.start_time, sys.float_info.epsilon)
         fps = int((self.num_timesteps - self._num_timesteps_at_start) / time_elapsed)
         self.logger.record("time/episodes", self._episode_num, exclude="tensorboard")
         if len(self.ep_info_buffer) > 0 and len(self.ep_info_buffer[0]) > 0:
--- a/stable_baselines3/common/off_policy_algorithm.py
+++ b/stable_baselines3/common/off_policy_algorithm.py
@@ -1,5 +1,6 @@
 import io
 import pathlib
+import sys
 import time
 import warnings
 from copy import deepcopy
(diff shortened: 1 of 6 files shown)


stable_baselines3/common/off_policy_algorithm.py
```diff
@@ -427,14 +428,14 @@ def _dump_logs(self) -> None:
         """
         Write log.
         """
-        time_elapsed = time.time() - self.start_time
-        fps = int((self.num_timesteps - self._num_timesteps_at_start) / (time_elapsed + 1e-8))
+        time_elapsed_ns = max(time.time_ns() - self.start_time, sys.float_info.epsilon)
```
araffin, line 431: why not multiply here directly by 1e9? (as you do it twice later)

araffin (commented):
Except minor comment, LGTM =)

Once it is merged, could you do a similar PR for the contrib repo?

stable_baselines3/common/off_policy_algorithm.py
AdamGleave: Good point, I think I was trying to avoid rounding error, but it's fine to divide by `1e9` after computing the delta.

AdamGleave:
Thanks for the review! Addressed the comment now.

Sure, I'll work on the contrib PR

araffin (approved):
LGTM, thanks =)

PR #1004, opened 2022-08-11
Include `running_mean` and `running_val` when updating target networks

### LINKED ISSUE #996: [Question] Does DQN copy `running_mean` and `running_var` of batch norm layers to the target network?
I have been spending quite some time reading the codes here and I have been learning quite a lot so far. I got a small question when I backtrack some codes to find out why I got some unstable agents.

https://github.com/DLR-RM/stable-baselines3/blob/b1cc15970a40c86b26a247fab5783e025bfe3da1/stable_baselines3/dqn/dqn.py#L172

This line above seems to be the place where the target network is updated. But if there are batch norm layers in the network, we have some extra variables `running_mean` and `running_var` which are not contained in `Module.parameters()`. We know these are slightly different than those SGD updated parameters and they just keep tracking batch mean/var actively unless `.eval()` is called. 

It doesn't look like they are copied during target network update. But I believe the DDQN paper wants the target network to be a "periodic copy" of the original network which seems to imply that everything should be copied and stayed the same for certain amount of steps. 

Or are those `running_mean` and `running_var` updated somewhere else in the code base?

honglu2875:
## Description
- Created a function `get_parameters_by_name`.
- Created a test `test_get_parameters_by_name()`.
- Before `polyak_update` is applied on target networks in DQN, SAC, TD3, a list of parameters is fetched directly from `state_dict()` by `get_parameters_by_name` whenever the names contain one of the given strings: `weight`, `bias`, `running_`. This way, we not only include weights and biases as those in `model.parameters()`, but also `running_mean` and `running_var` from batch norm layers. When a batch norm layer is present, there is another parameter `num_batches_tracked` but it is ignored (it is `int64` and will become messy when mixed up with float operations anyway).

A few remarks:
- I am not entirely confident about whether it is correct when `tau` is less than 1 (especially because `running_var` is not calculated in a linear way but `polyak_update` is a linear operation). Didn't find discussions on the papers so maybe I am overthinking. But at least I'm sure that when `tau==1`, the original version is not enough and we have to copy everything including `running_mean` and `running_var` as target networks are supposed to be delayed copies of the original networks.
- stable-baseline3 doesn't seem to natively include batch norm layers by design (for example, see `common/torch_layers.py`). So it is quite understandable that the original code does not consider their existence. But people like me already started to smuggle batch norm layers into our projects that use sb3. This fix is merely an extra safeguard so that batch norm users can still get the target update part correct.
- There might be slightly more concise ways to do this. But my thinking is that, if in the future there are other parameters that need to be included, the modification on the code base would be minimal (just include the keywords in the name strings of `state_dict()`).

## Motivation and Context
This is my attempt to fix #996 confirmed by the comment https://github.com/DLR-RM/stable-baselines3/issues/996#issuecomment-1207205719
Basically, when the offline policies DQN, SAC, TD3 update their target networks, only the parameters in `policy.parameters()` are updated. But when a batch norm layer is present, its `running_mean` and `running_var` are not in `policy.parameters()`. This results in the target networks being only partially updated.

closes #996.

## Types of changes

## Checklist:


stable_baselines3/common/utils.py
```diff
@@ -392,6 +392,16 @@ def safe_mean(arr: Union[np.ndarray, list, deque]) -> np.ndarray:
     return np.nan if len(arr) == 0 else np.mean(arr)
 
 
+def get_parameters_by_name(model: th.nn.Module, included_names: Iterable[str]) -> Iterable:
```
araffin, line 395:
```suggestion
def get_parameters_by_name(model: th.nn.Module, included_names: Iterable[str]) -> List[th.Tensor]:
```
a bit more specific for the type hint.

stable_baselines3/dqn/dqn.py
```diff
@@ -169,7 +169,12 @@ def _on_step(self) -> None:
         """
         self._n_calls += 1
         if self._n_calls % self.target_update_interval == 0:
-            polyak_update(self.q_net.parameters(), self.q_net_target.parameters(), self.tau)
+            included_names = ["weight", "bias", "running_"]
```
araffin, line 172: I would rather pre-compute all the keys for batch norm when creating the policy `setup_model()`
and then do a hard copy of the statistics (because a linear interpolation may not make sense).
This would similar to synchronizing the observation normalization.
I think we can take a look at BYOL implementation (if I recall, there is a delayed target net and it has batch norm)

honglu2875: Yeah the linear interpolation was my concern too. In my little project I used tau=0.9 with linear interpolation on running_mean/running_var and for reasons I don't know, it worked well. But I totally agree that a hard copy makes a lot more sense. Haven't read BYOL but I will take a look.

araffin: Looking at BYOL implementation, they actually keep `training=True` for the target net (sounds a bit hacky, I need to ask if it is intented or a lucky oversight)

honglu2875: Wow! I remember at first I also set `training=True` but that wasn't my intention. I specifically compared the running mean and var between the original net and my target net and they are quite different. The result was bad. This was how I noticed this batch norm problem at the beginning.

honglu2875: > I would rather pre-compute all the keys for batch norm when creating the policy `setup_model()` and then do a hard copy of the statistics (because a linear interpolation may not make sense). This would similar to synchronizing the observation normalization. I think we can take a look at BYOL implementation (if I recall, there is a delayed target net and it has batch norm)

Just made the changes. Now I collected all batch norm parameters in `_setup_model()` and made `polyak_update` with `tau=1.0` on them. Let me know if this is what you meant.

honglu2875:
On another note, I have been reading some RL+NLP works and a lot of them also have this target net updating part. In a lot of large language models, people use layer norm instead of batch norm all the time. Layer norm calculates mean/variance and normalize but all within the network itself. It's just that everything in the mean/variance can now have derivatives (and thus can be updated by gradient descent). But they all still seem to do a linear averaging when updating the target networks.

honglu2875:
https://github.com/DLR-RM/stable-baselines3/blob/57e0054e62ac5a964f9c1e557b59028307d21bff/tests/test_train_eval_mode.py#L213
Oh just noticed that (unfortunately after the commit) if we do a hard copy of the running_mean/var, this line of test will be blocking (`tau=0` and nothing should be changed).  @araffin Let me know if we change the test or revert back to linear averaging of batch norm parameters

stable_baselines3/dqn/dqn.py
```diff
@@ -170,6 +173,8 @@ def _on_step(self) -> None:
         self._n_calls += 1
         if self._n_calls % self.target_update_interval == 0:
             polyak_update(self.q_net.parameters(), self.q_net_target.parameters(), self.tau)
+            # Copy running stats, see GH issue #996
```
araffin, line 176: I will leave it like that for now, but if in experiments it doesn't work, we should copy the stats after each gradient step

honglu2875:
> Sure! Will work on it later this week.

PR #1025, opened 2022-08-23
Fix: only yield full batches => avoid nans when len(advantages) == 1, closes #325

### LINKED ISSUE #325: [Bug] NaN's if a single observation is sampled from PPO's rollout_buffer
### 🐛 Bug

I am using PPO and am adapting a few things of it to experiment with an unusual adversary setting, including for now in an ad-hoc manner adjusting the rollout-buffer way to have a dynamic size for it. Doing this I found that PPO can run into nan values whenever it samples a batch size of size 1 due to this in `PPO.py`, line 170:

```
                advantages = (advantages - advantages.mean()) / (advantages.std() + 1e-8)
```
where `advantages.std()` over a batch of size 1 is not defined.

While my current scenario is of course nothing common, this seems to me like a case that could more generally occur whenever someone uses a btach_size of size rollout_buffer.size + 1 or someone wants to use a dynamically sized replay buffer. It could also easily be accounted for.

### To Reproduce

Run PPO with a batch_size of 1 or a rollout_buffer of size batch_size + 1 (i.e. by setting `n_steps` to this value)

### Expected behavior

I would suggest to either discard batches when only a single observation is left in the rollout_buffer or check the size of the sampled buffer and set the std to either 1 or the minimum value in this case.

### System Info

stable-baselines v. 0.10.0

### Checklist

- [X] I have read the [documentation](https://stable-baselines3.readthedocs.io/en/master/) (**required**)
- [X] I have provided a minimal working example to reproduce the bug (**required**)

hughperkins:
Avoid nans caused by sampling batches of 1, when `batch_size` is not a factor of `n_env * n_step`

## Description

Spent 4 days wondering why my training kept ending in nan. Yes, I'd seen the warning that `batch_size` should be a factor of `n_env * n_step`. But it is was only a warning. But it turns out that partial batches get sent through ppo, and partial batches with batch size 1 break the training.

Therefore, this PR modifies the buffer to only yield full batches of length `batch_size`, and no partial batches. I suppose an alternative solution could be to modify ppo.py to reject batches of either size 1, or of less than batch size. (Advantage of the former: is there ever any reason why someone would want partial batch sizes? Advantage of the latter: doesn't modify rollout buffer behavior for people using other algorithms, who might be fine with partial batches)

## Motivation and Context

Spent 4 days wondering why my training kept ending in nan. Yes, I'd seen the warning that `batch_size` should be a factor of `n_env * n_step`. But it is was only a warning. But it turns out that partial batches get sent through ppo, and partial batches with batch size 1 break the training, crashing with an exception for nan.

Related issue here https://github.com/DLR-RM/stable-baselines3/issues/325

https://github.com/DLR-RM/stable-baselines3/issues/325

## Types of changes

## Checklist:


hughperkins:
Alright, I concur that the nans should be fixed by just tweaking the advantage normalization. However, ther is still scope for this PR, but as an option, that deafults to false, orthogonal to the advantage change.

hughperkins:
I'm going to close this for now, since 1. for nans, I created a smaller simpler PR for that, 2. for the batching, I found I'm actually training on all batches now, using the other PR.

PR #1043, opened 2022-09-01
Fix return type for load, learn in BaseAlgorithm

### LINKED ISSUE #1040: [Bug] Too general a return type for load and learn in BaseAlgorithm
### 🐛 Bug

The return type of methods `.load()` and `.learn()` in `BaseAlgorithm` is annotated as `"BaseAlgorithm"`, which means that for any subclass that does not override the methods with suitable annotations, the type won't check if the user is expecting an instance of the subclass. Example:

```python
def load_ppo(cache_path, venv) -> PPO:
    return PPO.load(cache_path, venv)
```

The code above does not type check ("expected PPO, got BaseAlgorithm").

### To Reproduce

Run type checker on code above.

### Expected behavior

Code should type check.

### Solution

The solution to this issue (happy to submit a PR for it) is to replace the return type with the `Self` type variable introduced in [PEP0673](https://peps.python.org/pep-0673). It is built into python 3.11 but it is also available through the PyPI package `typing_extensions`.

Instead of:
https://github.com/DLR-RM/stable-baselines3/blob/304c17dc78dbaeba77c709ec03c1b3847991018d/stable_baselines3/common/base_class.py#L533-L544

we use
```python
if sys.version_info >= (3, 11):
    from typing import Self
else:
    from typing_extensions import Self

class BaseAlgorithm(ABC):
    ...

    @abstractmethod
    def learn(self, ...) -> Self: 
        ...

    @classmethod
    def load(cls, ...) -> Self:
        ...
```

### Checklist

Rocamonde:
## Description
Fixes the return type of `.load()` and `.learn()` methods in `BaseAlgorithm` so that they now use the `Self` type [PEP 0673](https://peps.python.org/pep-0673/) instead of `BaseAlgorithm`, which breaks type checkers for use with any subclass.

## Motivation and Context

 Closes #1040.

## Types of changes

## Checklist:


stable_baselines3/common/base_class.py
```diff
@@ -10,6 +10,7 @@
 import gym
 import numpy as np
 import torch as th
+from typing_extensions import Self
```
araffin, line 13: typing extensions might not be installed (pytype is an optional dependency), I would also keep it optional if possible

AdamGleave: Yep, I think to avoid this dependency we have two options.

Simplest but somewhat ugly is just to go with the TypeVar solution described in https://peps.python.org/pep-0673/ that Self is aimed to replace. Perhaps not so bad if we only need to do this in a few places.

Alternatively, can make the import only happen for type evaluation:
```
if typing.TYPE_CHECKING:
    from typing_extensions import Self
```
Then we might just be able to write `"Self"` instead of `Self`? If string literals not supported, you could also do something janky like:
```
if typing.TYPE_CHECKING:
    from typing_extensions import Self
else:
    Self = "BaseAlgorithm"
```

qgallouedec: > typing extensions might not be installed

No, typing-extensions is installed, since it is a torch dependency since at least version 1.11.

qgallouedec:
Users may use python >= 3.11. So I suggest that you distinguish between the cases:

```python
if sys.version_info >= (3, 11):
    from typing import Self
else:
    from typing_extensions import Self
```

Rocamonde:
This is what I was doing initially. But if typing extensions is installed anyways, that would work regardless. Though I agree it’s cleaner.

stable_baselines3/common/base_class.py
qgallouedec: But to keep it clean, I recommend adding it to the required dependencies of SB3.

Rocamonde:
Apparently pytype does not support typing_extensions.Self yet, lol.

qgallouedec:
> Apparently pytype does not support typing_extensions.Self yet, lol.

It's a shame, I opened a question about it: https://github.com/google/pytype/issues/1283

AdamGleave:
Given `Self` is not supported, we should probably go with the TypeVar solution from https://peps.python.org/pep-0673/ ?

Rocamonde:
it seems like it won't be supported for some time. Best to use the TypeVar solution for now, update it later.

Rocamonde:
We might want to open another issue (or not mark the current one as completed) so that we remember to go back to this in a couple of months when support is added.

Rocamonde:
@AdamGleave @qgallouedec what do you think of the current status?

setup.py
```diff
@@ -82,6 +82,8 @@
         "pandas",
         # Plotting learning curves
         "matplotlib",
+        # for better typing support for users with Python <3.11
```
araffin, line 85: this is not needed anymore, no?
and requirement should be conditioned by python version

docs/misc/changelog.rst
```diff
@@ -29,6 +29,7 @@ Bug Fixes:
 - Fixed missing verbose parameter passing in the ``EvalCallback`` constructor (@burakdmb)
 - Fixed the issue that when updating the target network in DQN, SAC, TD3, the ``running_mean`` and ``running_var`` properties of batch norm layers are not updated (@honglu2875)
 - Fixed incorrect type annotation of the replay_buffer_class argument in ``common.OffPolicyAlgorithm`` initializer, where an instance instead of a class was required (@Rocamonde)
+- Fixed the return type of ``.load()`` and ``.learn()`` methods in ``BaseAlgorithm`` so that they now use the ``Self`` type `PEP 0673 <https://peps.python.org/pep-0673/>`_. (@Rocamonde)
```
AdamGleave, line 32: This needs to be updated now that it uses TypeVar not self.

stable_baselines3/common/base_class.py
```diff
@@ -704,7 +708,10 @@ def load(
             get_system_info()
 
         data, params, pytorch_variables = load_from_zip_file(
-            path, device=device, custom_objects=custom_objects, print_system_info=print_system_info
```
AdamGleave, line 707: Was this auto-reformatted? (I don't object to the change just seems unrelated to the rest of the PR).

AdamGleave (commented):
Changelog needs to be updated and I agree with araffin can clean up the setup.py. Otherwise looks fine. Happy to merge once that's addressed and I or araffin LGTM.

stable_baselines3/common/base_class.py
Rocamonde: Yeah, I think vscode formatted it, then the linter didn't roll it back as there's some style flexibility. I actually think it's cleaner this way, so if it looks good to you too I would be happy to leave it.

Rocamonde:
Sounds good. Made those changes. Left the small format change in the `load_from_zip_file()` call as despite it being accidental I think it makes the call more readable now. Let me know your thoughts.

qgallouedec:
You also should remove the unused `import sys`

Rocamonde:
Sorry about that. Just did that now.

qgallouedec (approved):

stable_baselines3/common/base_class.py
```diff
@@ -2,10 +2,11 @@
 
 import io
 import pathlib
+import sys
```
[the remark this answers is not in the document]
Rocamonde: I think you might have reviewed an outdated version; already fixed these comments before :)

Rocamonde:
Was going to merge the approved PR but don't have write access.

qgallouedec:
> Was going to merge the approved PR but don't have write access.

Just wait for the checks to to completed, I'll merge it soon.

qgallouedec:
Should we do the same fix here?

https://github.com/DLR-RM/stable-baselines3/blob/18b29a68e8d5a11d0e98aeea539c247f0a913019/stable_baselines3/common/policies.py#L161

Rocamonde:
Sure, let's do that too.

qgallouedec:
Could you also correct the following lines? I think it is within the scope of this PR.

https://github.com/DLR-RM/stable-baselines3/blob/18b29a68e8d5a11d0e98aeea539c247f0a913019/stable_baselines3/common/off_policy_algorithm.py#L332

https://github.com/DLR-RM/stable-baselines3/blob/18b29a68e8d5a11d0e98aeea539c247f0a913019/stable_baselines3/common/on_policy_algorithm.py#L238

https://github.com/DLR-RM/stable-baselines3/blob/18b29a68e8d5a11d0e98aeea539c247f0a913019/stable_baselines3/a2c/a2c.py#L196

https://github.com/DLR-RM/stable-baselines3/blob/18b29a68e8d5a11d0e98aeea539c247f0a913019/stable_baselines3/ddpg/ddpg.py#L129

https://github.com/DLR-RM/stable-baselines3/blob/18b29a68e8d5a11d0e98aeea539c247f0a913019/stable_baselines3/dqn/dqn.py#L268

https://github.com/DLR-RM/stable-baselines3/blob/18b29a68e8d5a11d0e98aeea539c247f0a913019/stable_baselines3/ppo/ppo.py#L310

https://github.com/DLR-RM/stable-baselines3/blob/18b29a68e8d5a11d0e98aeea539c247f0a913019/stable_baselines3/sac/sac.py#L302

https://github.com/DLR-RM/stable-baselines3/blob/18b29a68e8d5a11d0e98aeea539c247f0a913019/stable_baselines3/td3/td3.py#L218

Rocamonde:
Added those changes.

qgallouedec (approved):

araffin:
@Rocamonde in case you have time in the coming weeks, could you do a similar PR to our contrib repo?
Otherwise, I will open an issue in that repo not to forget ;)

Rocamonde:
Hey! @araffin are you referring to https://github.com/hill-a/stable-baselines?

Rocamonde:
@araffin I suppose you're referring to the return types of specific algorithms? As I guess the base classes are shared with sb3.

What is a good way to search for all occurrences in the codebase?

araffin:
> I suppose you're referring to the return types of specific algorithms? 

yes, and I mean this repo: https://github.com/Stable-Baselines-Team/stable-baselines3-contrib
It follows the exact same structure as SB3.

> What is a good way to search for all occurrences in the codebase?

i think there will be only one occurence in each algorithm folder (for the learn method).

PR #1062, opened 2022-09-11
Transfer `ABC` inheritance from `BaseModel` to `BasePolicy`

qgallouedec:
Forgotten in my review of #1061

### DIFF
--- a/stable_baselines3/common/policies.py
+++ b/stable_baselines3/common/policies.py
@@ -34,7 +34,7 @@ from stable_baselines3.common.type_aliases import Schedule
 from stable_baselines3.common.utils import get_device, is_vectorized_observation, obs_as_tensor
 
 
-class BaseModel(nn.Module, ABC):
+class BaseModel(nn.Module):
     """
     The base model object: makes predictions in response to observations.
 
@@ -251,7 +251,7 @@ class BaseModel(nn.Module, ABC):
         return observation, vectorized_env
 
 
-class BasePolicy(BaseModel):
+class BasePolicy(BaseModel, ABC):
     """The base policy object.
 
     Parameters are mostly the same as `BaseModel`; additions are documented below.


stable_baselines3/common/policies.py
```diff
@@ -251,7 +251,7 @@ def obs_to_tensor(self, observation: Union[np.ndarray, Dict[str, np.ndarray]]) -
         return observation, vectorized_env
 
 
-class BasePolicy(BaseModel):
+class BasePolicy(BaseModel, ABC):
```
araffin, line 254: because it doesn't contain any unimplemented method anymore?

qgallouedec: Yes

araffin: well, it does contain unimplemented `forward()`...

araffin: but pytorch doesn't define it like that because of typing... we could just do `# noqa: B024` in that case.

qgallouedec: Inheriting `nn.Module` without implementing `forward` is a problem only if the method is intended to be instantiated (
[PEP 3119](https://peps.python.org/pep-3119/#the-abc-module-an-abc-support-framework)).
Since `BaseModel` is not intended to be instanciated, I think the right way is the following:


```python
from abc import ABC, abstractmethod


class Module:  # Doesn't inherit from ABC, from some reason
    def forward(self):
        raise NotImplementedError()


class BaseModel(Module):
    ...


class BasePolicy(BaseModel, ABC):
    @abstractmethod
    def _predict(self):
        ...

```

Note that even if we try to instantiate `BaseModel`, an error is raised. This error comes directly from torch and not from abc, but I think it is out of the scope of SB3.

qgallouedec: To clarify my previous post, even if `nn.Module` was an abstract class inheriting from `ABC`, it would be the same thing:

```python
from abc import ABC, abstractmethod


class Module(ABC):
    @abstractmethod
    def forward(self):
        ...


class BaseModel(Module):
    ...


class BasePolicy(BaseModel, ABC):
    @abstractmethod
    def _predict(self):
        ...
```

araffin: I guess removing `ABC` completely is not an option neither?

qgallouedec: We could find a sneaky way, like the one used in `nn.Module`. But what for?

araffin (approved):

Issue #925, opened 2022-05-29
[Feature Request] Remove `eval_env` parameter


araffin:
### 🚀 Feature

`eval_env` and `eval_freq` parameters were new in SB3 and were meant to facilitate evaluation in a separate env, by automatically creating evaluation env and `EvalCallback`.
However, they don't seem to be used that much and don't offer the flexibility of the `EvalCallback` at the cost of code duplication and maybe confusing parameters.

The idea would be to first deprecate (next SB3 version) and then remove all the parameters related to that feature.
Main drawback is that this would be a breaking change for people implementing custom algorithms.

### Checklist

- [x] I have checked that there is no similar [issue](https://github.com/DLR-RM/stable-baselines3/issues) in the repo (**required**)

tobirohrer:
Hey, 
I wanted to check if this change is still planned. If so, I would take a look at it in the coming days.

araffin:
Hello,
it is still planned, but it's not a priority.
I would like to integrate LN and dropout (https://github.com/DLR-RM/stable-baselines3/issues/1069), and probably progress bar callback first (https://github.com/DLR-RM/rl-baselines3-zoo/pull/287).
Actually, I would welcome more help to improve the RL Zoo.

However, you can of course work on such PR, please only deprecate the feature first (it should output a warning and remove the feature).

PR #1082, opened 2022-09-30
Deprecate eval_env and eval_freq parameter (#925)

tobirohrer:
## Description

- Added deprecation warning when using `eval_env` or `eval_freq`
- Added notice about deprecation in docs as well
See (#925 ) for more details

I have not updated the changelog yet, as there was a new version just released... 

## Motivation and Context

- [x] An issue was raised proposing this change ([required](https://github.com/DLR-RM/stable-baselines3/blob/master/CONTRIBUTING.md) for new features and bug fixes)

## Types of changes

## Checklist:


araffin:
Hello, thanks for the PR,
you somehow removed your changelog entry in https://github.com/DLR-RM/stable-baselines3/pull/1082/commits/822cadfdedf820503b6c39e2c20e138ae3780c2a ...
We probably need to update the tests too to catch some warnings.

tobirohrer:
Hey,

thanks for your feedback on my PR.
I see 3 options of how to handle the warnings about the deprecated parameters in the tests:

1. Directly catch and suppress the warnings in the code:
 ``` 
 ... Random test code ...
with warnings.catch_warnings():
        warnings.simplefilter("ignore", category=DeprecationWarning)
        model.learn(100, eval_env=eval_env)
 ```
  Disadvantage: We would need to add this to every test which uses one of the deprecated parameters.
  Disadvantage: Catches all `DeprecationWarning` of `model.learn()`
  Advantage: We could remove all but one warning, which would help spread the information about the deprecated feature.
 
 2. Remove usage of `eval_env` and `eval_freq` parameters from tests. As we want to deprecate those parameters anyway, we could remove them from the tests using `EvalCallback`. 
 
  (Possible) disadvantage: This might fit better in the upcoming PR, which removes those parameters.
  Advantage: We could remove all but one warning, which would help spread the information about the deprecated feature.
   
 3. Add `tests/pytest.ini` file which configures the tests to ignore our deprecation message via regex expression:
 ```
 # pytest.ini
 [pytest]
filterwarnings =
    ignore:Parameters `eval_env` and `eval_freq` are deprecated and will be removed in the future.
 ```
   (Possible) disadvantage: Added complexity by introducing a config file.
   (Possible) disadvantage: All warnings would be removed.
   
 --------------
 
 I would personally go for option 3 because of the ease of implementation. What do you think @araffin

qgallouedec:
I would go for option 1 with a slight modification:

```python
with pytest.warns(DeprecationWarning):
        model.learn(100, eval_env=eval_env)
```

> Disadvantage: We would need to add this to every test which uses one of the deprecated parameters.

Anyway, at the moment of definitively removing these parameters, it will be necessary to identify them. We had better do it now.

stable_baselines3/common/base_class.py
```diff
@@ -427,6 +428,15 @@ def _setup_learn(
         :param tb_log_name: the name of the run for tensorboard log
         :return:
         """
+
+        if eval_env is not None or eval_freq != -1:
+            warnings.warn(
+                "Parameters `eval_env` and `eval_freq` are deprecated and will be removed in the future."
+                " Please use `EvalCallback` or a custom Callback instead.",
```
qgallouedec, line 435:
```diff
- "[...] future."
- " Please [...]"
+ "[...] future. "
+ "Please [...]"
```

qgallouedec (changes requested):

araffin:
> I would go for option 1 with a slight modification:

I would also go for option 1, after a quick search there are 10 occurences of it it in the tests, mostly in the `test_run` via `create_eval_env=True`.

stable_baselines3/common/base_class.py
```diff
@@ -435,6 +436,15 @@ def _setup_learn(
         :param progress_bar: Display a progress bar using tqdm and rich.
         :return: Total timesteps and callback(s)
         """
+
+        if eval_env is not None or eval_freq != -1:
+            warnings.warn(
+                "Parameters `eval_env` and `eval_freq` are deprecated and will be removed in the future."
+                " Please use `EvalCallback` or a custom Callback instead.",
+                DeprecationWarning,
+                stacklevel=4,
```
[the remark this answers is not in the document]
tobirohrer: good catch. There are two reasons why I choose `stacklevel=4`
1. To refer to the warning where the deprecated feature was used (the call of `model.learn()`). Therefore, the warning, in the end, looks like this:
```
DeprecationWarning: Parameters `eval_env` and `eval_freq` are deprecated and will be removed in the future. Please use `EvalCallback` or a custom Callback instead.
  model.learn(total_timesteps=10000, eval_freq=100)
```

and not like this (if `stacklevel=1`, which is the default value):
```
Parameters `eval_env` and `eval_freq` are deprecated and will be removed in the future. Please use `EvalCallback` or a custom Callback instead.
  warnings.warn(
```
2. By default, warnings of the category `DeprecationWarning` are ignored, unless triggered by code in `__main__` (See [python docs](https://docs.python.org/3/library/warnings.html#warning-categories)). By setting `stacklevel=4`, we are referring to the initial call (`model.learn()`), which is mainly triggered by code in `__main__`. Therefore, we can show the warning without changing the behavior of the default warning filter.

Alternatively to setting `stacklevel=4`, we could change the default warning filter like (see [python docs](https://docs.python.org/3/library/warnings.html#overriding-the-default-filter)):
```
import sys

if not sys.warnoptions:
    import warnings
    warnings.simplefilter("ignore")
    warnings.warn('do the warning here`)
    warnings.simplefilter("default")
```

- An argument for using the `stacklevel` argument is: We don't need to change any default behaviour and ease of implementation.
- An argument against `stacklevel` argument is: The stacklevel is static. If the call does not refer to `model.learn()` "in level 4", or `model.learn()` is not called in `__main__` the warning is "swallowed"

In the end, we need to go for one of both if we want to show the warning when the parameter is used when using sb3 (In order to inform developers about the deprecated parameter). Otherwise, the warning is simply swallowed most of the time and only shown in rare cases e.g. when running the tests. 

Please let me know if I got something wrong or if I should change it from using the `stacklevel` parameter to changing the warning filter.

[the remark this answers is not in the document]
qgallouedec: Thank you for the detailed explanation. I agree with your current use of stack level.
Can you add a comment line so that we remember why `stack_level = 4`?

tobirohrer:
> > I would go for option 1 with a slight modification:
> 
> I would also go for option 1, after a quick search there are 10 occurences of it it in the tests, mostly in the `test_run` via `create_eval_env=True`.

Just for clarification: The ultimate goal is to remove the automatic creation of the evaluation procedure in the `BaseAlgorithm` class and to facilitate the usage of `EvalCallback` and `Callback` directly. Right? 

Because then It would make sense to deprecate the `create_eval_env` parameter as well. (I did so in  [my last commit](https://github.com/DLR-RM/stable-baselines3/pull/1082/commits/d3ffe1287fadf877fdeb2f16830e86e124f10f70))

stable_baselines3/common/base_class.py
```diff
@@ -161,6 +163,15 @@ def __init__(
...
+                    warnings.warn(
+                        "The parameter `create_eval_env` is deprecated and will be removed in the future. "
+                        "Please use `EvalCallback` or a custom Callback instead.",
+                        DeprecationWarning,
+                        # By setting the `stacklevel` we refer to the initial caller of the deprecated feature.
+                        # This causes the the `DepricationWarning` to not be ignored and to be shown to the user. See
+                        # https://github.com/DLR-RM/stable-baselines3/pull/1082#discussion_r989842855 for more details.
+                        stacklevel=4,
+                    )
                     self.eval_env = maybe_make_env(env, self.verbose)
```
qgallouedec, line 175: From https://github.com/DLR-RM/stable-baselines3/issues/925#issuecomment-1251488507, I understand that this feature should be removed. Shouldn't this line be deleted?

qgallouedec (changes requested):

stable_baselines3/common/base_class.py
araffin: > Shouldn't this line be deleted?

only after the next release, no?
First deprecate and then remove?

tobirohrer: I understood that we would first deprecate the parameters and then remove them in a subsequent release (As it would be breaking changes for those who use them).

stable_baselines3/common/base_class.py
```diff
@@ -567,8 +590,10 @@ def learn(
         :param callback: callback(s) called at every step with state of the algorithm.
         :param log_interval: The number of timesteps before logging.
         :param tb_log_name: the name of the run for TensorBoard logging
-        :param eval_env: Environment that will be used to evaluate the agent
-        :param eval_freq: Evaluate the agent every ``eval_freq`` timesteps (this may vary a little)
+        :param eval_env: Environment that will be used to evaluate the agent. Caution, this parameter
+            is deprecated and will be removed in the future. Please use ``EvalCallback`` instead.
+        :param eval_freq: Evaluate the agent every ``eval_freq`` timesteps (this may vary a little).
+            Caution, this parameter is deprecated and will be removed in the future.
```
qgallouedec, line 596:
```diff
+ Please use `EvalCallback` or a custom Callback instead.
```

stable_baselines3/ddpg/ddpg.py
```diff
@@ -45,7 +45,8 @@ class DDPG(TD3):
         at a cost of more complexity.
         See https://github.com/DLR-RM/stable-baselines3/issues/37#issuecomment-637501195
     :param create_eval_env: Whether to create a second environment that will be
-        used for evaluating the agent periodically. (Only available when passing string for the environment)
+        used for evaluating the agent periodically (Only available when passing string for the environment).
+        Caution, this parameter is deprecated and will be removed in the future.
```
qgallouedec, line 49:
```diff
+ Please use `EvalCallback` or a custom Callback instead.
```

stable_baselines3/dqn/dqn.py
```diff
@@ -53,7 +53,8 @@ class DQN(OffPolicyAlgorithm):
     :param max_grad_norm: The maximum value for the gradient clipping
     :param tensorboard_log: the log location for tensorboard (if None, no logging)
     :param create_eval_env: Whether to create a second environment that will be
-        used for evaluating the agent periodically. (Only available when passing string for the environment)
+        used for evaluating the agent periodically (Only available when passing string for the environment).
+        Caution, this parameter is deprecated and will be removed in the future.
```
qgallouedec, line 57:
```diff
+ Please use `EvalCallback` or a custom Callback instead.
```

stable_baselines3/ppo/ppo.py
```diff
@@ -58,7 +58,8 @@ class PPO(OnPolicyAlgorithm):
         By default, there is no limit on the kl div.
     :param tensorboard_log: the log location for tensorboard (if None, no logging)
     :param create_eval_env: Whether to create a second environment that will be
-        used for evaluating the agent periodically. (Only available when passing string for the environment)
+        used for evaluating the agent periodically (Only available when passing string for the environment).
+        Caution, this parameter is deprecated and will be removed in the future.
```
qgallouedec, line 62:
```diff
+ Please use `EvalCallback` or a custom Callback instead.
```

stable_baselines3/sac/sac.py
```diff
@@ -66,7 +66,8 @@ class SAC(OffPolicyAlgorithm):
     :param use_sde_at_warmup: Whether to use gSDE instead of uniform sampling
         during the warm up phase (before learning starts)
     :param create_eval_env: Whether to create a second environment that will be
-        used for evaluating the agent periodically. (Only available when passing string for the environment)
+        used for evaluating the agent periodically (Only available when passing string for the environment).
+        Caution, this parameter is deprecated and will be removed in the future.
```
qgallouedec, line 70:
```diff
+ Please use `EvalCallback` or a custom Callback instead.
```

stable_baselines3/td3/td3.py
```diff
@@ -54,7 +54,8 @@ class TD3(OffPolicyAlgorithm):
         (smoothing noise)
     :param target_noise_clip: Limit for absolute value of target policy smoothing noise.
     :param create_eval_env: Whether to create a second environment that will be
-        used for evaluating the agent periodically. (Only available when passing string for the environment)
+        used for evaluating the agent periodically (Only available when passing string for the environment).
+        Caution, this parameter is deprecated and will be removed in the future.
```
qgallouedec, line 58:
```diff
+ Please use `EvalCallback` or a custom Callback instead.
```

qgallouedec (changes requested):
Could you also add the caution about deprecation in the docstring of all concerned methods?
Not sure if it's exaustive:

https://github.com/tobirohrer/stable-baselines3/blob/cdf79a4ba60c53a973d2da5923bf8d491db3945c/stable_baselines3/common/base_class.py#L439

https://github.com/tobirohrer/stable-baselines3/blob/cdf79a4ba60c53a973d2da5923bf8d491db3945c/stable_baselines3/common/off_policy_algorithm.py#L63

https://github.com/tobirohrer/stable-baselines3/blob/cdf79a4ba60c53a973d2da5923bf8d491db3945c/stable_baselines3/common/on_policy_algorithm.py#L41

stable_baselines3/common/base_class.py
qgallouedec: ok for me

qgallouedec (approved):
I think we're good now. I don't think we need any changes on rl3-zoo. But there is the documentation of sb3-contrib to update.

Issue #1055, opened 2022-09-09
[Feature Request] different activation functions in the network_architecture through the policy_kwargs


AlexPasqua:
### 🚀 Feature
Introduce the possibility of passing multiple activation functions to the policy newtork using the `policy_kwargs`.

### Motivation
From what I understand, through the `policy_kwargs` it is possible to pass an activation function to be used by the `net_arch` part of the policy network. Oftentimes, though, the policy net (`pi`) and value function net (`vf`) need different activation functions.
It looks like the only way to have different activation functions in these two sub-networks is to implement our own policy network, as shown in the advanced example [here](https://stable-baselines3.readthedocs.io/en/master/guide/custom_policy.html#advanced-example) in your documentation. This is mentioned also in this issue #481 

### Alternatives

Ideally it would be possible to have multiple activation functions as follows: one for the shared layers and one for each of the layers of the two sub-networks (policy net (`pi`) and value net (`vf`)), mimicking how the architecture is passed.
The architecture is passed this way: `[<shared layers>, dict(vf=[<non-shared value network layers>], pi=[<non-shared policy network layers>])]` (source: [here](https://stable-baselines3.readthedocs.io/en/master/guide/custom_policy.html#on-policy-algorithms)), so I think it would be possible to use the same structure, but using PyTorch's activation functions instead of integers.

Example:
```
from torch.nn import ReLU, Softmax, Tanh

model = A2C('MultiInputPolicy', env,
             policy_kwargs=dict(
                 net_arch=[256, dict(pi=[128, 50], vf=[32, 1])],
                 activation_fn=[Tanh, dict(pi=[ReLU, Softmax], vf=[ReLU, ReLU])]
             )
        )
```

araffin:
Hello,
sounds reasonable (even though I doubt  changing the activation per layer will make a big difference).
Could you do a draft PR to see how much complexity it adds? (we would have to add that feature for off-policy algorithms too for consistency)

> Oftentimes, though, the policy net (pi) and value function net (vf) need different activation functions.

do you have references for that?

AlexPasqua:
> Could you do a draft PR to see how much complexity it adds? (we would have to add that feature for off-policy algorithms too for consistency)

Initially I'll need to create my own policy network, because I need to implement this thing in a project pretty quickly, but I'll start to work on this draft PR in some days.

> > Oftentimes, though, the policy net (pi) and value function net (vf) need different activation functions.
> 
> do you have references for that?

[...]

PR #1116, opened 2022-10-14
Added support for different activ funcs in MLPExtractor layers

AlexPasqua:
Made possible to have different activation functions in each layer of the `MLPExtractor` through the `policy_kwargs` when using `ActorCriticPolicy`.

## Description

Modified `ActorCriticPolicy` to accept the activation function in this way: not `Type[nn.Module]` anymore, but `List[Union[Type[nn.Module], Dict[str, List[Type[nn.Module]]]]]`.
Then modified the `MLPExtractor` class so it creates not only lists of layers for the shared / policy only / value only, but also similar lists of activation functions. Then, after "appending" a layer in the `policy_net` and `value_net`, **the corresponding** activation function is "appended" as well. This is done iterating not only on the policy and value layers (`zip_longest(policy_only_layers, value_only_layers)`) and using always the same type of activation, but iterating also on the policy and value activation functions (`zip_longest(policy_only_layers, value_only_layers, policy_only_activs, value_only_activs)`).

After creating the `policy_net` and `value_net` lists this way, everything interfaces correctly and the model is created correctly.

## Question

Why is it never checked that the activation is a `nn.Module`?
I see that there are checks for the layer to be an `int`, but not the corresponding ones for the activation.

## Motivation and Context

In agreement with what discussed in issue #1055 

## Types of changes

## Checklist:

(**I mean the docstring**)


qgallouedec:
I'm afraid that it makes the code quite complex without necessity. For example, does this code remain working?

```python
from torch import nn

policy_kwargs = dict(activation_fn=nn.ELU)

model = PPO("MlpPolicy", "CartPole-v1", policy_kwargs=policy_kwargs)
model.learn(total_timesteps=10_000)
```

AlexPasqua:
> I'm afraid that it makes the code quite complex without necessity. For example, does this code remain working?
> 
> ```python
> from torch import nn
> 
> from stable_baselines3 import PPO
> 
> policy_kwargs = dict(activation_fn=nn.ELU)
> 
> model = PPO("MlpPolicy", "CartPole-v1", policy_kwargs=policy_kwargs)
> model.learn(total_timesteps=10_000)
> ```

Not yet, but I plan to make it compatible, it should be pretty easy.

(I wanted to make a draft PR, but I just noticed it's an actual PR).

araffin:
Thanks for the PR, now we have a better overview of the required changes =)

I think I agree with @qgallouedec (and that was my fear in https://github.com/DLR-RM/stable-baselines3/issues/1055#issuecomment-1243034700) and I'm afraid the added complexity is not worth the benefits: there is no study/benchmark that using different activation per layer or per network gives much better results than using the same activation for all layers.

I'm obviously happy to change my mind if there is evidence of the opposite in the future ;)

AlexPasqua:
> Thanks for the PR, now we have a better overview of the required changes =)

Anyway I think I didn't explain the feature very well in detail -> now it's be possible to pass the activation function in basically 3 ways:

### Option 1. As done previously
E.g.: `activation_fn=nn.Tanh` --> same activation function for every layer

### Option 2. Specify once for actor and once for critic
In this case you specify the activation function layer by layer for what concerns the shared net, while for the actor and the critic you simply pass one function each, which will be used in every one of their non-shared layers.

E.g.
```python
net_arch=[128, 64, dict(pi=[32, 32], vf=[32, 16, 8])
```
The corresponding `activation_fn` will have the same structure, but only one activation function is specified for `pi` and one for `vf`
```python
activation_fn=[nn.Tanh, nn.Tanh, dict(pi=nn.Sigmoid, vf=nn.ReLU)
```
The sigmoid will be applied to the 2 `pi` layers and the ReLU to the 3 `vf` layers.

### Option 3. Specify layer by layer
If a more fine-grained management is necessary, it's possible to specify the activaiton function layer by layer, copying the structure of `net_arch`.

E.g.
```python
net_arch=[128, 64, dict(pi=[32, 32], vf=[32, 16, 8]),
activation_fn=[nn.Tanh, nn.Tanh, dict(pi=[nn.Sigmoid, nn.ReLU], vf=[nn.Sigmoid, nn.Sigmoid, nn.ReLU])
```

---

I think this would offer a lot of flexibility to developers, who can rely on the library's APIs even for more custom implementations, without the necessity to re-implement the whole network from scratch.

> there is no study/benchmark that using different activation per layer or per network gives much better results than using the same activation for all layers.
> 
> I'm obviously happy to change my mind if there is evidence of the opposite in the future ;)

Well, I cannot search for all the possible examples, but in my case I'm following the work of the authors of [this paper](https://ieeexplore.ieee.org/document/9632824) and they tend to use two different activation functions in the actor and in the critic networks.
I had to do it as well and I had to implement everything from scratch. I felt like it was a limit of the library, because it looked like a simple feature, but that offers a lot more freedom in the design of the model.

---

### Code complexity
To me it doesn't look very complex, in the end the constructor of the `MlpExtractor` simply creates a list like the one of `net_arch` and use that one to build the layers, but anyway, if we want to simplify, we could get rid of (or modify) option 2 or 3.

Option 3 is the one which offers the most flexibility, so I'd keep it. It's the most verbose too, but it'd be used less frequently than the other alternatives, so I think that if a developer really needs a finge-grained control, it'd be worth the extra verbosity.

Option 2 lays in between options 1 and 3: it's slightly less flexible than option 3, but also less verbose. I'd personally get rid of that one instead of option 3. This should simplify a bit the code.

Otherwise, instead of removing one of the options, I could modify option 2 so that the developers only specify one activation function for the shared net, one for the policy net and one for the value net, regardless of their number of layers (currently, for the shared net it's necessary to specify an activation function for each layer). This would reduce the flexibility, but option 3 is always there for it.

Furthermore I could comment the code a bit more or reorganize it, for example putting the following block in its own method, in order not to interrupt the readability flow of the constructor's code.
```python
if not isinstance(activation_fn, list):
    act_fn_list = []
    for i in range(len(net_arch)):
        if isinstance(net_arch[i], int):
            act_fn_list.append(activation_fn)
        elif isinstance(net_arch[i], dict):
            act_fn_list.append(dict())
            for key, dims in net_arch[i].items():
                act_fn_list[-1][key] = [activation_fn] * len(dims)
    activation_fn = act_fn_list
```

---

Let me know what you think

araffin:
> I'm following the work of the authors of [this paper](https://ieeexplore.ieee.org/document/9632824) and they tend to use two different activation functions in the actor and in the critic networks.

I think we discussed that already in https://github.com/DLR-RM/stable-baselines3/issues/1055#issuecomment-1243835247
They are not using different activation function, there are using softmax at the end because they need to sample from a categorical distribution.
SB3 already does that behind the scene.

> I think this would offer a lot of flexibility to developers

I know it offers more flexibility but as in https://github.com/DLR-RM/stable-baselines3/pull/1148#issuecomment-1299747968, flexibility doesn't make sense if it doesn't provide any significant gains.

AlexPasqua:
> > I'm following the work of the authors of [this paper](https://ieeexplore.ieee.org/document/9632824) and they tend to use two different activation functions in the actor and in the critic networks.
> 
> I think we discussed that already in [#1055 (comment)](https://github.com/DLR-RM/stable-baselines3/issues/1055#issuecomment-1243835247) They are not using different activation function, there are using softmax at the end because they need to sample from a categorical distribution. SB3 already does that behind the scene.
 
I was referring to the fact that they have completely separate actor and critic (so no weight sharing whatsoever), and in the actor they use a tanh activation function, while in the critic the opt for a ReLU.

> I know it offers more flexibility but as in [#1148 (comment)](https://github.com/DLR-RM/stable-baselines3/pull/1148#issuecomment-1299747968), flexibility doesn't make sense if it doesn't provide any significant gains.

Alright, if you think it's not worth it, let's leave it as it is

araffin:
>  and in the actor they use a tanh activation function, while in the critic the opt for a ReLU.

oh I see, I didn't see any ablation study about that choice in the paper, or did they do one?

Actually, https://github.com/DLR-RM/stable-baselines3/pull/1148#issuecomment-1299747968 is motivated towards making separate custom actor/critic more accessible (currently, because we allow the part of the network to be shared, it is not trivial to implement a custom policy).

AlexPasqua:
> > and in the actor they use a tanh activation function, while in the critic the opt for a ReLU.
> 
> oh I see, but I didn't see any ablation study about that choice in the paper, or did they do one?

Not that I'm aware of.

> Actually, [#1148 (comment)](https://github.com/DLR-RM/stable-baselines3/pull/1148#issuecomment-1299747968) is motivated towards making separate custom actor/critic more accessible (currently, because we allow the part of the network to be shared, it is not trivial to implement a custom policy).

Alright, so in case we might re-consider the idea of a different activation function for actor and critic in case this "more accessible separate custom actor/critic" gets implemented. Obviously, if it's in your interest and if it doesn't introduce complexity when implemented in the "new" code.

araffin:
Closing in favor of https://github.com/DLR-RM/stable-baselines3/pull/1292

PR #1143, opened 2022-10-27
Mypy type checking

qgallouedec:
## Motivation and Context

## Types of changes

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@ pytest:
 
 type:
 	pytype -j auto
+	mypy ${LINT_PATHS}
 
 lint:
 	# stop the build if there are Python syntax errors or undefined names
--- a/setup.cfg
+++ b/setup.cfg
@@ -25,6 +25,58 @@ markers =
 inputs = stable_baselines3
 disable = pyi-error
 
+[mypy]
+ignore_missing_imports = True
+exclude = (?x)(
+    stable_baselines3/a2c/a2c.py
+	| stable_baselines3/common/atari_wrappers.py
+	| stable_baselines3/common/base_class.py
+    | stable_baselines3/common/buffers.py
+    | stable_baselines3/common/callbacks.py
+    | stable_baselines3/common/distributions.py
+    | stable_baselines3/common/env_util.py
+    | stable_baselines3/common/envs/bit_flipping_env.py
+    | stable_baselines3/common/envs/identity_env.py
+    | stable_baselines3/common/envs/multi_input_envs.py
+    | stable_baselines3/common/logger.py
+    | stable_baselines3/common/monitor.py
+    | stable_baselines3/common/off_policy_algorithm.py
+    | stable_baselines3/common/on_policy_algorithm.py
+    | stable_baselines3/common/policies.py
+    | stable_baselines3/common/preprocessing.py
+    | stable_baselines3/common/save_util.py
+    | stable_baselines3/common/sb2_compat/rmsprop_tf_like.py
+    | stable_baselines3/common/torch_layers.py
+    | stable_baselines3/common/type_aliases.py
+    | stable_baselines3/common/utils.py
+    | stable_baselines3/common/vec_env/__init__.py
+    | stable_baselines3/common/vec_env/base_vec_env.py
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -6,6 +6,7 @@ pytest:
 
 type:
 	pytype -j auto
+	mypy ${LINT_PATHS}
```
araffin, line 9: I would also add a `mypy` entry (would make debugging easier as pytype is quite slow as soon as you change one file)

araffin: I would also activate the color output as we do for pytest, see https://github.com/python/mypy/issues/7771

qgallouedec: > I would also add a `mypy` entry (would make debugging easier as pytype is quite slow as soon as you change one file)

Should I replace `type` by `pytype` then?

araffin: type is still fine as we want to run both in the CI or when developping.

qgallouedec:
I need help here. I can't understand how to ignore files in `setup.cfg`. The current config seems to work locally, but sometimes it doesn't. I've failed so far to get a reproductible procedure.

Rocamonde:
What we do in imitation is we define a variable called `EXCLUDE_MYPY` 
https://github.com/HumanCompatibleAI/imitation/blob/d6f8ca2748981deecd5b660b3c665ae98c92b56b/ci/code_checks.sh#L5
then we pass this as an arg when we run mypy
https://github.com/HumanCompatibleAI/imitation/blob/d6f8ca2748981deecd5b660b3c665ae98c92b56b/ci/code_checks.sh#L37
crucially you need to pass `--follow-imports=silent --show-error-codes` otherwise it will follow imports and still raise errors for the files that you're ignoring.

setup.cfg
```diff
@@ -28,53 +28,53 @@ disable = pyi-error
...
-    | stable_baselines3/sac/sac.py
-    | stable_baselines3/td3/policies.py
-    | stable_baselines3/td3/td3.py
-    | tests/test_distributions.py
-    | tests/test_logger.py
-    | tests/test_tensorboard.py
-    | tests/test_train_eval_mode.py
-    | tests/test_vec_normalize.py
+    stable_baselines3/a2c/a2c.py$
+	| stable_baselines3/common/atari_wrappers.py$
```
Rocamonde, line 33: Do you nee the indentation?

qgallouedec: thank you for reporting

qgallouedec:
> crucially you need to pass --follow-imports=silent otherwise it will follow imports and still raise errors for the files that you're ignoring

It works! Thanks!

Makefile
qgallouedec: > I would also activate the color output as we do for pytest, see [python/mypy#7771](https://github.com/python/mypy/issues/7771)

I'm not sure to use `MYPY_FORCE_COLOR=1` properly, it does not seem to work as expected.

araffin: yep, I saw that :/ will see if I can find an example online

araffin (approved):
From my part, I don't have much to say, LGTM =)

but I will let @Rocamonde comment on it and he is the one experienced with such transition.

setup.cfg
```diff
@@ -25,6 +25,60 @@ markers =
 inputs = stable_baselines3
 disable = pyi-error
 
+[mypy]
```
araffin, line 28: did you had to add each file manually :see_no_evil: ?

qgallouedec: Worse than that. First I added all the files, then for each ignored file I checked if removing it from the ignore list really produced a mypy error.
And I deleted the ones that didn't cause an error. 😮‍💨

Makefile
qgallouedec: What about this @araffin?

setup.cfg
araffin: oh boy... thanks for the effort =)

Makefile
araffin: I double check and you seem to be doing things right...

araffin:
I will merge this as it prevents from fixing type annotations.
@Rocamonde if you have time in the coming days, feel free to do a "post-review".

For the color issue, it's minor so let's fix it later if we find out what we are doing wrong.

PR #1146, opened 2022-10-28
Update evaluate_policy type annotation to support policies as well as RL algorithms

AdamGleave:
## Description
Currently `evaluate_policy` expects a `model: BaseAlgorithm` i.e. RL algorithm. But it runs fine on any `BasePolicy`, as it just calls `model.predict()` which has an identical type signature between `BaseAlgorithm` and `BasePolicy` -- indeed, `BaseAlgorithm` just forwards it to the policy contained inside it. This PR introduces a `PolicyPredictor` protocol for anything that matches that `predict` type signature, and updates `evaluate_policy`'s type annotation to use that.

I ended up putting `PolicyPredictor` in `type_aliases.py` but I can see an argument for putting it in `policies.py`.

`Protocol` is only available Python >=3.8 so I had to use `typing_extensions` but per discussion in https://github.com/DLR-RM/stable-baselines3/pull/1043#discussion_r961688423 we already have this installed transitively via `pytorch` so this is not a new dependency, although I've added it to `setup.py` to make it explicit

## Motivation and Context
`evaluate_policy` can already be used with policies and this is a fairly common use case. So currently code is failing to type check that works fine at runtime. This PR allows that code to type check.

## Types of changes

## Checklist:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -76,7 +76,7 @@ setup(
         "gym==0.21",  # Fixed version due to breaking changes in 0.22
         "numpy",
         "torch>=1.11",
-        "typing_extensions",
+        "typing_extensions>=4.0,<5",
         # For saving models
         "cloudpickle",
         # For reading logs
--- a/setup.py
+++ b/setup.py
@@ -76,6 +76,7 @@ setup(
         "gym==0.21",  # Fixed version due to breaking changes in 0.22
         "numpy",
         "torch>=1.11",
+        "typing_extensions",
         # For saving models
         "cloudpickle",
         # For reading logs
--- a/setup.py
+++ b/setup.py
@@ -76,7 +76,7 @@ setup(
         "gym==0.21",  # Fixed version due to breaking changes in 0.22
         "numpy",
         "torch>=1.11",
-        "typing_extensions>=4.0,<5",
+        'typing_extensions>=4.0,<5; python_version < "3.8.0"',
         # For saving models
         "cloudpickle",
         # For reading logs
--- a/stable_baselines3/common/evaluation.py
+++ b/stable_baselines3/common/evaluation.py
@@ -4,12 +4,12 @@ from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 import gym
 import numpy as np
 
-from stable_baselines3.common import base_class, policies
+from stable_baselines3.common import base_class, type_aliases
 from stable_baselines3.common.vec_env import DummyVecEnv, VecEnv, VecMonitor, is_vecenv_wrapped
 
 
 def evaluate_policy(
-    model: policies.PolicyPredictor,
+    model: "type_aliases.PolicyPredictor",
     env: Union[gym.Env, VecEnv],
     n_eval_episodes: int = 10,
     deterministic: bool = True,
--- a/stable_baselines3/common/evaluation.py
+++ b/stable_baselines3/common/evaluation.py
@@ -4,12 +4,12 @@ from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 import gym
 import numpy as np
 
-from stable_baselines3.common import base_class
+from stable_baselines3.common import base_class, policies
 from stable_baselines3.common.vec_env import DummyVecEnv, VecEnv, VecMonitor, is_vecenv_wrapped
 
 
 def evaluate_policy(
-    model: "base_class.BaseAlgorithm",
+    model: policies.PolicyPredictor,
     env: Union[gym.Env, VecEnv],
     n_eval_episodes: int = 10,
     deterministic: bool = True,
@@ -34,7 +34,9 @@ def evaluate_policy(
         results as well. You can avoid this by wrapping environment with ``Monitor``
         wrapper before anything else.
 
-    :param model: The RL agent you want to evaluate.
+    :param model: The RL agent you want to evaluate. This can be any object
+        that implements a `predict` method, such as an RL algorithm (``BaseAlgorithm``)
+        or policy (``BasePolicy``).
     :param env: The gym environment or ``VecEnv`` environment.
     :param n_eval_episodes: Number of episode to evaluate the agent
     :param deterministic: Whether to use deterministic or stochastic actions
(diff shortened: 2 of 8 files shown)


setup.py
```diff
@@ -76,6 +76,7 @@
         "gym==0.21",  # Fixed version due to breaking changes in 0.22
         "numpy",
         "torch>=1.11",
+        "typing_extensions",
```
araffin, line 79: there should a max python version here, no?

stable_baselines3/common/type_aliases.py
```diff
@@ -69,3 +70,26 @@ class TrainFrequencyUnit(Enum):
 class TrainFreq(NamedTuple):
     frequency: int
     unit: TrainFrequencyUnit  # either "step" or "episode"
+
+
+class PolicyPredictor(Protocol):
```
araffin, line 75: nice, I guess protocol is the equivalent of "Trait" in Rust?
(shared behavior)

araffin (commented):
Apart from minor remark on requirements, LGTM =)
(I think we should probably use protocols more)

qgallouedec (changes requested):

stable_baselines3/common/evaluation.py
```diff
@@ -4,12 +4,12 @@
 import gym
 import numpy as np
 
-from stable_baselines3.common import base_class
+from stable_baselines3.common import type_aliases
```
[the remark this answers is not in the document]
araffin: i think it is to avoid circular import and doc build failure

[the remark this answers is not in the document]
qgallouedec: I don’t know about doc built, but it does not prevent circular import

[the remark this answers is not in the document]
qgallouedec: Actually, you're right, it does prevent circular import.

stable_baselines3/common/type_aliases.py
AdamGleave: Yep, it's similar! Defined in https://peps.python.org/pep-0544/#protocol-members

setup.py
AdamGleave: Oh, I perhaps misunderstood, are you suggesting we change this to e.g:
```suggestion
    'typing_extensions; python_version < "3.8.0"',
```
Right now that would break things as we always import from `typing_extensions`; it would work if I also made the change @qgallouedec suggested of gating the import by Python version.

However, right now Torch unconditionally imports on `typing_extensions` and they don't seem inclined to change this, so we will install it either way: https://github.com/pytorch/vision/issues/5942

stable_baselines3/common/type_aliases.py
```diff
@@ -1,11 +1,12 @@
 """Common aliases for type hints"""
 
 from enum import Enum
-from typing import Any, Callable, Dict, List, NamedTuple, Tuple, Union
+from typing import Any, Callable, Dict, List, NamedTuple, Optional, Tuple, Union
 
 import gym
 import numpy as np
 import torch as th
+from typing_extensions import Protocol
```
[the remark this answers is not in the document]
AdamGleave: I'm OK making this change but I don't understand the motivation. Is there any harm from using the `Protocol` version from `typing_extensions` even on Python versions where it's supported? The nice thing here is it self-documents when we can remove `typing_extensions`, but it does make the code a bit harder to understand.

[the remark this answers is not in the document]
qgallouedec: > Is there any harm from using the Protocol version from typing_extensions even on Python versions where it's supported

AFAIK, there isn't any.

> I don't understand the motivation

I have not found a strict recommendation on this matter. The main reasons that make me prefer this syntax are

1.
> The nice thing here is it self-documents when we can remove `typing_extensions`

2. This seems to be the common usage: although it is not uniform I found several occurrences in `torch` and `numpy`

3. For newer Python versions, `Protocol` is not an extension, so it tickles me semantically to import it from `typing-extensions`.

setup.py
araffin: sorry, yes i meant python version, but if pytorch do so, then don't put any version bound (otherwise it might clash with PyTorch in the future)

AdamGleave: I've added Python version range for consistency with [qgallouedec](https://github.com/qgallouedec)'s suggestion. I don't think this should cause any clash with PyTorch as if Python version is new enough it's as if we just don't have the requirement.

qgallouedec:
While we're at it, let's fix the observation type annotation in predict: [d36750c](https://github.com/DLR-RM/stable-baselines3/pull/1146/commits/d36750c5f95abe695e104a647f6d3f2d6e6e2651)

stable_baselines3/common/evaluation.py
```diff
@@ -4,12 +4,12 @@
...
 from stable_baselines3.common.vec_env import DummyVecEnv, VecEnv, VecMonitor, is_vecenv_wrapped
 
 
 def evaluate_policy(
-    model: "base_class.BaseAlgorithm",
+    model: "type_aliases.PolicyPredictor",
```
qgallouedec, line 12: What do you think about that? This would allow not to break the typing consistency for the variable named "model".

```suggestion
    model_or_policy: "type_aliases.PolicyPredictor",
```

Rocamonde: Maybe make this positional-only using `/`.

araffin: not a big fan of `model_or_policy`, we could use `predictor` but I don't like much neither...

> Maybe make this positional-only using /.

why not, but unsure if needed

qgallouedec: > Maybe make this positional-only using `/`.

I don't think this is a good solution. My main concern here is that the term "model" refers everywhere else to an instance of BaseAlgorithm. Making it only positional does not solve the problem.
But if you think it's okay, we can leave it like that.

araffin: > My main concern here is that the term "model" refers everywhere else to an instance of BaseAlgorithm

I understand your point, and in practice, it is the case in all examples except one in the doc (when we show the use of `model.policy`).
So, I would say it's fine.

araffin (approved):
LGTM, thanks =)

qgallouedec (approved):

Issue #1066, opened 2022-09-13
[Question] Not shared features extractor in on-policy algorithm


AlexPasqua:
### Question
I've checked [the docs (custom policy -> advanced example)](https://stable-baselines3.readthedocs.io/en/master/guide/custom_policy.html#advanced-example), but it is not clear to me how to create a custom policy without sharing the features extractor between the actor and the critic networks in on-policy algorithms.

If I pass a `features_extractor_class` in the `policy_kwargs`, this is shared by default I think.

I can have a non-shared `mlp_extractor` by implementing my own `_build_mlp_extractor` method in my custom policy and creating a network with 2 distinct sub-networks (`self.policy_net` and `self.value_net`), but I didn't understand how to do the same with the features extractor.

On [the docs (custom policy -> custom features extractor)](https://stable-baselines3.readthedocs.io/en/master/guide/custom_policy.html#custom-feature-extractor), it says:
![image](https://user-images.githubusercontent.com/33461106/190012676-93cfc792-d439-46bc-9993-565b7447520a.png)
Therefore, since I'm using A2C, I think it should be possible to have a non-shared features extractor by implementing my own policy, just I didn't understand how to do it.

Thanks in advance any clarification!

### Checklist

- [x] I have read the [documentation](https://stable-baselines3.readthedocs.io/en/master/) (**required**)
- [x] I have checked that there is no similar [issue](https://github.com/DLR-RM/stable-baselines3/issues) in the repo (**required**)

AlexPasqua:
I managed to make it run without errors! :tada:

But since I haven't found a guide/demo nor a similar issue here, I'll briefly explain **how I did it**:
1. Create a custom policy (in my case, subclassing from `ActorCriticPolicy`).
2. In this custom policy, I created 2 attributes, one for the features extractor of the actor and one for the features extractor of the critic.
3. Override the super-policy's methods that use the feature extractor, in order to use yours. In my case they were: `forward`, `extract_features`, `evaluate_actions` and `predict_values`).

### Quick demo
```python
class MyFeaturesExtractor(BaseFeaturesExtractor):
    ...


class MyMLPExtractor(nn.Module):
    # not mandatory, just in my case I had also a custom MLP extractor
    # because I had to use 2 different activation functions for the actor and the critic
    ...


class MyPolicy(ActorCriticPolicy):
    def __init__(
            self,
            observation_space: gym.spaces.Space,
            action_space: gym.spaces.Space,
[...]

PR #1148, opened 2022-11-02
Modified ActorCriticPolicy to support non-shared features extractor

AlexPasqua:
According to #1066 ([this comment](https://github.com/DLR-RM/stable-baselines3/issues/1066#issuecomment-1250030864) in particular).

## Description

Added a parameter `shared` to the `features_extractor_kwargs` (in the `policy_kwargs`).

If `shared` is false, the policy will create 2 instances of the same class of features extractor, called `pi_features_extractor` and `vf_features_extractor`.

The method `extract_features` has been overridden:
- in case `shared` is true, it simply calls the superclass' method;
- if it's false, it extracts the `pi_features` using the `pi_features_extractor` and the `vf_features` with the `vf_feratures_extractor`.

All other methods that were containing a features extraction (e.g. `forward`, `evaluate_actions`, `get_distribution`, `predict_values`) have been adapted to work with 2 sets of features in case `shared` is false: the `pi_features` and `vf_features` extracted by the method `extract_features` as just explained.

## Motivation and Context

According to #1066 ([this comment](https://github.com/DLR-RM/stable-baselines3/issues/1066#issuecomment-1250030864) in particular).

## Types of changes

## Checklist:


stable_baselines3/common/policies.py
```diff
@@ -443,8 +443,17 @@ def __init__(
         self.activation_fn = activation_fn
         self.ortho_init = ortho_init
 
+        self.share_features_extractor = self.features_extractor_kwargs.get("shared", True)
```
araffin, line 446: I would rather add a parameter to the class, even though it means updating all classes.

stable_baselines3/common/policies.py
```diff
@@ -617,14 +660,19 @@ def evaluate_actions(self, obs: th.Tensor, actions: th.Tensor) -> Tuple[th.Tenso
...
-        :param actions:
+        :param obs: Observation
+        :param actions: Actions
         :return: estimated value, log likelihood of taking those actions
             and entropy of the action distribution.
         """
         # Preprocess the observation if needed
         features = self.extract_features(obs)
-        latent_pi, latent_vf = self.mlp_extractor(features)
+        if not self.share_features_extractor:
```
araffin, line 670: that's something I forgot but need to be mentioned in the doc: if the feature extractor is not shared, then you cannot share layers later on.

stable_baselines3/common/policies.py
```diff
@@ -639,17 +687,21 @@ def get_distribution(self, obs: th.Tensor) -> Distribution:
         :return: the action distribution.
         """
         features = self.extract_features(obs)
+        if not self.share_features_extractor:
```
araffin, line 690: I think it would be cleaner to add a `extract_actor_features()` and `extract_critic_features()` methods

araffin:
Thanks for the PR =)

Looking at the code and thinking for the future, it would probably make sense to always separate the networks after the feature extractor (so deprecate and remove the shared network in the mlp).
That would allow to have a cleaner distinction between actor and critic and have consistent code with the off-policy algorithms.

cc @ernestum @Miffyli @qgallouedec

stable_baselines3/common/policies.py
AlexPasqua: Alright! I had put it in the `features_extractor_kwargs` in order not to introduce too many modifications (and also because it's an argument related to the features extractor), but I'll put it in the class as you said 👌

stable_baselines3/common/policies.py
AlexPasqua: Ok, I can add that to the doc!

Also, I could put a check after that line (670), something like "if shared_net not empty, raise error", but if, as you say in [this comment](https://github.com/DLR-RM/stable-baselines3/pull/1148#issuecomment-1299747968), you plan to remove the `shared_net` entirely, this might be useless.
Should I put the check in the meanwhile, and then remove it when the code gets refactored to remove the `shared_net`, in case you decide to do so?

stable_baselines3/common/policies.py
AlexPasqua: I agree, I'll do it!

stable_baselines3/common/policies.py
AlexPasqua: Ok I put it in the doc and I also put a check in the constructor of `ActorCriticPolicy`

AlexPasqua:
I've also modified the **doc**, but in doing this I removed the link to [issue #1066](https://github.com/DLR-RM/stable-baselines3/issues/1066#issuecomment-1246866844), because that guide would not be necessary anymore if we add this flag for shared/non-shared features extractor.
I don't know if you prefer to keep that link somewhere in the doc, in case I'll put it back.

Also, do you think the **tests** need to be updated for this features?

stable_baselines3/common/policies.py
```diff
@@ -547,11 +559,15 @@ def _build(self, lr_schedule: Schedule) -> None:
             # features_extractor/mlp values are
             # originally from openai/baselines (default gains/init_scales).
             module_gains = {
-                self.features_extractor: np.sqrt(2),
                 self.mlp_extractor: np.sqrt(2),
                 self.action_net: 0.01,
                 self.value_net: 1,
             }
+            if not self.share_features_extractor:
```
ernestum, line 566: Removing the negation and switching the blocks might make this easier to read. Same in line 587 below.

ernestum:
@araffin so you mean removing the ability to have any shared layers between value and policy network? That was always intended for reducing dimensionality or runtime but for me both never really improved anything. I don't know how it is for other applications though. If you are sufficiently confident that is never used anyways I am happy to remove the weight sharing feature to reduce the complexity of SB3.

araffin:
> o you mean removing the ability to have any shared layers between value and policy network? 

yes, I opened an issue here describing a tit bit more my motivation: https://github.com/DLR-RM/stable-baselines3/issues/1161

>  If you are sufficiently confident that is never used anyways I am happy to remove the weight sharing feature to reduce the complexity of SB3.

I would first deprecate and then remove.

stable_baselines3/common/policies.py
AlexPasqua: Did it

araffin (commented):
I simplified the implementation and added a test, it's almost ready to be merged but the results are changed for `share_features_extractor=True` (I ran `python -m rl_zoo3.train --algo a2c --env PongNoFrameskip-v4 -n 10000 --seed 1619064180 -P` and check the values at 8000 steps).
I need to see if it's expected, for instance, if we changed the pseudo random generator state or if a bug was introduced.

EDIT: it's fine, it just comes from the order in `module_gains`

stable_baselines3/common/policies.py
```diff
@@ -559,15 +563,18 @@ def _build(self, lr_schedule: Schedule) -> None:
             # features_extractor/mlp values are
             # originally from openai/baselines (default gains/init_scales).
             module_gains = {
+                self.features_extractor: np.sqrt(2),
```
AlexPasqua, line 566: The attributes `self.pi_features_extractor` and `self.vf_features_extractor` are created anyway in the `ActorCriticPolicy`'s constructor, and, in case the features extractor is shared, they are simply aliases for `self.features_extractor`.
So, at this point, wouldn't it be possible to do something like what follows?

```suggestion
                self.pi_features_extractor: np.sqrt(2),
```

araffin: I removed `del self.features_extractor`, should be the cleaner solution

araffin:
@ernestum @qgallouedec could you have a look at it in the coming week?
the changes are relatively small but it's the start of a more major change, so I want to be sure we make it right ;)

stable_baselines3/common/policies.py
```diff
@@ -114,16 +114,17 @@ def make_features_extractor(self) -> BaseFeaturesExtractor:
         """Helper method to create a features extractor."""
         return self.features_extractor_class(self.observation_space, **self.features_extractor_kwargs)
 
-    def extract_features(self, obs: th.Tensor) -> th.Tensor:
+    def extract_features(self, obs: th.Tensor, features_extractor: Optional[BaseFeaturesExtractor] = None) -> th.Tensor:
         """
         Preprocess the observation if needed and extract features.
 
         :param obs:
         :return:
```
qgallouedec, line 122:
```suggestion
        Preprocess the observation if needed and extract features.

         :param obs: The observation
         :param features_extractor: The features extractor to use. If it is set to None, the features extractor of the policy is used.
         :return: The features

```

qgallouedec:
> That would allow to have a cleaner distinction between actor and critic and have consistent code with the off-policy algorithms.

I'm currently reviewing this PR and I agree. I think it would makes sense to simplify. I think it should be left for the future and that it is out of the scope of this PR, but I am willing to work on it.

stable_baselines3/common/policies.py
```diff
@@ -114,16 +114,18 @@ def make_features_extractor(self) -> BaseFeaturesExtractor:
...
+         :param obs: The observation
+         :param features_extractor: The features extractor to use. If it is set to None, the features extractor of the policy is used.
+         :return: The features
         """
-        assert self.features_extractor is not None, "No features extractor was set"
+        features_extractor = features_extractor or self.features_extractor
+        assert features_extractor is not None, "No features extractor was set"
         preprocessed_obs = preprocess_obs(obs, self.observation_space, normalize_images=self.normalize_images)
-        return self.features_extractor(preprocessed_obs)
+        return features_extractor(preprocessed_obs)
```
qgallouedec, line 128: I'm not very satisfied with this method, when you see it it's not easy to understand why in some cases you need to use the feature extractor of the instance, and in others not. I think that in the future it should be simplified.

qgallouedec (approved):

stable_baselines3/common/policies.py
araffin: > I think that in the future it should be simplified.

how would you simplify it? 
always call it with a `features_extractor` parameter?

AlexPasqua: > > I think that in the future it should be simplified.
> 
> how would you simplify it? always call it with a `features_extractor` parameter?

Yeah I personally think it could be a bit clearer to always pass the features extractor to that method, both in case it corresponds to `self.features_extractor` and in case it's a `pi` or `vf` features extractor.

I could do it, in case you opt for that

AlexPasqua:
> > That would allow to have a cleaner distinction between actor and critic and have consistent code with the off-policy algorithms.
> 
> I'm currently reviewing this PR and I agree. I think it would makes sense to simplify. I think it should be left for the future and that it is out of the scope of this PR, but I am willing to work on it.

I'd like to contribute too, if it's ok for you

qgallouedec:
> I'd like to contribute to this too, if it's ok for you

Of course it's ok! Thanks for your contributions @AlexPasqua!

stable_baselines3/common/policies.py
araffin: @qgallouedec I added deprecation warning so we can make it the default in v1.8.0.

tests/test_custom_policy.py
```diff
@@ -28,17 +28,23 @@ def test_custom_offpolicy(model_class, net_arch):
...
 
-@pytest.mark.parametrize("model_class", [A2C, PPO, SAC, TD3])
+@pytest.mark.parametrize("model_class", [A2C, DQN, PPO, SAC, TD3])
 @pytest.mark.parametrize("optimizer_kwargs", [None, dict(weight_decay=0.0)])
 def test_custom_optimizer(model_class, optimizer_kwargs):
+    # Use different environment for DQN
+    if model_class is DQN:
+        env_id = "CartPole-v1"
+    else:
+        env_id = "Pendulum-v1"
```
AlexPasqua, line 38: I'd make this one line for a slightly better reading flow
```suggestion
    env_id = "CartPole-v1" if model_class is DQN else "Pendulum-v1"
```

araffin:
@AlexPasqua could you make a similar PR to SB3 contrib?

> Looking at the code and thinking for the future, it would probably make sense to always separate the networks after the feature extractor (so deprecate and remove the shared network in the mlp).
That would allow to have a cleaner distinction between actor and critic and have consistent code with the off-policy algorithms.

and for that, I would first do a PR that deprecates, then remove in a future major version

AlexPasqua:
> @AlexPasqua could you make a similar PR to SB3 contrib?

Sure! 👍 

> > Looking at the code and thinking for the future, it would probably make sense to always separate the networks after the feature extractor (so deprecate and remove the shared network in the mlp).
> > That would allow to have a cleaner distinction between actor and critic and have consistent code with the off-policy algorithms.
> 
> and for that, I would first do a PR that deprecates, then remove in a future major version

Alright! 👍

araffin:
@AlexPasqua unless you are already working on it (and do the PR today), I will have to do the SB3 contrib PR myself (we need a hotfix for ARS on SB3 contrib master)

AlexPasqua:
> @AlexPasqua unless you are already working on it (and do the PR today), I will have to do the SB3 contrib PR myself (we need a hotfix for ARS on SB3 contrib master)

Yes I'll do the PR today

araffin:
> Yes I'll do the PR today

thanks =)
in case you don't have time, tell me and I'll do it ;)

AlexPasqua:
Opened PR Stable-Baselines-Team
/
stable-baselines3-contrib#130

AlexPasqua:
> > Looking at the code and thinking for the future, it would probably make sense to always separate the networks after the feature extractor (so deprecate and remove the shared network in the mlp).
> > That would allow to have a cleaner distinction between actor and critic and have consistent code with the off-policy algorithms.
> 
> and for that, I would first do a PR that deprecates, then remove in a future major version

@araffin 
About the removal of the shared layers, should I start working on it already or should I wait that version 1.8 appears as WIP?

araffin:
> About the removal of the shared layers, should I start working on it already or should I wait that version 1.8 appears as WIP?

You could start working but I want to release v1.8.0 first (very soon, I will probably do it right atfer https://github.com/DLR-RM/stable-baselines3/pull/1260).
I also have some things specific in mind (I will probably help you for that):
- patch the loading (and move current patches to `save_util`)
- change the behavior of `net_arch=[64, 64]`
- try to provide meaningful error when someone try to load a shared net with SB3 v1.8.0

One thing that I'm not sure yet: shall we remove `MlpExtractor` completely? (and replace it with Actor/Critic class, the way it is done for off-policy algorithms)
I would say no for now.

AlexPasqua:
> * change the behavior of `net_arch=[64, 64]`

You mean making `net_arch=[64, 64]` equivalent to `net_arch=dict(pi=[64, 64], vf=[64, 64])`?

> One thing that I'm not sure yet: shall we remove `MlpExtractor` completely? (and replace it with Actor/Critic class, the way it is done for off-policy algorithms) I would say no for now.

I'd say no as well for now

araffin:
> You mean making net_arch=[64, 64] equivalent to net_arch=dict(pi=[64, 64], vf=[64, 64])?

yes! (to be consistent with SAC/TD3)

PR #1154, opened 2022-11-03
Raise error when same env object instance is passed in vectorized environment

### LINKED ISSUE #1151: [Feature Request] Raise error when same object in memory passed to vectorized environment
### 🚀 Feature

TLDR: add a check to the result of env construction in `make_vec_env` that raises an error if all envs have the same memory ID. Happy to submit a PR for this.

### Motivation

Currently when one creates a vectorized environment using `make_vec_env` and passes a function to create an environment multiple times, there are no checks making sure that the function indeed returns a fresh environment every time.

A priori it might seem that it's the user's responsibility to ensure this. However, if one is constructing an environment using hyperparameters and has a function that takes arguments to do this, e.g.
```python
def make_my_env(*args, **kwargs): ...
```

it is relatively easy to forget to wrap this as a closure. One should do:
```python
def make_env():
    return make_my_env(*args, **kwargs): ...

venv = make_vec_env(make_env, ...)
```
or
```python
venv = make_vec_env(lambda: make_my_env(*args, **kwargs), ...)
```
but if one instead does
```python
env = make_my_env(*args, **kwargs)
venv = make_vec_env(lambda: env, ...)
```
which is not something totally unreasonable to do at first sight, one will get totally undefined behavior where the same environment is being operated sequentially with actions that were calculated from multiple steps ago. 

Same happens if you do 
```python
venv = DummyVecEnv([lambda: env] * n_envs, ...)
```
since it's fairly common to write
```python
venv = DummyVecEnv([lambda: env], ...)
```
and to write
```python
venv = DummyVecEnv([make_env] * n_envs, ...)
```
so mixing both up is not so unexpected.
This has actually happened to me twice in the space of two months, and it took me several hours to realize of the errors. environments start behaving in totally bizarre ways, and it's hard to realize that it's not a policy training issue or a reward mis-specification issue unless you try to run with a single parallel environment and notice that it works fine.

### Pitch

Raising an error if the same env instance is returned every time the env constructor is called is fairly straightforward and would avoid this type of error. I bet that no one would ever actually do this intentionally and it is almost surely always an error.

Just do in the DummyVecEnv initializer:
```python
        if len(set(map(id, self.envs))) != len(self.envs):
            raise ValueError("The environments should be different instances")
```
or `set([id(env) for env in self.envs]))` using a list comprehension which is probably more pythonic.

### Alternatives

_No response_

### Additional context

_No response_

### Checklist

Rocamonde:
## Description
Raises an error when the same gym environment object instance is passed when creating a vectorized environment with more than one environment.

## Motivation and Context

Fixes #1151 to prevent hard-to-debug errors.

## Types of changes

## Checklist:


qgallouedec:
Does the same problem occur with `SubprocVecEnv`?

Rocamonde:
Probably but I am not confident with how process spawning / memory sharing works well enough to propose a solution by myself.

Rocamonde:
It's possible that if the object is copied to each process, they go out of sync, which is normally something to avoid but in this case it's exactly what we want. Not sure if the objects are shared or copied, there's some custom pickling going on and someone that understands this part of the codebase better should weigh in.

qgallouedec:
I checked, they are copied and not shared. However, since it is a fortunate behavior, I would still like to raise an error in this case.

araffin:
> I checked, they are copied and not shared.

yes as we default to safe option (`forkserver` instead of the default `fork`).

qgallouedec:
Looking closer, it looks much harder to raise an error in `SubprocVecEnv`.

qgallouedec:
So that's what I thought. It's not the size of the observation that changes, but the size of the input batch. 
It is 64 during training (`batch_size=64`), and 1 during the interaction phase.

qgallouedec:
I find the error message unclear about what I am supposed to do as a user to solve the problem.
Do you have an alternative proposal?
I would go to something like

```
The environment making functions all return the same instance. You probably use
something like ``env_fns=[lambda: env] * n_envs``. Instead, you should use
``env_fns=[make_env] * n_envs`` where ``gym.make`` is called inside ``make_env``.
```

araffin:
>  I would go to something like

I would also link the github issue.

I would make it even more detailed:
```
You tried to create multiple environments, but the function to create them returned the same instance
instead of creating different objects.
You are probably using `make_vec_env([lambda : env])` or `DummyVecEnv([lambda : env] * n_envs)`.
You should replace `lambda : env` by a `make_env` function that creates a new instance of the environment at every call
(using `gym.make()` for instance). You can take a look at the documentation for an example.

Please read https://github.com/DLR-RM/stable-baselines3/issues/1151 for more information.
```

tests/test_vec_envs.py
```diff
@@ -62,6 +62,14 @@ def custom_method(dim_0=1, dim_1=1):
         return np.ones((dim_0, dim_1))
 
 
+def test_vecenv_func_checker():
+    """The functions in ``env_fns'' must return distinct instances since we need distinct environments."""
+    env = CustomGymEnv(gym.spaces.Box(low=np.zeros(2), high=np.ones(2)))
+
+    with pytest.raises(ValueError):
```
araffin, line 69: although it should be the same, I would also add a test with `make_vec_env(lambda: env, n_envs=2)` as it is the other common mistake I think.

qgallouedec: You have a nose for it.

The next test does not pass.

```python
env = gym.make("CartPole-v1")

with pytest.raises(ValueError):
    make_vec_env(lambda: env, n_envs=2)

env.close()
```

 I am investigating.

araffin: i guess because of the monitor wrapper?
you have to use unwrapped to get the original env

araffin: i guess because of the monitor wrapper?
you have to use unwrapped to get the original env

qgallouedec: Ok... It's because it's wrapped, so the id is the different, even though the inner env is the same. @Rocamonde, can you fix this?

```python
import gym

env = gym.make("CartPole-v1")

print(id(env))  # 4551646560

venv = make_vec_env(lambda: env, n_envs=2)

print([id(env) for env in venv.envs])  # [5311397600, 5311397648]
print([id(env.env) for env in venv.envs])  # [4551646560, 4551646560]

env.close()
```

araffin: `env.unwrapped` is what you are looking for, but you need to check that this method exist first.

Rocamonde: Hmm. What do we actually want to do here, raise an error if any two levels of nesting match? I think we only care about the object at the top of the hierarchy. If the instances at the top are distinct, then that's a guarantee that all wrappers are distinct, since you can't a same wrapper instance wrapping two different environment instances.

In think we should write a helper that extracts the primitive environment for each environment, then compare those ids. Does that sound sensible?

araffin: >  I think we only care about the object at the top of the hierarchy.

we care about the original gym env, which is returned by `unwrapped`.

> If the instances at the top are distinct, then that's a guarantee that all wrappers are distinct, since you can't have a same wrapper instance wrapping two different environment instances.

not sure to understand.
But for short, if the gym instance is the same, even though the wrappers are different, there is an issue (see example in the tests).

> In think we should write a helper that extracts the primitive environment for each environment,

this is what you get with the `unwrapped` attribute ;)

Rocamonde: Wdym by "almost"? Why is it an attribute that sometimes does not exist? Surely you should have a function that maps `f(any_env: Union[gym.Env, vec_env.VecEnv]) -> unwrapped_env: gym.Env`

Rocamonde: Yes, agreed that this is an issue. I was trying to say that you need to unwrap the env all the way, since that's what contains the stateful attributes. And that two envs having different id's when fully unwrapped always implies that any wrappers have different id's, but not the other way around (since the user could still be using the same gym env but then e.g. creating different wrapper instances with different params).

araffin: > Why is it an attribute that sometimes does not exist? 

it always exists (it is in the base class for gym env and wrappers).

> EDIT: this helper almost exist actually

I meant we have helpers to unwrap `VecEnv` and `gym.Env`:  https://github.com/DLR-RM/stable-baselines3/blob/7e1db1aaaa0f486cceb79faf5a08a25c5fded670/stable_baselines3/common/env_util.py#L11

> still be using the same gym env but then e.g. creating different wrapper instances with different params).

How would that not create issues?
Because at the end, you are still updating the same env instance despite different wrappers.

araffin: > Great will use that in the PR!

there might be a misunderstanding, `id(env.unwrapped)` is what you should use, no need for helper

araffin:
@qgallouedec I managed to finally get some colors with mypy =) (https://github.com/DLR-RM/stable-baselines3/pull/1154/commits/3d0f49d7845f2639599c104ebb074e8cb875597b)

araffin (approved):
LGTM, thanks =)

PR #1167, opened 2022-11-10
Fix `Self` return type

qgallouedec:
## Description

Partially fixed by #1043

See the python doc: https://docs.python.org/3/library/typing.html#typing.Self

## Motivation and Context

- [x] I have raised an issue to propose this change ([required](https://github.com/DLR-RM/stable-baselines3/blob/master/CONTRIBUTING.md) for new features and bug fixes) #1040

## Types of changes

## Checklist:

### DIFF
--- a/stable_baselines3/common/distributions.py
+++ b/stable_baselines3/common/distributions.py
@@ -38,7 +38,7 @@ class Distribution(ABC):
         concrete classes."""
 
     @abstractmethod
-    def proba_distribution(self, *args, **kwargs) -> SelfDistribution:
+    def proba_distribution(self: SelfDistribution, *args, **kwargs) -> SelfDistribution:
         """Set parameters of the distribution.
 
         :return: self
@@ -151,7 +151,9 @@ class DiagGaussianDistribution(Distribution):
         log_std = nn.Parameter(th.ones(self.action_dim) * log_std_init, requires_grad=True)
         return mean_actions, log_std
 
-    def proba_distribution(self, mean_actions: th.Tensor, log_std: th.Tensor) -> SelfDiagGaussianDistribution:
+    def proba_distribution(
+        self: SelfDiagGaussianDistribution, mean_actions: th.Tensor, log_std: th.Tensor
+    ) -> SelfDiagGaussianDistribution:
         """
         Create the distribution given its parameters (mean, std)
 
@@ -217,7 +219,9 @@ class SquashedDiagGaussianDistribution(DiagGaussianDistribution):
         self.epsilon = epsilon
         self.gaussian_actions = None
 
-    def proba_distribution(self, mean_actions: th.Tensor, log_std: th.Tensor) -> SelfSquashedDiagGaussianDistribution:
+    def proba_distribution(
+        self: SelfSquashedDiagGaussianDistribution, mean_actions: th.Tensor, log_std: th.Tensor
+    ) -> SelfSquashedDiagGaussianDistribution:
         super().proba_distribution(mean_actions, log_std)
         return self
 
@@ -281,7 +285,7 @@ class CategoricalDistribution(Distribution):
         action_logits = nn.Linear(latent_dim, self.action_dim)
         return action_logits
 
-    def proba_distribution(self, action_logits: th.Tensor) -> SelfCategoricalDistribution:
+    def proba_distribution(self: SelfCategoricalDistribution, action_logits: th.Tensor) -> SelfCategoricalDistribution:
         self.distribution = Categorical(logits=action_logits)
         return self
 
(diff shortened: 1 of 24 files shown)


qgallouedec:
I'm not sure how to handle that one:

https://github.com/DLR-RM/stable-baselines3/blob/7e1db1aaaa0f486cceb79faf5a08a25c5fded670/stable_baselines3/common/vec_env/base_vec_env.py#L207

stable_baselines3/common/distributions.py
```diff
@@ -298,6 +311,9 @@ def log_prob_from_params(self, action_logits: th.Tensor) -> Tuple[th.Tensor, th.
         return actions, log_prob
 
 
+SelfMultiCategoricalDistribution = TypeVar("SelfMultiCategoricalDistribution", bound="MultiCategoricalDistribution")
```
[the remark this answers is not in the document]
qgallouedec: you are right, it seems to be the usual.

araffin:
> I'm not sure how to handle that one:
> 
> https://github.com/DLR-RM/stable-baselines3/blob/7e1db1aaaa0f486cceb79faf5a08a25c5fded670/stable_baselines3/common/vec_env/base_vec_env.py#L207

`SelfVecEnv` is not ok?

qgallouedec:
I'm not sure it is correct, because. Look at this example:

```python
import gym
from stable_baselines3.common.vec_env import DummyVecEnv, VecNormalize

venv = DummyVecEnv([lambda: gym.make("Pendulum-v1")])
venv = VecNormalize(venv)
print(venv.unwrapped.__class__.__name__)  # DummyVecEnv
```

If I annotate `unwrapped`, I would except `venv.unwrapped` type to be `VecNormalize`, and not `DummyVecEnv`

qgallouedec:
I think this difficulty comes from the fact that we refer to `VecEnvWrapper` in `VecEnv` when `VecEnvWrapper` is not defined yet. A more correct way could be to overwrite the property:

```python
class VecEnv:
    def unwrapped(self):
        if isinstance(self, VecEnvWrapper):
            return self.venv
        else:
            return self

class VecEnvWrapper(VecEnv):
    venv: VecEnv
```

to

```python
class VecEnv:
    @property
    def unwrapped(self):
        return self

class VecEnvWrapper(VecEnv):
    venv: VecEnv

    @property
    def unwrapped(self):
        return self.venv
```
It does not solve the current type hinting issue but it would make thing clearer.

qgallouedec:
At this point in my reflection, I think the correct way is the following. I would like confirmation.

```python
from abc import abstractmethod
from typing import TypeVar

SelfDummyVecEnv = TypeVar("SelfDummyVecEnv", bound="DummyVecEnv")

class VecEnv:
    @property
    @abstractmethod
    def unwrapped(self) -> "VecEnv":
        ...

class DummyVecEnv(VecEnv):
    @property
    def unwrapped(self: SelfDummyVecEnv) -> SelfDummyVecEnv:
        return self

class VecEnvWrapper(VecEnv):
    venv: VecEnv

    @property
    def unwrapped(self) -> VecEnv:
        return self.venv
```

stable_baselines3/common/distributions.py
```diff
@@ -28,7 +38,7 @@ def proba_distribution_net(self, *args, **kwargs) -> Union[nn.Module, Tuple[nn.M
         concrete classes."""
 
     @abstractmethod
-    def proba_distribution(self, *args, **kwargs) -> "Distribution":
+    def proba_distribution(self: SelfDistribution, *args, **kwargs) -> SelfDistribution:
```
araffin, line 41: is the `self: SelfDistribution` needed?

looks like duplicate info.

araffin:
> At this point in my reflection, I think the correct way is the following. I would like confirmation.

I would say I'm half happy about it.
Before you needed to look at one method to understand how it works (even though `VecEnvWrapper` was defined only afterward), now you need to look at several methods depending on what you are using, and the case where a user defines its own `VecEnv` (as I do in SB3 contrib for instance) is not covered (so it's a breaking change).

stable_baselines3/common/distributions.py
qgallouedec: It is. And yes I seems duplicate, that's the [motivation behind the new Self annotation](https://peps.python.org/pep-0673/) that we do not use for the moment (see https://github.com/DLR-RM/stable-baselines3/pull/1043#issuecomment-1240437433):

> The present workaround for such instances is to define a `TypeVar` with the base class as the bound and use it as the annotation for the `self` parameter and the return type:
> 
> ```python
> from typing import TypeVar
> 
> TShape = TypeVar("TShape", bound="Shape")
> 
> class Shape:
>     def set_scale(self: TShape, scale: float) -> TShape:
>         self.scale = scale
>         return self
> 
> 
> class Circle(Shape):
>     def set_radius(self, radius: float) -> Circle:
>         self.radius = radius
>         return self
> 
> Circle().set_scale(0.5).set_radius(2.7)  # => Circle
> ```
> 
> Unfortunately, this is verbose and unintuitive. Because `self` is usually not explicitly annotated, the above solution doesn’t immediately come to mind, and even if it does, it is very easy to go wrong by forgetting either the bound on the `TypeVar(bound="Shape")` or the annotation for self.

qgallouedec:
> Before you needed to look at one method to understand how it works (even though VecEnvWrapper was defined only afterward), now you need to look at several methods depending on what you are using

I don't think it's a problem.

> EDIT: I would therefore keep it as-is for now.

Ok for me.

araffin (approved):
LGTM, thanks =)

PR #1193, opened 2022-11-29
Fix `stable_baselines3/common/vec_env/__init__.py` type hint 

qgallouedec:
## Motivation and Context

## Types of changes

## Checklist:


stable_baselines3/common/vec_env/__init__.py
```diff
@@ -19,8 +19,10 @@
 if typing.TYPE_CHECKING:
     from stable_baselines3.common.type_aliases import GymEnv
 
+VecEnvWrapperT = TypeVar("VecEnvWrapperT", bound=VecEnvWrapper)
```
araffin, line 22: not sure to understand the difference with before, but if it works...

qgallouedec:
```python
T = TypeVar("T")

def func(arg: T) -> T: ...
```

ensures that the return type of func is the same as the arg type.

Before, `unwrap_vec_wrapper` was allowed to return any subtype of `VecEnvWrapper`, regardless of the type of `vec_wrapper_class`.

stable_baselines3/common/vec_env/__init__.py
```diff
@@ -65,7 +67,9 @@ def sync_envs_normalization(env: "GymEnv", eval_env: "GymEnv") -> None:
     """
     env_tmp, eval_env_tmp = env, eval_env
     while isinstance(env_tmp, VecEnvWrapper):
+        assert isinstance(eval_env_tmp, VecEnvWrapper)  # Ensure that env and eval_env are similarly wrapped.
```
araffin, line 70: I think this is a breaking change (see how it is handled in `EvalCallback`).
There are also several issues related to that.
It might be ok to add those assert but we must improve the error message then.

PR #1194, opened 2022-11-29
Fix `test_vec_normalize.py`, `test_tensorboard.py` and `common/monitor.py` type hint

qgallouedec:
## Description

In this PR, I also replaced `Union[int, float]` by just `float`, as they are equivalent for type hint: 
 In https://peps.python.org/pep-0484/#the-numeric-tower
> Rather than requiring that users write import numbers and then use numbers.Float etc., this PEP proposes a straightforward shortcut that is almost as effective: when an argument is annotated as having type float, an argument of type int is acceptable

> **Warning**
> This is only true for type hint, and you can replace `isinstance(x, (int, float))` by `isinstance(x, float)`

## Motivation and Context

## Types of changes

## Checklist:


tests/test_tensorboard.py
```diff
@@ -18,21 +21,19 @@
...
-        hparam_dict = {
-            "algorithm": self.model.__class__.__name__,
-            "learning rate": self.model.learning_rate,
-            "gamma": self.model.gamma,
-        }
+        hparam_dict: Dict[str, Union[str, float]] = {"algorithm": self.model.__class__.__name__}
+        if isinstance(self.model.learning_rate, float):  # Can also be Schedule, in that case, we don't report
+            hparam_dict["learning rate"] = self.model.learning_rate
+        if isinstance(self.model, (OnPolicyAlgorithm, OffPolicyAlgorithm)):
+            hparam_dict["gamma"] = self.model.gamma
```
qgallouedec, line 33: Currently, gamma is defined in `OnPolicyAlgorithm` and `OffPolicyAlgorithm`, but not in `BaseAlgorithm`. I am not very satisfied with this solution, but it works.

araffin: gamma makes sense for most algorithms but not ES ones like ARS/CEM (actually we could set gamma=1 for them for consistency even though it is not used)

araffin: > I am not very satisfied with this solution, but it works.

yes, not pretty but:
1. it is for a test
2. you could probably ignore mypy here

araffin (commented):
What was missing for that one?

qgallouedec:
The CI didn't pass, so I put it off. But if the CI passes now, I think it 's ready to be merged.

araffin (approved):

Issue #1201, opened 2022-12-06
Unable to replicate SAC SB results in SB3


vdesai2014:
### 🐛 Bug

I am trying to use Stable Baselines to learn a grasping policy for a robot and transfer it to a real robot. I found this project by Baris Yazici (https://github.com/BarisYazici/deep-rl-grasping) that was originally done in the older version of Stable Baselines, the agent is able to learn a useful policy starting at around 50,000 timesteps and pretty quickly converges to a high success rate. The original SB observation was passed in as a float from 0 to 1, and SB3 only accepts input of integer type from 0 to 255 for Cnn Policies so I modified the observation in order to match this requirement. The very minor code changes I made are listed in the "Code Example" below, along with the original SB implementation . When using SB3, however, the agent does not learn even after 500,000 timesteps. 

Are folks familiar with differences in implementation of Cnn Policies in SB vs. SB3 able to help with explaining what may be going on here? For whatever reason, training using SB becomes very slow after 20,000 timesteps (1 frame per second), otherwise I would just stick with SB. SB3 runs much faster, but fails to converge. To be honest I'm pretty surprised that the original SB implementation works, as the observation is passed in as a float from 0 to 1, and gets normalized (/= 255) by the VecNormalize function. It seems my issue has to do with the way the observation is changed to be compatible with SB3 CnnPolicy requirements, but I don't see why the minor changes I am making break the learning process. I should note the environment itself is identical in all my trials & hyper parameters are kept constant. 

### Code example

```python
#Original SB Implementation for observation, returns an array float's with shape (64, 64, 2)

 def _observe(self):
         rgb, depth, _ = self._camera.get_state()
         sensor_pad = np.zeros(self._camera.state_space.shape[:2])
         sensor_pad[0][0] = self._actuator.get_state() #gripper_width 
         obs_stacked = np.dstack((depth, sensor_pad))
         return obs_stacked 

 #Original SB Implementation for obs space

 self.observation_space = gym.spaces.Box(low=0, high=255, shape=(shape[0], shape[1], 2))

 #Original SB code that defines policy & initializes training 

 import stable_baselines as sb
 from stable_baselines.sac.policies import CnnPolicy as sacCnn
 from stable_baselines.common.vec_env import VecNormalize

 model = sb.SAC('CnnPolicy',
 self.env,
 verbose=2,
 gamma=0.99,
 buffer_size=1000000,
 batch_size=64,
 learning_rate=0.0003)

[...]

PR #1214, opened 2022-12-18
Fix support of image like normalized inputs

araffin:
## Description

Many things were done wrong when using normalized 3D input, see https://github.com/DLR-RM/stable-baselines3/issues/1201

This PR fixes the default behavior and improve error messages.

## Motivation and Context

Fixes https://github.com/DLR-RM/stable-baselines3/issues/1201

## Types of changes

## Checklist:


stable_baselines3/common/policies.py
```diff
@@ -87,6 +87,9 @@ def __init__(
 
         self.features_extractor_class = features_extractor_class
         self.features_extractor_kwargs = features_extractor_kwargs
+        # Automatically deactivate dtype and bounds checks
+        if normalize_images is False and features_extractor_class in {NatureCNN, CombinedExtractor}:
+            self.features_extractor_kwargs.update(dict(normalized_image=True))
```
qgallouedec, line 92: I'm not sure to understand this.
Is that to force normalization in `NatureCNN` and `CombinedExtractor`, even though it is disabled by the user?

stable_baselines3/common/policies.py
```diff
@@ -87,6 +87,9 @@ def __init__(
 
         self.features_extractor_class = features_extractor_class
         self.features_extractor_kwargs = features_extractor_kwargs
+        # Automatically deactivate dtype and bounds checks
+        if normalize_images is False and features_extractor_class in {NatureCNN, CombinedExtractor}:
```
qgallouedec, line 91: I'd also include any subclass of those two features extractors.

stable_baselines3/common/policies.py
qgallouedec: Oh, I got it. I read "normalize_images" instead of "normalized_image".

stable_baselines3/common/torch_layers.py
```diff
@@ -55,19 +55,31 @@ class NatureCNN(BaseFeaturesExtractor):
     :param observation_space:
     :param features_dim: Number of features extracted.
         This corresponds to the number of unit for the last layer.
+    :param normalized_image: Whether to assume that the image is already normalized
+        or not (this disables dtype and bounds checks): when True, it only checks that
+        the space is a Box and has 3 dimensions.
+        Otherwise, it checks that it has expected dtype (uint8) and bounds (values in [0, 255]).
```
qgallouedec, line 61: Do you think that we have to check that dtype is a Box when `normalized_image` is True?

stable_baselines3/common/policies.py
araffin: yes... the idea is to have good defaults for the user and throw informative error when some explicit parameters are missing

stable_baselines3/common/torch_layers.py
araffin: dtype is either `float32` os `uint8`, we check that the obs space is a box in all cases.

qgallouedec: typo:
Do you think that we have to check that dtype is a float32 when normalized_image is True?

araffin: not sure, I think it doesn't matter as `obs.float()` is called in pre-processing afterward, and at that point the user should better know what he/she is doing (passing `normalize_images=False` is an active choice).

stable_baselines3/common/vec_env/vec_normalize.py
```diff
@@ -50,9 +51,27 @@ def __init__(
...
                 self.obs_spaces = None
                 self.obs_rms = RunningMeanStd(shape=self.observation_space.shape)
+                # Update observation space when using image
+                if is_image_space(self.observation_space):
+                    self.observation_space = gym.spaces.Box(
+                        low=-clip_obs,
+                        high=clip_obs,
+                        shape=self.observation_space.shape,
+                        dtype=np.float32,
+                    )
```
qgallouedec, line 74: Why that applies only for images?

araffin: we could apply it for others, but this would create issues with already saved models (we check the obs space at load time).
In the case here, it's to raise proper error when using image-like input with vecnormalize.
For other cases, Vecnormalize is meant to be used with Box space only, so we would only change the bounds (i agree it would be more correct but I'm afraid it would create more issues).
I'm happy to change my mind though ;)

qgallouedec: I understand. I'm in two minds about it. Maybe the best thing for now is to stay like this and add a comment stating what you just explained to me, to potentially help a later improvement.

qgallouedec (approved):

Issue #1124, opened 2022-10-16
SB2 vs SB3 - Performance difference


MatPoliquin:
### ❓ Question

EDIT: After doing some more digging I updated the post title and added more details with a newer version of SB3 (1.6.2)

I am using OpenAI gym-retro env to train on games and migrated from SB2 to SB3 1.6.2. I noticed the training FPS reduced by a lot from  1300fps to 900fps.

- gym-retro env: Pong-Atari2600
- num_env==24
- PPO
- CnnPolicy

Using Nvidia Nsight I profiled both versions (you can find the reports in the link to google drive below, you need Nsight to view it):
[https://drive.google.com/drive/folders/1Lqxf-qKXTj__Hp8WUXgNHejZaJGy8oct?usp=sharing](https://drive.google.com/drive/folders/1Lqxf-qKXTj__Hp8WUXgNHejZaJGy8oct?usp=sharing)

Here are the parameters I use for PPO with SB3 (with SB1 I just use the default parameters provided by SB):
`PPO(policy=args.nn, env=env, verbose=1, n_steps = 128, n_epochs = 4, batch_size = 256, learning_rate = 2.5e-4, clip_range = 0.2, vf_coef = 0.5, ent_coef = 0.01, max_grad_norm=0.5, clip_range_vf=None)`

My specs:

- Dual Xeon 2666v3
- RTX 2060 Super 8g
- Ubuntu 20.04
- stable-baselines3 1.6.2
- gym 0.26.2

Code I use to wrap the retro env (same for both SB2 and SB3 cases):

```
 def make_retro(*, game, state=None, num_players, max_episode_steps=4500, **kwargs):
      import retro
      if state is None:
         state = retro.State.DEFAULT
     env = retro.make(game, state, **kwargs, players=num_players)
     return env
```
 
 ```
def init_env(output_path, num_env, state, num_players, args, use_frameskip=True, use_display=False):
     seed = 0
     start_index = 0
     start_method=None
     allow_early_resets=True
    
     def make_env(rank):
        def _thunk():
             env = make_retro(game=args.env, use_restricted_actions=retro.Actions.FILTERED, state=state, num_players=num_players)
 
             env.seed(seed + rank)
             env = Monitor(env, output_path and os.path.join(output_path, str(rank)), allow_early_resets=allow_early_resets)
             if use_frameskip:
                 env = StochasticFrameSkip(env, n=4, stickprob=0.25)
 
             env = WarpFrame(env)
[...]

PR #1218, opened 2022-12-18
Construct tensors directly on GPUs

qgallouedec:
## Description

7.5% speedup on this example.

```python
from stable_baselines3 import DQN

DQN("CnnPolicy", "Pong-v4", learning_starts=10_000, buffer_size=30_000).learn(100_000)
```

## Motivation and Context

## Types of changes

## Checklist:

### DIFF
--- a/stable_baselines3/common/buffers.py
+++ b/stable_baselines3/common/buffers.py
@@ -132,8 +132,8 @@ class BaseBuffer(ABC):
         :return:
         """
         if copy:
-            return th.tensor(array).to(self.device)
-        return th.as_tensor(array).to(self.device)
+            return th.tensor(array, device=self.device)
+        return th.as_tensor(array, device=self.device)
 
     @staticmethod
     def _normalize_obs(
--- a/stable_baselines3/common/buffers.py
+++ b/stable_baselines3/common/buffers.py
@@ -127,8 +127,8 @@ class BaseBuffer(ABC):
         Note: it copies the data by default
 
         :param array:
-        :param copy: Whether to copy or not the data
-            (may be useful to avoid changing things be reference)
+        :param copy: Whether to copy or not the data (may be useful to avoid changing things
+            by reference). This argument is inoperative if the device is not the CPU.
         :return:
         """
         if copy:
--- a/stable_baselines3/common/buffers.py
+++ b/stable_baselines3/common/buffers.py
@@ -133,6 +133,8 @@ class BaseBuffer(ABC):
         """
         if copy:
             return th.tensor(array, device=self.device)
+        if self.device != th.device("cpu"):
+            warnings.warn("``copy`` is set to False but the device is not cpu. The data is therefore copied anyway.")
         return th.as_tensor(array, device=self.device)
 
     @staticmethod
--- a/stable_baselines3/common/buffers.py
+++ b/stable_baselines3/common/buffers.py
@@ -133,8 +133,6 @@ class BaseBuffer(ABC):
         """
         if copy:
             return th.tensor(array, device=self.device)
-        if self.device != th.device("cpu"):
-            warnings.warn("``copy`` is set to False but the device is not cpu. The data is therefore copied anyway.")
         return th.as_tensor(array, device=self.device)
 
     @staticmethod
--- a/docs/misc/changelog.rst
+++ b/docs/misc/changelog.rst
@@ -49,6 +49,7 @@ Others:
 - Fixed ``stable_baselines3/common/atari_wrappers.py`` type hints
 - Exposed modules in ``__init__.py`` with the ``__all__`` attribute (@ZikangXiong)
 - Upgraded GitHub CI/setup-python to v4 and checkout to v3
+- Set tensors construction directly on the device
 
 Documentation:
 ^^^^^^^^^^^^^^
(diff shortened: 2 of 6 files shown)


araffin:
i guess that's the same as https://github.com/DLR-RM/stable-baselines3/issues/1124#issuecomment-1356235336 ? (similar speedup)
(in the sense that if you set copy to False, there won't be anymore speedup)

qgallouedec:
Note that a large amount of the time is dedicated to GPU-CPU data transfer (more time than forward and backward pass)

Screenshot 2022-12-18 at 22 56 15

Generated with (snakeviz)

qgallouedec:
> i guess that's the same as https://github.com/DLR-RM/stable-baselines3/issues/1124#issuecomment-1356235336 ? (similar speedup)

I think it's not. The main difference between `torch.tensor([]).to(device)` and `torch.tensor([], device)` is that the first creates a tensor on the CPU and then moves it to the GPU, while the second creates the tensor directly on the GPU. The latter avoids the overhead of transferring the data between devices. This difference is independent of whether the tensor shares memory with the input data or not.

qgallouedec:
Some interesting behavior though:

```python
>>> import torch
>>> import numpy as np
>>> device = torch.device("cuda")
>>> a = np.ones(3, dtype=np.float32)
>>> b = torch.as_tensor(a)
>>> c = torch.as_tensor(a, device=device)
>>> a[0] = 0
>>> b
tensor([0., 1., 1.])
>>> c
tensor([1., 1., 1.], device='cuda:0')
```

qgallouedec:
[a5d20d2](https://github.com/DLR-RM/stable-baselines3/pull/1218/commits/a5d20d2c67e2b5097a80484d082f0c1134b20fec): The latest version of numpy (1.24) was released and removed `np.bool`(deprecated since a while). While gym==0.21 uses it. We need to set numpy<1.24 until sb3 is upgraded to be compatible with gym>0.21.

stable_baselines3/common/buffers.py
```diff
@@ -132,8 +132,10 @@ def to_torch(self, array: np.ndarray, copy: bool = True) -> th.Tensor:
         :return:
         """
         if copy:
-            return th.tensor(array).to(self.device)
-        return th.as_tensor(array).to(self.device)
+            return th.tensor(array, device=self.device)
+        if self.device != th.device("cpu"):
```
araffin, line 136: I don't think we need the warning there, I would put it in the doc as it is for `as_tensor`: https://pytorch.org/docs/stable/generated/torch.as_tensor.html

"If data is already a tensor with the requested dtype and device then data itself is returned, but if data is a tensor with a different dtype or device then it’s copied as if using data.to(dtype=dtype, device=device)."

stable_baselines3/common/buffers.py
```diff
@@ -132,8 +132,10 @@ def to_torch(self, array: np.ndarray, copy: bool = True) -> th.Tensor:
         :return:
         """
         if copy:
-            return th.tensor(array).to(self.device)
```
araffin, line 135: as this PR is about optimization, could you give `copy=False` as default a try?
(both CPU/GPU and check quickly that it doesn't change the results)

stable_baselines3/common/buffers.py
qgallouedec: ok for me

araffin:
> Note that a large amount of the time is dedicated to GPU-CPU data transfer (more time than forward and backward pass)

what color is the GPU-CPU data transfer in your graph?

qgallouedec:
> what color is the GPU-CPU data transfer in your graph?

Approx 27% runtime.

Screenshot 2022-12-19 at 10 43 44

stable_baselines3/common/buffers.py
araffin: I also tried `non_blocking=True` but it didn't work, probably because I was using `to()`?

https://github.com/DLR-RM/stable-baselines3/commit/6b147d8cca54e525d960d6d076bc7bc30cb1c73f

qgallouedec:
Here are the results (in seconds)

```python
from stable_baselines3 import DQN

DQN("CnnPolicy", "Pong-v4", learning_starts=1_000, buffer_size=3_000, verbose=1).learn(10_000)
```

### CPU

| | old (`to()`) | new (`device=device`) |
|-|-|-|
| copy | 533 | 539 |
| no copy | 540 | 563 |

### GPU

| | old (`to()`) | new (`device=device`) |
|-|-|-|
| copy | 50 | 45 |
| no copy | 48 | 48 |

qgallouedec:
With longer run on GPU:

```python
from stable_baselines3 import DQN

DQN("CnnPolicy", "Pong-v4", learning_starts=1_000, buffer_size=3_000, verbose=1).learn(50_000)
```

| | old (`to()`) | new (`device=device`) |
|-|-|-|
| copy | 230 | 212 |
| no copy | | |

araffin (approved):
LGTM, thanks =)

araffin:
EDIT: we need to update the version

qgallouedec:
1.7.0a6?

araffin:
> 1.7.0a6?

yep and maybe add a comment about speed boost

qgallouedec:
See #1220

PR #1238, opened 2022-12-25
Refactor observation stacking

qgallouedec:
### Summary of changes:

- `StackedObservations` now supports Dict as an observation space. It handles this by nesting instances of StackedObservation.
- `DictStackedObservation` has therefore been deprecated.
- Add tests for StackedObservations
- Fix type hints.

#### `stack_observation_space` deprecation

I deprecated the `stack_observation_space` method for the following reasons:
- it is no longer used in sb3
- it is probably not used outside the internal workings of sb3.
- it has an unsatisfactory way of working between method and static method (uses `self.repeat_axis`, but takes `observation_space` as parameter).

#### `compute_stacking` new params and returns

static method `compute_stacking` no longer takes `num_envs` and return the `stacked_shape` instead of a zero-filled stacked obs of shape `(num_env, *stacked_shape)`

## Motivation and Context

## Types of changes

## Checklist


araffin:
will try to review this one this week ;)

araffin:
Funnily, this is the 2nd time this piece of code is refactored ^^ first time was https://github.com/DLR-RM/stable-baselines3/pull/243 (currently doing the review ;))

stable_baselines3/common/vec_env/stacked_observations.py
```diff
@@ -75,192 +92,132 @@ def compute_stacking(
...
         repeat_axis = 0 if channels_first else -1
-        low = np.repeat(observation_space.low, n_stack, axis=repeat_axis)
-        stackedobs = np.zeros((num_envs,) + low.shape, low.dtype)
-        return channels_first, stack_dimension, stackedobs, repeat_axis
+        stacked_shape = list(observation_space.shape)
+        stacked_shape[repeat_axis] *= n_stack
+        return channels_first, stack_dimension, tuple(stacked_shape), repeat_axis
 
-    def stack_observation_space(self, observation_space: spaces.Box) -> spaces.Box:
+    def stack_observation_space(self, observation_space: Union[spaces.Box, spaces.Dict]) -> Union[spaces.Box, spaces.Dict]:
```
araffin, line 99: I thought about removing it but after a quick check, it is a bit used: https://github.com/search?o=desc&q=stable_baselines3+stack_observation_space&s=indexed&type=Code

so, let's remove it only in 2.x

stable_baselines3/common/vec_env/stacked_observations.py
```diff
@@ -75,192 +92,132 @@ def compute_stacking(
...
         """
-        Given an observation space, returns a new observation space with stacked observations
+        This function is deprecated.
+
+        As an alternative, use
```
araffin, line 103: `.stacked_observation_space` is not a better alternative?

stable_baselines3/common/vec_env/vec_frame_stack.py
```diff
@@ -1,64 +1,49 @@
...
+        observation_space = self.stacked_obs.stacked_observation_space
+        super().__init__(venv, observation_space=observation_space)
 
-        else:
-            raise Exception("VecFrameStack only works with gym.spaces.Box and gym.spaces.Dict observation spaces")
-
-        observation_space = self.stackedobs.stack_observation_space(wrapped_obs_space)
-        VecEnvWrapper.__init__(self, venv, observation_space=observation_space)
+    @property
+    def n_stack(self):
```
araffin, line 32: was it used somewhere?

stable_baselines3/common/vec_env/stacked_observations.py
qgallouedec: On the contrary, it seems that most of the matches on this search match `def stack_observation_space` that come from the copies of sb3. I can't find any code where this function is actually used.

stable_baselines3/common/vec_env/stacked_observations.py
qgallouedec: This function is supposed to work by passing an observation space as input. This makes it a bit strange, as it is a bit of a cross between a class method and an instance method, in that it depends on `self.n_stack` and `self.repeat_axis`, but not on `self.observation_space`. Anyway, I think it should be removed, and if there is someone who actually uses it, then the correspodance that works in all cases (potentially with a different observation_space than the one of the instance) is the code that is proposed in the docstring.

qgallouedec:
> was it used somewhere? 

Just in  `stack_observation_space`; it's not used elsewhere in sb3, nor in sb3-contrib or rl-zoo3.

stable_baselines3/common/vec_env/vec_frame_stack.py
qgallouedec: Just in `stack_observation_space`; it's not used elsewhere in sb3, nor in sb3-contrib or rl-zoo3. It's hard to know if this arg is used by codes that depend on sb3.

araffin (commented):
I removed the stacked dict obs and simplify some code that was a bit duplicated.

I'm also not sure why you added all the `# pytype: disable=attribute-error`, I tried removing them and it does work locally.
And pytype doesn't use the type stubs if I recall, in comparison to mypy (so if mypy passes, should be fine).
Apart from that comment, the rest look good to me =)

qgallouedec:
> I'm also not sure why you added all the `# pytype: disable=attribute-error`, I tried removing them and it does work locally.
> And pytype doesn't use the type stubs if I recall, in comparison to mypy (so if mypy passes, should be fine).

Usually, we wonder why a bug appears when we have not changed anything. Here, I wonder why a bug disappears when we have not changed anything 🤔. So be it! LGTM too!

araffin:
i think i know what happened, pytype allow multiline skip, and that's what your comment did.

araffin (approved):

PR #1244, opened 2022-12-27
Changed installation prerequisites to python 3.7-3.10

sryu1:


sryu1:
Commit message says to 3.11 but it's to 3.10

araffin:
Hello,
SB3 does work with Python 3.11 (I'm using it locally, although it requires PyTorch nightly).
What the motivation behind that change? 
Python 3.7+ is the same as >= 3.7, so it includes 3.7 to 3.11.

qgallouedec:
What happens when you pip install sb3 with Python 3.11?

sryu1:
Doesn't sb3 only work for 3.7 to 3.10? I thought the documentation was for stable releases... Well, I guess it works in stable but in order for that to work, you need a unstable release of pytorch?

sryu1:
> What happens when you pip install sb3 with Python 3.11? (I mean, does it install PyTorch nightly automatically, or do you need to install it before?)

It doesn't install PyTorch nightly automatically so I'm guessing it'd have to be manually install it... @araffin could you please confirm?

araffin:
yes, anyway that's the recommended way to do. First install the correct PyTorch version (cuda/cpu only/nightly/...) , then install SB3.
Doing pip install works in most cases but always download stable pytorch with cuda which is not always needed/the right one.

qgallouedec:
By the way, this is a mistake right?

https://github.com/DLR-RM/stable-baselines3/blob/master/docs/guide/install.rst#L14

qgallouedec:
By the way, that's a mistake, isn't it?

https://github.com/DLR-RM/stable-baselines3/blob/e78ba6ffa40e76c06fce1b8047070c991fa5abef/docs/guide/install.rst#L14

sryu1:
You mean the Python version 3.6 or above part?

qgallouedec:
Yes

araffin:
not a mistake, more of an outdated statement.

sryu1:
Nevermind then.

Issue #1245, opened 2022-12-27
[Bug]: Very Poor A2C CPU Utilization


Bonifatius94:
### 🐛 Bug

When using the A2C algorithm with 50 vectorized environments on a Ryzen Threadripper machine with 12 cores / 24 threads, I'm getting this really poor, almost single-core CPU utilization.

![stablebaselines3_a2c_algo_poor_cpu_utilization](https://user-images.githubusercontent.com/29117597/209699619-e00de4dd-8971-4142-97ac-53ee0a85e968.jpg)

I've tested the A2C training algorithm with several Gym environments such as FrozenLake and Pong. Both didn't make better use of the CPU either (except for FrozenLake being very small and yielding a lot more steps/s which doesn't even rise a single core up to 100%).

### To Reproduce

See the README commands on my GitHub repository for the [robot_sf simulator](https://github.com/Bonifatius94/robot-sf). I'm using a dockerized training with NVIDIA GPU support. It shouldn't make a difference when using other Gym environments.

```sh
# clone the git repo
git clone --recurse-submodules https://github.com/Bonifatius94/robot-sf
cd robot-sf

# install docker and docker-compose
sudo apt-get update && sudo apt-get install -y docker.io docker-compose
sudo usermod -aG docker $USER && reboot

# install cURL for sending HTTP requests
sudo apt-get update && sudo apt-get install -y curl

# register the NVIDIA Docker PPA as apt source
distribution=$(. /etc/os-release;echo $ID$VERSION_ID) \
    && curl -s -L https://nvidia.github.io/nvidia-docker/gpgkey | sudo apt-key add - \
    && curl -s -L https://nvidia.github.io/nvidia-docker/$distribution/nvidia-docker.list \
        | sudo tee /etc/apt/sources.list.d/nvidia-docker.list

# install NVIDIA Docker from the official NVIDIA PPA
sudo apt-get update && sudo apt-get install -y nvidia-docker2

# build the docker image and launch the training
docker-compose build && docker-compose run \
    robotsf-cuda python ./scripts/training.py
```

I've also tested with common Gym environments like FrozenLake or Pong. FrozenLake doesn't even rise one CPU core up to 100% and Pong has about the same behavior as my own training environment (one core at 100%, other cores on idle). So, you can use Pong for your investigations if you don't wanna use my custom Gym environment; it should have about the same properties.

```py
import gym
from stable_baselines3 import A2C
from stable_baselines3.common.env_util import make_vec_env

[...]

Issue #413, opened 2021-04-30
[Bug] excessive CPU on 


weepingwillowben:
### 🐛 Bug

Excessive CPU utilization with CnnPolicy.

### To Reproduce

On a GCP deep learning instance, the following code snippet uses 8 full CPUs during training. Note that on our private servers, this does not occur. Not clear whether this is due to cloud virtualization or to the more recent version of cuda (11.1) on GCP.

```python
from stable_baselines3 import PPO
from stable_baselines3.common.atari_wrappers import AtariWrapper
import gym
import supersuit as ss

env = gym.make("SpaceInvadersNoFrameskip-v4")
env = AtariWrapper(env)

model = PPO("CnnPolicy", env)
model.learn(total_timesteps=2000000)
model.save("policy")
```

setting `torch.set_num_threads(1)` or `OMP_NUM_THREADS=1` or `MKL_NUM_THREADS=1` all successfully run the code in a single thread.

### Expected behavior

We expect this to be single threaded or to be using a small number of threads when CUDA is enabled. The fact that it doesn't is a bit disturbing, as it suggests that perhaps some operation is running on CPU instead of GPU.

### System Info

Describe the characteristic of your environment:
 * installed version 1.0 with pip
 * cuda 11.0
 * Cudnn 8.0.5
 * pytorch 1.7.1
 * python 3.7

Only seems to happen on a GCP instance

### Checklist

- [ ] I have checked that there is no similar [issue](https://github.com/DLR-RM/stable-baselines3/issues) in the repo (**required**)
- [ ] I have read the [documentation](https://stable-baselines3.readthedocs.io/en/master/) (**required**)
- [ ] I have provided a minimal working example to reproduce the bug (**required**)

Miffyli:
Have you checked that CUDA is available in PyTorch? SB3 uses `torch.is_cuda_available()` internally to see if it should use GPU or CPU with default parameters (GPU if CUDA is available). If CUDA is not available, CPU is used which could lead to high usage.

Note that some operations are still done on CPU, e.g. data storing/processing and environment itself, so CPU is not completely underutilized.

weepingwillowben:
Yes, CUDA is avaliable, and I have checked that it is indeed using the GPU when running nvidia-smi

weepingwillowben:
Indeed, it runs much faster when using the GPU.

Miffyli:
[...]

Issue #121, opened 2020-07-24
SubprocVecEnv performance compared to gym.vector.async_vector_env


kargarisaac:
Hi,

I'm trying to use `SubProcVecEnv` to create a vectorized environment and use it in my own PPO implementation. I have a couple of questions about the performance of this vectorization and hyperparameters.

I have a 28 cores CPU on my system and an RTX 2080Ti GPU. When I use `gym.vector.async_vector_env` to create vectorized envs, it is 3 to 6 times faster than `SubProcVecEnv` from stable_baselines3. 

In `SubProcVecEnv`, when I set the number of threads using `torch.set_num_threads(28)` all the cores are involved but again it is almost two times slower than using `torch.set_num_threads(10)`.

I also did all the comparisons by creating 100 parallel envs. I'm not sure how would I set this number of envs and torch number of threads. I think this slower performance compared to `gym.vector.async_vector_env` is because of the bad hyperparameters I used.

What are the parameters I can use to get the best performance? which parameters are the most important ones?

Thank you

araffin:
Hello,
Did you take a look at our tutorial?
https://github.com/araffin/rl-tutorial-jnrr19/tree/sb3

especially this part: https://colab.research.google.com/github/araffin/rl-tutorial-jnrr19/blob/sb3/3_multiprocessing.ipynb
Did you try with a DummyVecEnv (and also setting num threads to a lower number, see #90 )

kargarisaac:
Thank you @araffin for the links.

I get the same speed with `DummyVecEnv` by using 10 torch threads or without setting the number of threads. Based on the colab results, it seems more environments don't have a higher return. I need to test more. It is very time consuming and comparing is not that easy. 

But I cannot understand why higher number of threads and using all cores is slower.

Thank you

AlessandroZavoli:
I have a similar behavior with stable-baselines (tensorflow).
It seems to me that when the environment is quite fast, using parallel environment is not effective.
As an example, on a IBM power pc, I have a 2x speedup factor using 128core vs 4core.

It might be that most of the time is spent by the "learner" and not by the "worker/experience-collecting" processes, but I might be wrong

Miffyli:
> But I cannot understand why higher number of threads and using all cores is slower.

[...]

PR #1250, opened 2022-12-28
Add scaling section to A2C documentation

Bonifatius94:
As requested by @araffin and @qgallouedec, I've added a documentation section on how to scale the A2C algorithm with SubprocVecEnv.

## Description
It's a new section in the rst documentation file for the A2C algorithm that covers scaling training steps throughput with multiple CPU cores.

## Motivation and Context
The question arose because the formerly outlined example didn't scale with CPU cores. See issue #1245.

closes https://github.com/DLR-RM/stable-baselines3/issues/1245.

## Types of changes

## Checklist
Your CI pipeline of the PR will cover this, no worries.


docs/modules/a2c.rst
```diff
@@ -76,6 +76,49 @@ Train a A2C agent on ``CartPole-v1`` using 4 environments.
...
+Scaling with CPU Cores
+----------------------
+
+The A2C algorithm allows for scaling with multiple CPU cores through parallel processing.
+To enable this feature, you'll have to override the vector environment implementation
+instantiated by ``make_vec_env()`` such that it creates a ``SubprocVecEnv``.
+
+The default ``DummyVecEnv`` implementation will only run a single thread for processing actions
+and collecting observations / rewards. With ``SubprocVecEnv``, each environment is running in its
+own thread, facilitating parallel processing for superior CPU utilization.
```
qgallouedec, line 88: I would replace this by ``"(See :ref:`Vectorized Environments <vec_env>`)"``

Bonifatius94: I agree, putting a link for further information on VecEnvs is a good idea. 

But I wouldn't leave out the short summary on how the different VecEnvs work. I think it fits there quite well.

But yeah, if you don't like it due to duplication for ease of maintainability, go ahead and delete it. I don't mind.

qgallouedec:
Thank you for contributing @Bonifatius94 ! I'll review it soon.

docs/modules/a2c.rst
araffin: you could also link to the Notebook about multiprocessing

Bonifatius94:
@qgallouedec I've added the cross reference you've posted. Don't know if it works though. I'm not so familiar with rst files tbh.

docs/modules/a2c.rst
Bonifatius94: @araffin Feel free to make your suggestion. I'm done for my part. You can do whatever you want with my changes 😄

qgallouedec:
I have simplified the section dedicated to this clarification quite a bit. (I prefer to keep the documentation as synthetic as possible, although I do think that adding this clarification is justified). Actually, I think most of what you describe doesn't apply only to A2C and so I prefer to link this to the more general documentation part about it. I also corrected and simplified the python code. @araffin could you review?

qgallouedec:
To avoid having to build the doc.

image

Bonifatius94:
One question: Why do you think turning the GPU off is beneficial?

I know, for processing fully-connected NN layers, it doesn't really matter whether the compute is carried out on the CPU or GPU. But if that's the case, then I would assume that disabling the GPU isn't relevant at all to performance optimization :thinking: 

Maybe we should test this before merging the docs update?

qgallouedec:
Using the example code (with 16 workers), I get 6966 FPS with GPU and 6880 without. So it seems to be equivalent.

Bonifatius94:
Hmm yeah, I'd say leave the "disable GPU" part out? :thinking: It might be very misleading to users that are not so familiar with the SB3 codebase when the actual benefit is coming from SubprocVecEnv :thinking:

qgallouedec:
> Using the example code (with 16 workers), I get 6966 FPS with GPU and 6880 without. So it seems to be equivalent.

@araffin can you take a look?

araffin (approved):
Look fine for me. The actual benefit here is hard to see as cartpole is a very simple/fast env (and the dummy vec env might be faster, see the notebook where we explore different settings). The CPU/GPU transfer cost might be minimal too.

Bonifatius94:
> Look fine for me. The actual benefit here is hard to see as cartpole is a very simple/fast env (and the dummy vec env might be faster, see the notebook where we explore different settings). The CPU/GPU transfer cost might be minimal too.

If you have a hypothesis when and why disabling the GPU might be better, then try to set up an experiment and prove it like a scientist :wink: In case you're not familiar with optimization philosophies, have a look at this video: https://youtu.be/pgoetgxecw8?t=79 :smile:

I fear you're practicing "fake optimization" where you just say "I saw one guy on the internet do this, so it has to be that way" instead of actually measuring. If there's no real-world case backing your statement, your hypothesis most likely doesn't apply to the concrete piece of code you're trying to optimize :sweat_smile: (I know it's hard not to fall into the fake optimization trap from my own experience. But there's really no shame in getting fooled because PCs are very complicated machines :smile:)

Please consider measuring the performance and maybe even setting up some performance benchmarks such that you'll notice which changes even affect the performance. Starting to measure instead of just guessing is crucial :smile:

qgallouedec:
I agree with you in principle. In this case, disabling the CPU will have two impacts: 
- the calculations will be done on the CPU, which, for non-image observations, is equivalent in terms of time (see https://github.com/DLR-RM/stable-baselines3/pull/1250#issuecomment-1404849705) and,
- there will be no CPU-GPU and GPU-CPU transfer. This transfer time is in fact far from negligible (see https://github.com/DLR-RM/stable-baselines3/pull/1218 where I measure it only with images). All in all, we can assume that when not using images, working on CPU is generally faster.

But assuming is not proving. I would be happy to have this performance analysis applied specifically to A2C. However, this is not a priority at the moment. In the meantime, we have to choose what you put in the doc. In my opinion, and considering the concordant clues in this specific case, it is best to put this trick in the doc, even if we discover that there is no performance gain later on.

araffin:
> "I saw one guy on the internet do this, so it has to be that way"

my statement was more based on experience (and interaction with users: https://github.com/DLR-RM/stable-baselines3/issues/1124, https://github.com/DLR-RM/stable-baselines3/issues/413, https://github.com/DLR-RM/stable-baselines3/issues/121)

but if you want a concrete experiment:

GPU
```
python -m rl_zoo3.train --algo a2c --env BipedalWalkerHardcore-v3 --eval-freq -1 --verbose 0 -P  --vec-env subproc
```
vs
CPU only (using too many threads)
```
CUDA_VISIBLE_DEVICES= python -m rl_zoo3.train --algo a2c --env BipedalWalkerHardcore-v3 --eval-freq -1 --verbose 0 -P  --vec-env subproc 
```
vs
CPU only (using one thread only to remove communication overhead)
```
OMP_NUM_THREADS=1 CUDA_VISIBLE_DEVICES= python -m rl_zoo3.train --algo a2c --env BipedalWalkerHardcore-v3 --eval-freq -1 --verbose 0 -P --vec-env subproc
```

Gives me:

GPU:  8,500 it/s
CPU only: 7, 200 it/s
CPU only (one thread): 9,950 it/s

(I'm using a 16 core machine with NVIDIA GeForce RTX 2080 GPU)

For environment with image input, deactivating the GPU will most probably slow down or not show any benefit (I tested on Breakout).

Issue #1150, opened 2022-11-02
Training exceeds total_timesteps


timosturm:
### ❓ Question

Consider this setup:
```python
import stable_baselines3
import gym
from stable_baselines3 import DQN, A2C, PPO
#from sb3_contrib import ARS, TRPO

env = gym.make('MountainCar-v0')

seed = 42
verbose = 1
timesteps = 10_000

DQN("MlpPolicy", env, verbose=verbose, seed=seed).learn(total_timesteps=timesteps) # 9600
A2C("MlpPolicy", env, verbose=verbose, seed=seed).learn(total_timesteps=timesteps) # 10_000
PPO("MlpPolicy", env, verbose=verbose, seed=seed).learn(total_timesteps=timesteps) # 10_240
#ARS("MlpPolicy", env, verbose=verbose, seed=seed).learn(total_timesteps=timesteps) # 12_800
#TRPO("MlpPolicy", env, verbose=verbose, seed=seed).learn(total_timesteps=timesteps) # 10_240
```

The problem is, that some of the agents train for (much) more time steps than specified. This changes depending on the number of timesteps set, e.g., DQN trains for exactly `100_000` time steps if specified. DQN often seems to train for fewer steps and PPO for more steps than specified.

This behavior is also bad when using the `EvaluationCallback`, because for some algorithms we do more (time consuming) evaluations than requested and for DQN we miss the last evaluation(s).

The question was asked before, [here](https://github.com/DLR-RM/stable-baselines3/issues/978) but no real solution was provided. Also setting `reset_num_timesteps=False` does not change anything (and I am not sure what it is supposed to change). I also tested this for different gym environments, but the problem persists.

What is the reason for this behavior? Can it be changed?

### Checklist

- [X] I have checked that there is no similar [issue](https://github.com/DLR-RM/stable-baselines3/issues) in the repo
- [X] I have read the [documentation](https://stable-baselines3.readthedocs.io/en/master/)
- [X] If code there is, it is minimal and working
- [X] If code there is, it is formatted using the [markdown code blocks](https://help.github.com/en/articles/creating-and-highlighting-code-blocks) for both code and stack traces.

araffin:
Hello,

Related to https://github.com/DLR-RM/stable-baselines3/issues/1059, probably duplicate of https://github.com/DLR-RM/stable-baselines3/issues/457

It is because of how the algorithms work.
For short:
[...]

PR #1260, opened 2023-01-05
Fix `ProgressBarCallback` under-reporting

### LINKED ISSUE #1259: [Bug]: `ProgressBarCallback` doesn't match *actual* `total_timesteps`
### 🐛 Bug

In a similar way to Issue #1150, a `ProgressBarCallback` can be stepped more/fewer times than the `total_timesteps` value provided to an algorithm's `.learn(...)` method.
Screenshot from 2023-01-03 16-46-34

As `tqdm.rich` doesn't always *flush* when `self.pbar.update` is called, `ProgressBarCallback` seems to typically under-report the number of steps actually processed (as seen above). However, this can be easily fixed by adding a `self.pbar.refresh()` call to the `ProgressBarCallback._on_training_end` method.
Screenshot from 2023-01-03 18-22-17

To fix the over-reporting of the final (`PPO`) progress bar, I think the `total` value passed into the callback's constructor (file: stable_baselines3/common/callbacks.py, line: 680) should be first rounded-up to a multiple of  
the algorithm's `n_steps`. This should then correctly report the algorithms progress while collecting experience - without this, I can imagine a very large rollout or computationally slow environment, resulting in a long wait without useful progress updates...

More generally, for vectorized environments, I think the `total` should also be rounded-up to a multiple of the callback's `training_env.num_envs`, as its `._on_step()` method advances the displayed progress by more than one. I believe this would be a one/two line change (plus, maybe a new unittest), which I'd be very happy to open!

### To Reproduce

```python
import gym
from stable_baselines3 import DQN, A2C, PPO

env = gym.make('MountainCar-v0')

DQN("MlpPolicy", env, seed=42).learn(total_timesteps=10_000, progress_bar=True) 
A2C("MlpPolicy", env, seed=42).learn(total_timesteps=10_000, progress_bar=True)
PPO("MlpPolicy", env, seed=42).learn(total_timesteps=10_000, progress_bar=True)
```

_No response_

### System Info

- OS: Linux-5.15.0-56-generic-x86_64-with-glibc2.35 # 62-Ubuntu SMP Tue Nov 22 19:54:14 UTC 2022
- Python: 3.10.6
- Stable-Baselines3: 1.7.0a11
- PyTorch: 1.13.0+cu117
- GPU Enabled: True
- Numpy: 1.23.5
- Gym: 0.21.0

### Checklist

dominicgkerr:
## Description
Added logic to `ProgressBarCallback` that gives a *better* estimate of the total number of timesteps that an algorithm processes, allowing the progress bar to more accurately report the remaining proportion of training steps. 

Note, environments with variable length episodes may still result in over/under-reported progress (potentially motivating the addition of an alternative `total_episodes` argument to `.learn(...)`, to compliment `total_timesteps`, which `ProgressBarCallback` could report the proportion of *episodes* remaining instead?) Likely out-of-scope of this PR...

Also, the *logic* for estimating the number of training timesteps might be better moved into `OffPolicyAlgorithm._setup_learn` and/or `OnPolicyAlgorithm._setup_learn`? It looks like this method already adjusts a provided `total_timesteps` for previously completed training, so could also round-up to the nearest `update_freq`/`n_steps` size? The function looks like it will need a bit of reordering though, so that the estimate can be calculated before the `ProgressBarCallback` is initialised in `BaseAlgorithm._setup_learn`...

## Motivation and Context
Closes #1259. 

## Types of changes

## Checklist
      See https://github.com/DLR-RM/stable-baselines3/issues/1259#issuecomment-1370202824


stable_baselines3/common/callbacks.py
```diff
@@ -675,15 +676,43 @@ def __init__(self) -> None:
...
-        # Initialize progress bar
+
+        # Get the total number of timesteps to do this training session
+        total = self.locals["total_timesteps"]
+
         # Remove timesteps that were done in previous training sessions
-        self.pbar = tqdm(total=self.locals["total_timesteps"] - self.model.num_timesteps)
+        total -= self.model.num_timesteps
+
+        # Round timesteps up-to a multiple of the algorithm's batch size
```
araffin, line 686: I'm not sure about that part.
As stated in https://github.com/DLR-RM/stable-baselines3/issues/1150, it is normal that some algorithms do more steps than expected.
The progress bar is only a rough indicator of how long a training will take and `total_timesteps` is a lower bound, so I think I would prefer to update the documentation (to make it clearer) rather than having some hacky code in there (that cover most cases but not all cases as you have probably noticed).

araffin: @qgallouedec what do you think?

qgallouedec: I agree. Under-reporting should be fixed, but not over-reporting because there is no such thing as over-reporting.

araffin (approved):
LGTM, thanks =)

PR #1321, opened 2023-02-07
Fix image-based normalized env loading

### LINKED ISSUE #1320: [Bug]: Image-based normalized envrionements failed to load
### 🐛 Bug

Loading normalized image-based fails.

### To Reproduce

```sh
python -m rl_zoo3.train -env MiniGrid-FourRooms-v0 --algo ppo --gym-packages gym_minigrid -n 10000
python -m rl_zoo3.push_to_hub --env MiniGrid-FourRooms-v0 --algo ppo -f logs -orga qgallouedec
```

```shell
python -m rl_zoo3.push_to_hub --env MiniGrid-FourRooms-v0 --algo ppo -f logs -orga qgallouedec
Loading latest experiment, id=2
Loading logs/ppo/MiniGrid-FourRooms-v0_2/MiniGrid-FourRooms-v0.zip
Loading running average
with params: {'norm_obs': True, 'norm_reward': False}
Wrapping the env in a VecTransposeImage.
  File "/home/qgallouedec/miniconda3/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/home/qgallouedec/miniconda3/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/qgallouedec/rl-baselines3-zoo/rl_zoo3/push_to_hub.py", line 383, in <module>
    model = ALGOS[algo].load(model_path, env=eval_env, custom_objects=custom_objects, device=args.device, **kwargs)
  File "/home/qgallouedec/stable-baselines3/stable_baselines3/common/base_class.py", line 689, in load
    check_for_correct_spaces(env, data["observation_space"], data["action_space"])
  File "/home/qgallouedec/stable-baselines3/stable_baselines3/common/utils.py", line 228, in check_for_correct_spaces
    raise ValueError(f"Observation spaces do not match: {observation_space} != {env.observation_space}")
ValueError: Observation spaces do not match: Dict(image:Box([[[-10. -10. -10.]
  [-10. -10. -10.]
  [-10. -10. -10.]
  [-10. -10. -10.]
  [-10. -10. -10.]
  [-10. -10. -10.]
  [-10. -10. -10.]]

 [[-10. -10. -10.]
  [-10. -10. -10.]
  [-10. -10. -10.]
  [-10. -10. -10.]
  [-10. -10. -10.]
  [-10. -10. -10.]
  [-10. -10. -10.]]

 [[-10. -10. -10.]
  [-10. -10. -10.]
  [-10. -10. -10.]
  [-10. -10. -10.]
  [-10. -10. -10.]
  [-10. -10. -10.]
  [-10. -10. -10.]]

 [[-10. -10. -10.]
  [-10. -10. -10.]
  [-10. -10. -10.]
  [-10. -10. -10.]
  [-10. -10. -10.]
  [-10. -10. -10.]
  [-10. -10. -10.]]

 [[-10. -10. -10.]
  [-10. -10. -10.]
  [-10. -10. -10.]
  [-10. -10. -10.]
  [-10. -10. -10.]
  [-10. -10. -10.]
  [-10. -10. -10.]]

 [[-10. -10. -10.]
  [-10. -10. -10.]
  [-10. -10. -10.]
  [-10. -10. -10.]
  [-10. -10. -10.]
  [-10. -10. -10.]
  [-10. -10. -10.]]

 [[-10. -10. -10.]
  [-10. -10. -10.]
  [-10. -10. -10.]
  [-10. -10. -10.]
  [-10. -10. -10.]
  [-10. -10. -10.]
  [-10. -10. -10.]]], [[[10. 10. 10.]
  [10. 10. 10.]
  [10. 10. 10.]
  [10. 10. 10.]
  [10. 10. 10.]
  [10. 10. 10.]
  [10. 10. 10.]]

 [[10. 10. 10.]
  [10. 10. 10.]
  [10. 10. 10.]
  [10. 10. 10.]
  [10. 10. 10.]
  [10. 10. 10.]
  [10. 10. 10.]]

 [[10. 10. 10.]
  [10. 10. 10.]
  [10. 10. 10.]
  [10. 10. 10.]
  [10. 10. 10.]
  [10. 10. 10.]
  [10. 10. 10.]]

 [[10. 10. 10.]
  [10. 10. 10.]
  [10. 10. 10.]
  [10. 10. 10.]
  [10. 10. 10.]
  [10. 10. 10.]
  [10. 10. 10.]]

 [[10. 10. 10.]
  [10. 10. 10.]
  [10. 10. 10.]
  [10. 10. 10.]
  [10. 10. 10.]
  [10. 10. 10.]
  [10. 10. 10.]]

 [[10. 10. 10.]
  [10. 10. 10.]
  [10. 10. 10.]
  [10. 10. 10.]
  [10. 10. 10.]
  [10. 10. 10.]
  [10. 10. 10.]]

 [[10. 10. 10.]
  [10. 10. 10.]
  [10. 10. 10.]
  [10. 10. 10.]
  [10. 10. 10.]
  [10. 10. 10.]
  [10. 10. 10.]]], (7, 7, 3), float32)) != Dict(image:Box([[[0 0 0 0 0 0 0]
  [0 0 0 0 0 0 0]
  [0 0 0 0 0 0 0]
  [0 0 0 0 0 0 0]
  [0 0 0 0 0 0 0]
  [0 0 0 0 0 0 0]
  [0 0 0 0 0 0 0]]

 [[0 0 0 0 0 0 0]
  [0 0 0 0 0 0 0]
  [0 0 0 0 0 0 0]
  [0 0 0 0 0 0 0]
  [0 0 0 0 0 0 0]
  [0 0 0 0 0 0 0]
  [0 0 0 0 0 0 0]]

 [[0 0 0 0 0 0 0]
  [0 0 0 0 0 0 0]
  [0 0 0 0 0 0 0]
  [0 0 0 0 0 0 0]
  [0 0 0 0 0 0 0]
  [0 0 0 0 0 0 0]
  [0 0 0 0 0 0 0]]], [[[255 255 255 255 255 255 255]
  [255 255 255 255 255 255 255]
  [255 255 255 255 255 255 255]
  [255 255 255 255 255 255 255]
  [255 255 255 255 255 255 255]
  [255 255 255 255 255 255 255]
  [255 255 255 255 255 255 255]]

 [[255 255 255 255 255 255 255]
  [255 255 255 255 255 255 255]
  [255 255 255 255 255 255 255]
  [255 255 255 255 255 255 255]
  [255 255 255 255 255 255 255]
  [255 255 255 255 255 255 255]
  [255 255 255 255 255 255 255]]

 [[255 255 255 255 255 255 255]
  [255 255 255 255 255 255 255]
  [255 255 255 255 255 255 255]
  [255 255 255 255 255 255 255]
  [255 255 255 255 255 255 255]
  [255 255 255 255 255 255 255]
  [255 255 255 255 255 255 255]]], (3, 7, 7), uint8))
```

### System Info

- OS: Linux-5.15.0-58-generic-x86_64-with-glibc2.35 # 64-Ubuntu SMP Thu Jan 5 11:43:13 UTC 2023
- Python: 3.9.12
- Stable-Baselines3: 1.8.0a4
- PyTorch: 1.11.0+cpu
- GPU Enabled: False
- Numpy: 1.24.1
- Gym: 0.21.0

### Checklist

qgallouedec:
## Description

Closes #1320 

The error comes from here, when called to create eval env:

https://github.com/DLR-RM/stable-baselines3/blob/489b1fdaf2958ba5d63ed5a90f20540cd4107b13/stable_baselines3/common/vec_env/vec_normalize.py#L162

At this point, `venv` is not normalized, but `self` is. When calling `VecEnvWrapper.__init__`, `self.observation_space` is overwritten with the unnormalized space. To fix this, we just need to replace 

```python
self.observation_space = venv.observation
```
by

```python
venv.observation_space = self.observation_space
```

## Motivation and Context

## Types of changes

## Checklist

- [x] I have opened an associated PR on the [SB3-Contrib repository](https://github.com/Stable-Baselines-Team/stable-baselines3-contrib) (if necessary)
- [x] I have opened an associated PR on the [RL-Zoo3 repository](https://github.com/DLR-RM/rl-baselines3-zoo) (if necessary)


stable_baselines3/common/vec_env/vec_normalize.py
```diff
@@ -159,7 +160,11 @@ def set_venv(self, venv: VecEnv) -> None:
         """
         if self.venv is not None:
             raise ValueError("Trying to set venv of already initialized VecNormalize wrapper.")
-        VecEnvWrapper.__init__(self, venv)
+        self.venv = venv
+        self.num_envs = venv.num_envs
+        venv.observation_space = self.observation_space
+        venv.action_space = self.action_space
+        self.class_attributes = dict(inspect.getmembers(self.__class__))
```
araffin, line 167: why is that needed?

stable_baselines3/common/vec_env/vec_normalize.py
```diff
@@ -159,7 +160,11 @@ def set_venv(self, venv: VecEnv) -> None:
         """
         if self.venv is not None:
             raise ValueError("Trying to set venv of already initialized VecNormalize wrapper.")
-        VecEnvWrapper.__init__(self, venv)
+        self.venv = venv
+        self.num_envs = venv.num_envs
+        venv.observation_space = self.observation_space
```
araffin, line 165: is that needed, if we don't overwrite `self.observation_space`, we should be fine

qgallouedec: It fails here otherwise.

https://github.com/DLR-RM/stable-baselines3/blob/d82b3d566273cee08f840dd3ead9eabe8efd17a6/stable_baselines3/common/vec_env/vec_normalize.py#L169-L170

stable_baselines3/common/vec_env/vec_normalize.py
qgallouedec:
```python
VecEnvWrapper.__init__(self, venv) 
```

is equivalent to

```python
self.venv = venv
self.num_envs = venv.num_envs
self.observation_space = venv.observation_space
self.action_space = self.action_space
self.class_attributes = dict(inspect.getmembers(self.__class__))
```

The only change in this PR is that we change 

```python
self.observation_space = venv.observation_space
self.action_space = self.action_space
```

to 

```python
venv.observation_space = self.observation_space
venv.action_space = self.action_space
```

Then:

```python
self.venv = venv  # is needed because the environment is not set
self.num_envs = venv.num_envs  # is needed because the loaded env, and the one to set don't necessarily share the number of env, see the unit test.
venv.observation_space = self.observation_space  # see other comment
venv.action_space = self.action_space  # not sure if it's needed, but it could lead to the same break for the same reason
self.class_attributes = dict(inspect.getmembers(self.__class__))  # not sure if its need, but it keeps consistency with the old implementation
```

stable_baselines3/common/vec_env/vec_normalize.py
qgallouedec: > I see, it's just that modifying the underlying obs space doesn't sounds like a good side effect (once we do that, there is no way to know what was the original obs space).

Agree

For what it's worth, all tests pass when venv spaces aren't modified and check skipped

qgallouedec: > do you think there is an easy way to know when we modify the obs space? (so we can skip/relax the check instead of overriding the venv obs space)

I'll think about it. Maybe a warning could suit:

```txt
the loaded model is normalized and the observation space does not match the set model. We strongly recommend to use this new send instead:
venv = model.get_env()
```
But this is not the ideal solution, because there would be no way to "solve" the warning for the user.

qgallouedec: On second thought, I think it's just the check that has no reason to be. There is no reason for the wrapped observation space to be the same as the unwrapped. I would simply remove those lines and put no warning.

stable_baselines3/common/vec_env/vec_normalize.py
qgallouedec: https://github.com/DLR-RM/stable-baselines3/blob/489b1fdaf2958ba5d63ed5a90f20540cd4107b13/stable_baselines3/common/vec_env/base_vec_env.py#L249-L262

And then `VecEnv.__init__`:

https://github.com/DLR-RM/stable-baselines3/blob/489b1fdaf2958ba5d63ed5a90f20540cd4107b13/stable_baselines3/common/vec_env/base_vec_env.py#L59-L62

stable_baselines3/common/vec_env/vec_normalize.py
araffin: for me, only the shape check makes sense

stable_baselines3/common/vec_env/vec_normalize.py
araffin: oh, I forgot about `self.class_attributes = dict(inspect.getmembers(self.__class__))`, yes we need that when doing recursion.

stable_baselines3/common/utils.py
```diff
@@ -230,6 +230,22 @@ def check_for_correct_spaces(env: GymEnv, observation_space: spaces.Space, actio
         raise ValueError(f"Action spaces do not match: {action_space} != {env.action_space}")
 
 
+def check_shape_equal(space1: Union[spaces.Box, spaces.Dict], space2: Union[spaces.Box, spaces.Dict]) -> None:
```
araffin, line 233: you should probably add a `keys` argument as the `norm_obs_keys` we have in `VecNormalize`, otherwise it will crash at load time, no?

qgallouedec: You mean, for example, if the subspaces are not all Box? I don't have any example of such env in mind but you're right. I would rather skip test for these case, see my next commit.

araffin: on a second thought, I remember gym defines `shape` for all spaces (https://github.com/openai/gym/blob/v0.21.0/gym/spaces/space.py#L22) but I would double check that.

>  if the subspaces are not all Box? 

yes, a mixed space like one is an indicator (or a switch, do discrete) and the other one is the rest of the state (box).
that's also why we have the `norm_obs_keys` to allow skipping spaces we don't want to normalize or that can't be normalized.

qgallouedec: > on a second thought, I remember gym defines `shape` for all spaces (https://github.com/openai/gym/blob/v0.21.0/gym/spaces/space.py#L22) so we should be fine ,but I would double check that.

Correct! I didn't know.

It should work that way.

tests/test_vec_normalize.py
```diff
@@ -244,7 +249,7 @@ def test_obs_rms_vec_normalize():
     assert np.allclose(env.ret_rms.mean, 5.688, atol=1e-3)
 
 
-@pytest.mark.parametrize("make_env", [make_env, make_dict_env])
+@pytest.mark.parametrize("make_env", [make_env, make_dict_env, make_image_env])
```
araffin, line 252: what is the difference with `test_save_load_vecnormalized_image` ? (in the sense, do we need the two tests or does `test_save_load_vecnormalized_image` cover the main use cases already?)

araffin: it seems that we could remove `test_save_load_vecnormalized_image` in favor of that one? (or the error from the issue won't be raised?)

qgallouedec: The difference is that the error is raised in `model.load` in `test_save_load_vecnormalized_image`. Here, is it raised when calling `check_vec_norm_equal`. I agree that we could remove `test_save_load_vecnormalized_image` in favor of this one.

qgallouedec: > or the error from the issue won't be raised?

It would be, since the check (previously checking space equality, now checking shape equality) is in `VecNormalize.set_env` which is called by `VecNormalize.load`.

stable_baselines3/common/vec_env/vec_normalize.py
```diff
@@ -159,10 +160,12 @@ def set_venv(self, venv: VecEnv) -> None:
         """
         if self.venv is not None:
             raise ValueError("Trying to set venv of already initialized VecNormalize wrapper.")
-        VecEnvWrapper.__init__(self, venv)
+        self.venv = venv
+        self.num_envs = venv.num_envs
+        self.class_attributes = dict(inspect.getmembers(self.__class__))
```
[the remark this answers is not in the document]
qgallouedec: No, the `class_attribute` is deleted when the object is pickled. 
https://github.com/DLR-RM/stable-baselines3/blob/9fe176ef6b4524fa62107b07ad2b0c6d3eb7a3a3/stable_baselines3/common/vec_env/vec_normalize.py#L134

Then, `VecNormalize` isn't really initialized, but just unpickled. So you need this line.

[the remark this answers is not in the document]
araffin: ah alright

PR #1361, opened 2023-03-06
Switch to `pyproject.toml` and `ruff`

### LINKED ISSUE #1294: [Feature Request] Switch to pyproject.toml
### 🚀 Feature

`setup.cfg` is being deprecated in favor of `pyproject.toml` for both config and dependencies.
We are already using one in SB3 contrib, the idea was be to finish the transfer, we will still keep the `setup.py` though (for `pip install -e .` installs).

### Motivation

This would allows having black config automatically and allow to have tools like `ruff` (fast alternative to flake8).

### Pitch

_No response_

### Alternatives

_No response_

### Additional context

_No response_

### Checklist

araffin:
## Description

Similar to https://github.com/Stable-Baselines-Team/stable-baselines3-contrib/pull/162
but switching to fast flake8 alternative (https://github.com/charliermarsh/ruff) that does support `pyproject.toml`.

Also solves the issue found in https://github.com/HumanCompatibleAI/imitation/pull/682 and https://github.com/DLR-RM/rl-baselines3-zoo/pull/362
EDIT: it might not fully solve it as `pip install autorom` is different from `pip install autorom[accept-rom-license]`, I might have to specify a `extra_no_roms`.

cc @charliermarsh (author of ruff)

## Motivation and Context

closes #1294 

## Types of changes

## Checklist


pyproject.toml
```diff
@@ -1,35 +1,39 @@
-[metadata]
-# This includes the license file in the wheel.
-license_files = LICENSE
-project_urls =
-    Code = https://github.com/DLR-RM/stable-baselines3
-    Documentation = https://stable-baselines3.readthedocs.io/
+[tool.ruff]
```
charliermarsh, line 1: 👍 👍

Makefile
```diff
@@ -14,29 +14,22 @@ type: pytype mypy
...
-	# see https://lintlyci.github.io/Flake8Rules/
-	flake8 ${LINT_PATHS} --count --select=E9,F63,F7,F82 --show-source --statistics
-	# exit-zero treats all errors as warnings.
-	flake8 ${LINT_PATHS} --count --exit-zero --statistics
-
-ruff:
-	# stop the build if there are Python syntax errors or undefined names
-	# see https://lintlyci.github.io/Flake8Rules/
+	# see https://www.flake8rules.com/
 	ruff ${LINT_PATHS} --select=E9,F63,F7,F82 --show-source
```
qgallouedec, line 18: `--select` seems to redundant with the config in `pyproject.toml`

qgallouedec: Also, I think it is `ruff check`, not just `ruff`. (It's weird that it doesn't raise an error).

qgallouedec: Why do you need to run 2 checks instead of one? Ie, why not just `--show-source --exit-zero`

araffin: > eems to redundant with the config in pyproject.toml
> Why do you need to run 2 checks instead of one? Ie, why not just --show-source --exit-zero

The first command allow to check errors that would crash the code (undefined variables, syntax error, missing imports, ...).
The second one (define in the config) is about warnings and nice to have, the code will run.
So it's ok that the second check fail (hence `--exit-zero`) but not the first one.

> Also, I think it is ruff check, not just ruff. (It's weird that it doesn't raise an error).

`check` is the default command so `ruff` and `ruff check` are equivalent

araffin: btw, as ruff is really fast, I should probably put the linter before the type checking, no?

qgallouedec: > btw, as ruff is really fast, I should probably put the linter before the type checking, no?

agree

qgallouedec: > `--select` seems to redundant with the config in `pyproject.toml`

@araffin can you just clarify this?

araffin: If I don't use select, it's gonna run all the rules defines in the `pyproject.toml` and fail at the first warning.

qgallouedec (approved):

Issue #1392, opened 2023-03-19
[Bug]: DictRolloutBuffer doesn't reshape action


younik:
### 🐛 Bug

DictRolloutBuffer doesn't reshape action in `add` method, making multi-dimensional action space fail to save due to shape mismatch.

### To Reproduce

```python
from stable_baselines3 import PPO
import gym

class MyEnv(gym.Env):
    def __init__(self):
        self.action_space = gym.spaces.Box(-1, 1, shape=(10, 10))
        self.observation_space = gym.spaces.Dict({
            "foo": gym.spaces.Box(-1, 1, shape=(1,)),
            "foo2": gym.spaces.Box(-1, 1, shape=(1,))
        })

    def reset(self):
        return self.observation_space.sample()
    
    def step(self, action):
        return self.observation_space.sample(), 0, False, {}

env = MyEnv()
model = PPO("MultiInputPolicy", env)
model.learn(1000)
```

### Relevant log output / Error message

```shell
/usr/local/lib/python3.9/dist-packages/stable_baselines3/common/buffers.py in add(self, obs, action, reward, episode_start, value, log_prob)
    735             self.observations[key][self.pos] = obs_
    736 
--> 737         self.actions[self.pos] = np.array(action).copy()
    738         self.rewards[self.pos] = np.array(reward).copy()
    739         self.episode_starts[self.pos] = np.array(episode_start).copy()

ValueError: could not broadcast input array from shape (10,10) into shape (1,100)
```

### Checklist

- [X] I have checked that there is no similar [issue](https://github.com/DLR-RM/stable-baselines3/issues) in the repo
- [X] I have read the [documentation](https://stable-baselines3.readthedocs.io/en/master/)
- [X] I have provided a minimal working example to reproduce the bug
- [X] I've used the [markdown code blocks](https://help.github.com/en/articles/creating-and-highlighting-code-blocks) for both code and stack traces.

araffin:
Hello,
thanks for spotting the bug, I would be happy to receive a PR that fixes it ;)

Btw, for the proposed env, it would be better (and faster) to flatten the observation dict to a single 1D vector.

araffin:
>  thanks for spotting the bug, I would be happy to receive a PR that fixes it ;)

PS: if you do so, please don't use master branch for your fork, that would prevent maintainers to push edits to it.

younik:
> > thanks for spotting the bug, I would be happy to receive a PR that fixes it ;)
> 
[...]

PR #971, opened 2022-07-19
Handling multi-dimensional action spaces

qgallouedec:
## Description

closes #970

To reproduce the error:

```python
import gym
import numpy as np
from gym.spaces import Box

class TestEnv(gym.Env):
    def __init__(self) -> None:
        super().__init__()
        self._i = 0
        self.action_space = Box(low=-1, high=1, shape=(5,5))
        self.observation_space = Box(low=-1, high=1, shape=(5,))

    def render(self, mode="console"):
        pass

    def close(self):
        pass

    def reset(self) -> np.array:
        return self.observation_space.sample()

    def step(self, action: np.array):
        self._i +=1
        obs = np.zeros(5)
        done = self._i % 10 == 0
        info = {}
        reward = 0
        return obs, reward, done, info


if __name__ == "__main__":
    from stable_baselines3 import DDPG
    
    env = TestEnv()
    model = DDPG("MlpPolicy", env)
    model.learn(total_timesteps=1000)
```

## Motivation and Context

- [x] I have raised an issue to propose this change ([required](https://github.com/DLR-RM/stable-baselines3/blob/master/CONTRIBUTING.md) for new features and bug fixes)

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation (update in the documentation)

## Checklist:

- [x] I've read the [CONTRIBUTION](https://github.com/DLR-RM/stable-baselines3/blob/master/CONTRIBUTING.md) guide (**required**)
- [x] I have updated the changelog accordingly (**required**).
- [ ] My change requires a change to the documentation.
- [x] I have updated the tests accordingly (*required for a bug fix or a new feature*).
- [ ] I have updated the documentation accordingly.
- [x] I have reformatted the code using `make format` (**required**)
- [x] I have checked the codestyle using `make check-codestyle` and `make lint` (**required**)
- [x] I have ensured `make pytest` and `make type` both pass. (**required**)
- [x] I have checked that the documentation builds using `make doc` (**required**)

Note: You can run most of the checks using `make commit-checks`.

Note: we are using a maximum length of 127 characters per line

PR #1395, opened 2023-03-19
Fix: Reshape action in DictRolloutBuffer

younik:
(Same as https://github.com/DLR-RM/stable-baselines3/pull/1393, but on a new branch)

## Description
Reshape the action before adding it to the DictRolloutBuffer since the buffer size is action shape flattened.

## Motivation and Context
Closes https://github.com/DLR-RM/stable-baselines3/issues/1392

## Types of changes

## Checklist


tests/test_buffers.py
```diff
@@ -44,7 +44,7 @@ class DummyDictEnv(gym.Env):
     """
 
     def __init__(self):
-        self.action_space = spaces.Box(1, 5, (1,))
+        self.action_space = spaces.Box(1, 5, (10, 7))
```
araffin, line 47: We might need to add a dedicated test for multi-dim action space.
As I'm a bit surprised it works for the rest of the algorithms (the line in the buffers was for the discrete actions initially), could you check that the output of the agent is correctly reshaped? (when using multiple env too)

It's in https://github.com/DLR-RM/stable-baselines3/blob/master/stable_baselines3/common/policies.py#L333

I know it runs but I'm a bit afraid the order might be wrong due to the reshape (the agent only outputs 1D actions and then it gets reshape into multi-D array)

araffin:
To explain my comment in https://github.com/DLR-RM/stable-baselines3/pull/1395/files#r1141459772:
when you reshape data, you need to be careful that the order is kept.
For instance, if you are not careful, when using multiple env, you might end up interleaving episodes: https://github.com/DLR-RM/stable-baselines3/blob/master/stable_baselines3/common/buffers.py#L59-L71

One quick check we could do is to artificially create an env with a multi-dim action space (let's say one of PyBullet/MuJoCo env) and check that the result is exactly the same for a given seed (and when using multi env).

younik:
> To explain my comment in https://github.com/DLR-RM/stable-baselines3/pull/1395/files#r1141459772: when you reshape data, you need to be careful that the order is kept. For instance, if you are not careful, when using multiple env, you might end up interleaving episodes: https://github.com/DLR-RM/stable-baselines3/blob/master/stable_baselines3/common/buffers.py#L59-L71
> 
> One quick check we could do is to artificially create an env with a multi-dim action space (let's say wrapping one of PyBullet/MuJoCo env) and check that the result is exactly the same for a given seed (and when using multi env).

I tested using the following code:

```python
import gym
import numpy as np
from stable_baselines3.common.vec_env import SubprocVecEnv, VecMonitor
import wandb
from wandb.integration.sb3 import WandbCallback


class MyWrapper(gym.Wrapper):
    def __init__(self, env, multi_dim) -> None:
        super().__init__(env)
        self.multi_dim = multi_dim
        if self.multi_dim:
            self.action_space = gym.spaces.Box(
                -1.0, 1.0, (2, 2, 2), dtype=np.float32
            )
        self.observation_space = gym.spaces.Dict({
            "dummy": gym.spaces.Box(-1, 1, (1,)),
            "obs": self.env.observation_space
        })

    def reset(self):
        obs = super().reset()
        return {"dummy": np.zeros(1), "obs": obs}

    def step(self, action):
        if self.multi_dim:
            action = action.flatten()
        obs, rew, done, info = super().step(action)
        return {"dummy": np.zeros(1), "obs": obs}, rew, done, info



def make_env(multi_dim):
    def thunk():
        env = gym.make("Ant-v3")
        env = MyWrapper(env, multi_dim=multi_dim)
        return env
    return thunk


if __name__ == "__main__":
    multi_dim = False
    run = wandb.init(name=f'multi_dim_{multi_dim}', project="SB3Test", sync_tensorboard=True)

    env = SubprocVecEnv([make_env(multi_dim) for _ in range(4)])
    env = VecMonitor(env)
    
    model = PPO("MultiInputPolicy", env, seed=563, verbose=1, tensorboard_log=f"runs/{run.id}")
    model.learn(total_timesteps=100_000, callback=WandbCallback())
```

is this what you meant? 

In this case, the results look identical: https://wandb.ai/younis/SB3Test

araffin:
> is this what you meant?

thanks for checking =)
yes, that's what I meant (I also meant it for the `MlpPolicy` case, no dict).

younik:
> > is this what you meant?
> 
> thanks for checking =) yes, that's what I meant (I also meant it for the `MlpPolicy` case, no dict).

I also tried with no dict (without overriding `observation_space` and using `MlpPolicy`) and it still works (I added the results on the same wandb project).

younik:
Is there anything else to do from my side? 

Do you want a custom test? The current ones use a dictionary mapping from dict buffers to one env.
How would you structure the test? Just a smoke test to check that it works?

araffin:
> Is there anything else to do from my side?

sorry, was quite busy in the last days, will try to take a look at it soon.

younik:
> > Is there anything else to do from my side?
> 
> sorry, was quite busy in the last days, will try to take a look at it soon.

No worries; just checking :)

araffin:
I forgot ^^ but we have tests and support for multi-dim action space since https://github.com/DLR-RM/stable-baselines3/pull/971 (done by @qgallouedec )

araffin (approved):
LGTM, thanks =)

PR #1567, opened 2023-06-21
Fix small mistake in docstring

pelmers:
The docstring for the `write_row` method was copied from the `close` method. This commit updates the summary to match what the method does more closely.

Background: I'm doing a research study on function comments that may be incorrect. You can find more info and a short and optional anonymous survey [here](https://forms.office.com/e/yHv4PRXM2i).
 
Thanks for reviewing!


araffin (approved):
lgtm, thanks =)

PR #1486, opened 2023-05-05
`vec_env`s fix `seed()` causing a reset

Kallinteris-Andreas:
## Description
fixes issue :https://github.com/DLR-RM/stable-baselines3/issues/1481 
by keeping a `self.seeds` variable inside the `env`

- removed compat code
- added some tests to check for deterministic behavior
- fixed the type hints for subprocess vec env
 - updated the env checker as seed parameter is now required
 - updated the tests accordingly

## Motivation and Context
- [X] I have raised an issue to propose this change ([required](https://github.com/DLR-RM/stable-baselines3/blob/master/CONTRIBUTING.md) for new features and bug fixes)

closes #1481 

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation (update in the documentation)

## Checklist

- [x] I've read the [CONTRIBUTION](https://github.com/DLR-RM/stable-baselines3/blob/master/CONTRIBUTING.md) guide (**required**)
- [ ] I have updated the changelog accordingly (**required**).
- [ ] My change requires a change to the documentation.
- [x] I have updated the tests accordingly (*required for a bug fix or a new feature*).
- [ ] I have updated the documentation accordingly.
- [x] I have opened an associated PR on the [SB3-Contrib repository](https://github.com/Stable-Baselines-Team/stable-baselines3-contrib) (if necessary)
- [ ] I have opened an associated PR on the [RL-Zoo3 repository](https://github.com/DLR-RM/rl-baselines3-zoo) (if necessary)
- [x] I have reformatted the code using `make format` (**required**)
- [ ] I have checked the codestyle using `make check-codestyle` and `make lint` (**required**)
- [ ] I have ensured `make pytest` and `make type` both pass. (**required**)
- [ ] I have checked that the documentation builds using `make doc` (**required**)

Note: You can run most of the checks using `make commit-checks`.

Note: we are using a maximum length of 127 characters per line

PR #1606, opened 2023-07-13
Add support for setting `options` at reset with `VecEnv`

### LINKED ISSUE #1523: [Feature Request] Gymnasium: Add `options` argument to VecEnv.reset()
### 🚀 Feature

Hello,

As `gymnasium` implements the very useful `options` argument in the `.reset()` method, it would be nice to propagate it to the `VecEnv.reset()` notably to reset the environment in **various states** (useful for instance during evaluation phases).

Same for the `seed` argument  even though an other workaround has been chosen.

Would it cause issues if we just add `option=None` to the signatures ?

PS: When writing _gymnasium_ environments now, IDE (e.g PyCharm) triggers warnings if the signature of the custom environments don't match with the abstract class, so it encourages defining `options` and `seed` arguments.

Have a good day,
Thanks,

### Motivation

_No response_

### Pitch

_No response_

### Alternatives

_No response_

### Additional context

_No response_

### Checklist

ReHoss:
## Description

Hi, I suggest the following code to solve #1523 

Basically, it is inspired by #1486 logic for seeds.

Note, only one option dictionary is duplicated to be passed to all VecEnv environments when `done` or `truncated` is reached.

### Change Details

I followed the logic from #1486 where the options dictionary are stored in a list whose size is the number of environments in the `VecEnv`.

However, I had to update the `make_vec_env` function signature with an `options` argument.

I updated all the tests signatures for `reset` and added one `wrong_reset` method.

### About pytest

However, running `make pytest` the following errors triggered:

> test_fps_logger[A2C] fails

In fact my Ubuntu 22 laptop with i9 proc was in mode eco, so the computational power was limited.
With standard power the _fps_ test passed. Should this test be made easier?

However, I still have a rendering problem ? Note my nvidia card is deactivated.

> tests/test_vec_envs.py::test_render[DummyVecEnv] make: *** [Makefile:5: pytest] Error 1

## Motivation and Context

closes #1523

## Types of changes

## Checklist


ReHoss:
I did not change the `env_checker`

https://github.com/DLR-RM/stable-baselines3/blob/d68ff2e17f2f823e6f48d9eb9cee28ca563a2554/stable_baselines3/common/env_checker.py#L375C1-L375C1



Do we want to check if the signature also contains `options` ?

araffin:
Hello,
thanks for the PR =), I'm taking some days off, will come back to you mid next week normally ;)

araffin:
> However, I had to update the make_vec_env function signature with an options argument.

I don't think that should be needed.

> Do we want to check if the signature also contains options ?

No, compared to `seed`, `options` is not critical but a nice-to-have feature.
So, it should be optional (so should be passed to the reset only if present).
Also, I think it would make sense to allow to pass a list of dict instead of a single dict.

ReHoss:
> > However, I had to update the make_vec_env function signature with an options argument.
> 
> I don't think that should be needed.

As  #1486 augment `make_vec_env` with a seed initialisation (see [*]), I applied the same logic with the options (despite the unique `options` is diffused to all environments for now).
Why would we not do that here? To avoid adding an optional parameter?



[*] : https://github.com/DLR-RM/stable-baselines3/pull/1486/files#diff-d5121994cbf23725faf52c1bbc0e5b1baeff4ab35ce176eefaf24b63c09f9cb0R125-R128

> > Do we want to check if the signature also contains options ?
> 
> No, compared to `seed`, `options` is not critical but a nice-to-have feature. So, it should be optional (so should be passed to the reset only if present). Also, I think it would make sense to allow to pass a list of dict instead of a single dict.


I'll be back by the end off August:)

stable_baselines3/common/vec_env/base_vec_env.py
```diff
@@ -283,6 +291,21 @@ def seed(self, seed: Optional[int] = None) -> Sequence[Union[None, int]]:
         self._seeds = [seed + idx for idx in range(self.num_envs)]
         return self._seeds
 
+    def set_options(self, options: Optional[Dict] = None) -> None:
+        """
+        Set environment options for all environments, based on an unique dict.
+        WARNING: Those options will only be passed to the environment at the next reset.
+
+        :param options: A dictionary of environment options to pass to each environment at the next reset.
+        :return:
```
ReHoss, line 300: Need to add _:return:_ information

stable_baselines3/common/vec_env/dummy_vec_env.py
```diff
@@ -73,10 +73,13 @@ def step_wait(self) -> VecEnvStepReturn:
 
     def reset(self) -> VecEnvObs:
         for env_idx in range(self.num_envs):
-            obs, self.reset_infos[env_idx] = self.envs[env_idx].reset(seed=self._seeds[env_idx])
+            obs, self.reset_infos[env_idx] = self.envs[env_idx].reset(
+                seed=self._seeds[env_idx], options=self._options[env_idx]
```
ReHoss, line 77: Not sure, but this way would need `reset` to have an `options` argument while `kwargs` could be more appropriate.
but which `reset` is called?

araffin: >  argument while kwargs could be more appropriate.

yes, for options you will need to use kwargs which should be an empty dict most of the time (so it is not a hard requirement).

>  but which reset is called? I guess it is the env_fns from self.envs so probably the true env reset.

yes, the env reset is called.

kmcminn:
quick +1 for this feature. Also can we make it easier to extend step() also with options/kwargs?

araffin (approved):

Issue #622, opened 2021-10-22
[Feature Request] RAINBOW


araffin:
**Important Note: We do not do technical support, nor consulting** and don't answer personal questions per email.
Please post your question on the [RL Discord](https://discord.com/invite/xhfNqQv), [Reddit](https://www.reddit.com/r/reinforcementlearning/) or [Stack Overflow](https://stackoverflow.com/) in that case.

### 🚀 Feature

Implement RAINBOW (DQN + all extensions: noisy-net, double dqn, dueling, c51, prioritized experience replay).

closes #487 if it is implemented.

### Motivation

This would allow to have DQN with all extensions while keeping DQN code simple.

- [x] I have checked that there is no similar [issue](https://github.com/DLR-RM/stable-baselines3/issues) in the repo (**required**)

Sino-Huang:
Any updates about the rainbow implementation?

araffin:
> Any updates about the rainbow implementation?

Contributions are welcomed ;) (if you do so, please read the contributing guide from SB3-Contrib, it explains how to test new algorithms)
It is planned but not a priority.

jenschwich:
Hi there, I hope you're having a great day. I was reading through #487 and it leads me here. I see that the double-DQN is now available as an exercise. I looked through sb3 and sb3-contrib repositories, but I couldn't find any code or reference to it. Would you be able to point me in the direction of the material or pseudo code that explains the implementation of double-DQN in sb3? I'd really appreciate it. Thank you!

qgallouedec:
Hi @jenschwich, check out https://github.com/Stable-Baselines-Team/rl-colab-notebooks/blob/sb3/README.md

noahfarr:
Hi,
do you already know how long it will take until Rainbow is available? I would like to use it for a project I'm currently working on. Is there already a preliminary implementation that maybe just hasn't been tested thoroughly? Is there something I could do to contribute to Rainbow being available faster?

araffin:
> Is there already a preliminary implementation that maybe just hasn't been tested thoroughly? 

If you read this issue and the linked one, you will see components that have been implemented but not fully tested.

> Is there something I could do to contribute to Rainbow being available faster?

Yes, we need help for https://github.com/DLR-RM/stable-baselines3/pull/1622 to debug and be able to reproduce the original results (and maybe re-write this implementation).
[...]

Issue #48, opened 2020-06-09
Performance check (Continuous Actions)


araffin:
Check that the algorithms reach expected performance.
This was already done prior to v0.5 for the [gSDE paper](https://arxiv.org/abs/2005.05719) but as we made big changes, it is good to check that again.

## SB2 vs SB3 (Tensorflow Stable-Baselines vs Pytorch Stable-Baselines3)

- [x] A2C (6 seeds)

[a2c.pdf](https://github.com/DLR-RM/stable-baselines3/files/5029909/a2c.pdf)
[a2c_ant.pdf](https://github.com/DLR-RM/stable-baselines3/files/5029912/a2c_ant.pdf)
[a2c_half.pdf](https://github.com/DLR-RM/stable-baselines3/files/5029913/a2c_half.pdf)
[a2c_hopper.pdf](https://github.com/DLR-RM/stable-baselines3/files/5029914/a2c_hopper.pdf)
[a2c_walker.pdf](https://github.com/DLR-RM/stable-baselines3/files/5029915/a2c_walker.pdf)

- [x] PPO (6 seeds)

[ppo.pdf](https://github.com/DLR-RM/stable-baselines3/files/5029043/ppo.pdf)
[ant_ppo.pdf](https://github.com/DLR-RM/stable-baselines3/files/5029045/ant_ppo.pdf)
[half_ppo.pdf](https://github.com/DLR-RM/stable-baselines3/files/5029046/half_ppo.pdf)
[hopper_ppo.pdf](https://github.com/DLR-RM/stable-baselines3/files/5029047/hopper_ppo.pdf)
[ppo_walker.pdf](https://github.com/DLR-RM/stable-baselines3/files/5029048/ppo_walker.pdf)

- [x] SAC (3 seeds)

[sac.pdf](https://github.com/DLR-RM/stable-baselines3/files/5024549/sac.pdf)
[sac_ant.pdf](https://github.com/DLR-RM/stable-baselines3/files/5024550/sac_ant.pdf)
[sac_half.pdf](https://github.com/DLR-RM/stable-baselines3/files/5024552/sac_half.pdf)
[sac_hopper.pdf](https://github.com/DLR-RM/stable-baselines3/files/5024553/sac_hopper.pdf)
[sac_walker.pdf](https://github.com/DLR-RM/stable-baselines3/files/5024554/sac_walker.pdf)

- [x] TD3 (3 seeds)

[...]

PR #1622, opened 2023-07-23
Prioritized experience replay

### LINKED ISSUE #1242: Prioritized Experience Replay for DQN
### 🚀 Feature

Prioritized Experience Replay for DQN

### Motivation

_No response_

### Pitch

_No response_

### Alternatives

_No response_

### Additional context

_No response_

### Checklist

AlexPasqua:
## Description

Implementation of prioritized replay buffer for DQN.
Closes #1242 

## Motivation and Context

In accordance with #1242

## Types of changes

## Checklist

### DIFF
--- a/stable_baselines3/dqn/prioritized_replay_buffer.py
+++ b/stable_baselines3/dqn/prioritized_replay_buffer.py
@@ -1,6 +1,8 @@
 import random
-from typing import Optional, Union
+from typing import Any, Dict, List, Optional, Union
+import warnings
 from gymnasium import spaces
+import numpy as np
 import torch as th
 from stable_baselines3.common.buffers import ReplayBuffer
 from stable_baselines3.common.type_aliases import ReplayBufferSamples
@@ -25,7 +27,7 @@ class SumTree:
     def p_total(self):
         return self.nodes[0]
 
-    def update(self, data_idx, value):
+    def update(self, data_idx: int, value: float):
         idx = data_idx + self.size - 1  # child index in tree array
         change = value - self.nodes[idx]
 
@@ -36,7 +38,7 @@ class SumTree:
             self.nodes[parent] += change
             parent = (parent - 1) // 2
 
-    def add(self, value, data):
+    def add(self, value: float, data):
         self.data[self.count] = data
         self.update(self.count, value)
 
@@ -88,8 +90,13 @@ class PrioritizedReplayBuffer(ReplayBuffer):
         n_envs: int = 1,
         alpha: float = 0.6,
         beta: float = 0.4,
+        optimize_memory_usage: bool = False,
     ):
         super().__init__(buffer_size, observation_space, action_space, device, n_envs)
+
+        # TODO: check this
+        if optimize_memory_usage:
+            warnings.warn("PrioritizedReplayBuffer does not support optimize_memory_usage=True during sampling")
 
         # PER params
         self.eps = 1e-8  # minimal priority, prevents zero probabilities
@@ -99,6 +106,27 @@ class PrioritizedReplayBuffer(ReplayBuffer):
 
         # SumTree: data structure to store priorities
         self.tree = SumTree(size=buffer_size)
+
+        self.real_size = 0
+        self.count = 0
+    
+    def add(self,
+        obs: np.ndarray,
+        next_obs: np.ndarray,
+        action: np.ndarray,
+        reward: np.ndarray,
+        done: np.ndarray,
+        infos: List[Dict[str, Any]],
+    ) -> None:
+        # store transition index with maximum priority in sum tree
+        self.tree.add(self.max_priority, self.count)
+
+        # update counters
+        self.count = (self.count + 1) % self.buffer_size
+        self.real_size = min(self.buffer_size, self.real_size + 1)
+        
+        # store transition in the buffer
+        super().add(obs, next_obs, action, reward, done, infos)
     
     def sample(self, batch_size: int, env: Optional[VecNormalize] = None) -> ReplayBufferSamples:
         """
--- a/stable_baselines3/dqn/prioritized_replay_buffer.py
+++ b/stable_baselines3/dqn/prioritized_replay_buffer.py
@@ -1,0 +1,160 @@
+import random
+from typing import Optional, Union
+from gymnasium import spaces
+import torch as th
+from stable_baselines3.common.buffers import ReplayBuffer
+from stable_baselines3.common.type_aliases import ReplayBufferSamples
+from stable_baselines3.common.vec_env.vec_normalize import VecNormalize
+
+
+class SumTree:
+    """
+    SumTree data structure for Prioritized Replay Buffer.
+    This code is inspired by: https://github.com/Howuhh/prioritized_experience_replay
+
+    :param size: Max number of element in the buffer.
+    """
+    def __init__(self, size: int):
+        self.nodes = [0] * (2 * size - 1)
+        self.data = [None] * size
+        self.size = size
+        self.count = 0
+        self.real_size = 0
+
+    @property
+    def p_total(self):
+        return self.nodes[0]
+
+    def update(self, data_idx, value):
+        idx = data_idx + self.size - 1  # child index in tree array
+        change = value - self.nodes[idx]
+
+        self.nodes[idx] = value
+
+        parent = (idx - 1) // 2
+        while parent >= 0:
+            self.nodes[parent] += change
+            parent = (parent - 1) // 2
+
+    def add(self, value, data):
+        self.data[self.count] = data
+        self.update(self.count, value)
+
+        self.count = (self.count + 1) % self.size
+        self.real_size = min(self.size, self.real_size + 1)
+
+    def get(self, cumsum):
+        assert cumsum <= self.p_total
+
+        idx = 0
+        while 2 * idx + 1 < len(self.nodes):
+            left, right = 2*idx + 1, 2*idx + 2
+
+            if cumsum <= self.nodes[left]:
+                idx = left
+            else:
+                idx = right
+                cumsum = cumsum - self.nodes[left]
+
+        data_idx = idx - self.size + 1
+
+        return data_idx, self.nodes[idx], self.data[data_idx]
+
+    def __repr__(self):
+        return f"SumTree(nodes={self.nodes.__repr__()}, data={self.data.__repr__()})"
+
+
+class PrioritizedReplayBuffer(ReplayBuffer):
+    """
+    Prioritized Replay Buffer.
+    Paper: https://arxiv.org/abs/1511.05952
+    This code is inspired by: https://github.com/Howuhh/prioritized_experience_replay
+
+    :param buffer_size: Max number of element in the buffer
+    :param observation_space: Observation space
+    :param action_space: Action space
+    :param device: PyTorch device
+    :param n_envs: Number of parallel environments
+    :param alpha: How much prioritization is used (0 - no prioritization, 1 - full prioritization)
+    :param beta: To what degree to use importance weights (0 - no corrections, 1 - full correction)
+    """
+
+    def __init__(
+        self,
+        buffer_size: int,
+        observation_space: spaces.Space,
+        action_space: spaces.Space,
+        device: Union[th.device, str] = "auto",
+        n_envs: int = 1,
+        alpha: float = 0.6,
+        beta: float = 0.4,
+    ):
+        super().__init__(buffer_size, observation_space, action_space, device, n_envs)
+
+        # PER params
+        self.eps = 1e-8  # minimal priority, prevents zero probabilities
+        self.alpha = alpha  # determines how much prioritization is used, alpha = 0 corresponding to the uniform case
+        self.beta = beta  # determines the amount of importance-sampling correction, beta = 1 fully compensate for the non-uniform probabilities
+        self.max_priority = self.eps  # priority for new samples, init as eps
+
+        # SumTree: data structure to store priorities
+        self.tree = SumTree(size=buffer_size)
+    
+    def sample(self, batch_size: int, env: Optional[VecNormalize] = None) -> ReplayBufferSamples:
+        """
+        Sample elements from the prioritized replay buffer.
+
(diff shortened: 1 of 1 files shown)


AlexPasqua:
@araffin could you (or anyone) please have a look at the 2 pytype errors? I don't quite understand how to fix them

stable_baselines3/dqn/prioritized_replay_buffer.py
```diff
@@ -0,0 +1,221 @@
...
+        weights = weights / weights.max()
+
+        env_indices = np.random.randint(0, high=self.n_envs, size=(len(sample_idxs),))
+
+        if self.optimize_memory_usage:
+            next_obs = self._normalize_obs(
+                self.observations[(np.array(sample_idxs) + 1) % self.buffer_size, env_indices, :], env
+            )
+        else:
+            next_obs = self._normalize_obs(self.next_observations[sample_idxs, env_indices, :], env)
```
araffin, line 212: pytype thinks this value might None, which is apparently true (see `buffers.py` for solutions)

stable_baselines3/dqn/prioritized_replay_buffer.py
```diff
@@ -0,0 +1,221 @@
...
+
+        batch = (
+            self._normalize_obs(self.observations[sample_idxs, env_indices, :], env),
+            self.actions[sample_idxs, env_indices, :],
+            next_obs,
+            self.dones[sample_idxs],
+            self.rewards[sample_idxs],
+        )
+        return ReplayBufferSamples(*tuple(map(self.to_torch, batch)))
```
araffin, line 221: you can probably ignore pytype here

stable_baselines3/dqn/prioritized_replay_buffer.py
AlexPasqua: `self.next_observations` might be None if `self.optimize_memory_usage` is True, but then we wouldn't enter the `else` at line 212.
The same thing happens in `buffers.py`, so I don't quite understand where the difference is

AlexPasqua: Hello @araffin,
could you have a look at this (or the PR in general) when you have time?

After that I'll work on the tests, documentation etc. in order to complete the PR.
Cheers!

araffin: I'm pretty busy right now (writing paper for end of the month) and I was waiting for the tests to pass, but I'll try to have a look beginning of October ;)

AlexPasqua:
Thanks @araffin !
Out of curiosity, may I ask why the switch between torch and numpy for the backend?

araffin:
> Thanks @araffin ! Out of curiosity, may I ask why the switch between torch and numpy for the backend?

to be consistent with the rest of the buffers and because PyTorch is not needed here (no gpu computation needed).

AlexPasqua:
Hello @araffin ,
as you moved the code to "common", I suppose you plan to make it usable in algorithms other than DQN. At this point, wouldn't it be clearer to put the code into `common/buffers.py`? Let me know, in case, I will move it there.

araffin:
> At this point, wouldn't it be clearer to put the code into common/buffers.py?

yes probably, but the most important thing for now is to test the implementation (performance test, check we can reproduce the results from the paper), document it and add additional tests/doc (for sumtree for instance).

araffin:
> performance test, check we can reproduce the results from the paper

After some initial test on Breakout following hyperparameters from the paper, the run didn't improve or worsen DQN performance so far...
I will try on other envs (it would be nice if you could help).

AlexPasqua:
> After some initial test on Breakout following hyperparameters from the paper, the run didn't improve or worsen DQN performance so far... I will try on other envs (it would be nice if you could help).

Thanks for starting to test it!
These days I'm travelling, and also writing a paper after work, but I'll try to squeeze some tests in

AlexPasqua:
I've also done some initial tests and it looks like PER might lead to a slightly faster convergence, for example on cartpole, but nothing super evident unfortunately.
Next I'd like to properly reproduce some of the paper's experiment, but computational power could become a bit of an issue for me

stable_baselines3/dqn/dqn.py
```diff
@@ -208,8 +209,20 @@ def train(self, gradient_steps: int, batch_size: int = 100) -> None:
...
+                # Weighted Huber loss using importance sampling weights
+                loss = (replay_data.weights * th.where(td_error < 1.0, 0.5 * td_error**2, td_error - 0.5)).mean()
+                # Update priorities, they will be proportional to the td error
+                assert replay_data.leaf_nodes_indices is not None, "Node leaf node indices provided"
+                self.replay_buffer.update_priorities(
+                    replay_data.leaf_nodes_indices, td_error, self._current_progress_remaining
+                )
+            else:
+                # Compute Huber loss (less sensitive to outliers)
+                loss = F.smooth_l1_loss(current_q_values, target_q_values)
```
qgallouedec, line 225: @AlexPasqua Ideally, we'd like to be able to associate it with all off-policy algo's without adaptation, but I don't see a simple way of doing it at this stage.
Also related, we had discussed not modifying DQN: https://github.com/Stable-Baselines-Team/stable-baselines3-contrib/pull/127#issuecomment-1381530029

richardjozsa:
Just a comment, I've tested this implementation with QR-DQN with Vecenv multiple environment but it fails because of the missing part.

But good job to start the work on it! I hope it will be merged soon! 👍

stable_baselines3/dqn/dqn.py
jbial: I'm interested in this PR. Since every algo-specific `train` method includes a `replay_buffer.sample` line, couldn't we just additionally add a `replay_buffer.update` line? The update function could take in the current and target q values whenever a value function is present or maybe even all the local variables. It would do nothing for the vanilla replay buffer. Would this be an acceptable modification?

qgallouedec: Thanks for your comment!
How do you handle the loss in your proposal?

jbial: If we want this to work for general off-policy algorithms, we could update the ReplayBufferSample-like classes to additionally include an `importance_sampling_weight` attribute which would be updated from the `replay_buffer.update` method.

Then I see two ways to handle the loss under this interface:

1. Estimate TD error from the loss as such:
```
losses = loss_fn(current_q_values, target_q_values, reduction='none')

# e.g. If loss is L2, then it's basically th.sqrt(loss). If loss is L1, td_error = loss
td_error = importance_sampling_weight * function_to_approx_td_error(losses)  

loss = losses.mean()
```
Obviously the downside of this is that it requires hand engineering for the different types of loss functions or priority metrics.

2. Make any value-based `train` methods "td-error" centric in the sense that we always compute `td_error = importance_sampling_weight * th.abs(current_q_values - target_q_values)` first, then the loss `loss = loss_fn(td_error)`. The downsides of this approach is that we cant use the pytorch api for computing the loss, and would have to write functions for those.

Either approach requires computing a `td_error` variable which unfortunately requires somewhat intrusive code changes. What do you think?

araffin: maybe to make things clearer: my plan is not to have for all algorithms, mainly for two reasons:

1. Keep the code concise (in fact, I would like to have RAINBOW and keep vanilla DQN, see https://github.com/DLR-RM/stable-baselines3/issues/622)
2. I don't think it works for entropy-RL algorithms (SAC and derivates), so it would be limited to DQN/QR-DQN and TD3

jbial: In that case, I'd be happy to contribute to the Rainbow DQN feature request. Dueling DQN and double DQN are straightforward to implement, and @araffin PER implementation seems to work. Also C51 is already implemented in contrib. We would just need n-step returns and noisy nets right?

araffin: "just" yes, I would be happy to receive such PR =)
the main thing is to benchmark the implementation and reproduce the published results.
This PR is also still open because I was not satisfied by the result of DQN + PER (I couldn't see significant different with respect to DQN).

araffin: One thing I had in mind was to implement CNN for SBX (https://github.com/araffin/sbx) in order to iterate faster and check the PER, but I had no time to do so until now...

jbial: Why don't we implement the toy environment from figure 1 of https://arxiv.org/pdf/1511.05952 as the PER benchmark? It would be a simpler initial check for correctness than the Atari environments

araffin: The toy environment can be a start for fast iteration and debugging, but what we learn in the past is that subtle bugs only show up when doing more complex task (see https://github.com/DLR-RM/stable-baselines3/issues/48 and https://github.com/DLR-RM/stable-baselines3/issues/47 where we found bugs like PyTorch and TF RMSProp are not the same)

jbial: I see, will definitely work towards it!

stable_baselines3/common/prioritized_replay_buffer.py
```diff
@@ -0,0 +1,259 @@
...
+        beta: float = 0.4,
+        final_beta: float = 1.0,
+        optimize_memory_usage: bool = False,
+        min_priority: float = 1e-8,
+    ):
+        super().__init__(buffer_size, observation_space, action_space, device, n_envs)
+
+        assert optimize_memory_usage is False, "PrioritizedReplayBuffer doesn't support optimize_memory_usage=True"
+
+        self.min_priority = 1e-8
```
janakact, line 129: Small mistake. Change this to `self.min_priority = min_priority` instead.  ;)

jbial:
I've just tried validating the implementation on blind cliffwalk and it seems much slower (~an order of magnitude) than the uniform replay buffer. The results below are for a one seed:
Screenshot 2024-05-27 at 10 20 10 PM

Not sure why this is. The details for blind cliffwalk are a bit vague from the paper (no code available as well), but I've tried to implement it as close to the description as possible.

Code for the test is in this gist:
https://gist.github.com/jbial/105299c00dc3bb7960f0f17f2fc4d6c9

stable_baselines3/common/prioritized_replay_buffer.py
```diff
@@ -0,0 +1,259 @@
...
+        # probability of sampling transition i as P(i) = p_i^alpha / \sum_{k} p_k^alpha
+        # where p_i > 0 is the priority of transition i.
+        probs = priorities / self.tree.total_sum
+
+        # Importance sampling weights.
+        # All weights w_i were scaled so that max_i w_i = 1.
+        weights = (self.size() * probs) ** -self.beta
+        weights = weights / weights.max()
+
+        # TODO: add proper support for multi env
```
richardjozsa, line 218: How could we add proper support for multiple envs? Is there any idea? Does the random line below could work?

araffin: not sure yet, the random line below might work but we need to check if it won't affect performance first.

araffin:
Some update from my part, I just added CNN support for SBX (SB3 + Jax) DQN, and it is 10x faster than the PyTorch equivalent: https://github.com/araffin/sbx/pull/49

That should allow to test and debug things more quickly on Atari (~1h40 for 10M steps instead of 15h =D)

Perf report: https://wandb.ai/openrlbenchmark/sbx?nw=nwuseraraffin (on-going)

stable_baselines3/common/prioritized_replay_buffer.py
araffin: https://github.com/araffin/sbx/pull/50/commits/b5ce09105aa8df2701e85d30f0584380b2ee19f8 should be better, see https://github.com/araffin/sbx/pull/50

araffin:
Some additional update: when trying to plug the PER implementation of this PR inside the Jax DQN implementation, the experience replay was the bottleneck (by a good margin, making things 40x slower...), so I investigated different ways to speed things up.

After playing with many different implementation (pure python, numpy, jax, jax jitted, ...), I decided to re-use the SB2 "SegmentTree" vectorized implementation and also implement proper multi-env support.
My current progress is here: https://github.com/araffin/sbx/pull/50

(still debugging, but at least I've got the first sign of life and this implementation is so much faster)

richardjozsa:
Hey @araffin , it is great to hear that. Does SBX/Jax means this much speed improvement? 

If you think it is ready for testing I can give a try, just let me know when it is ready to be tested. :)

araffin:
> Does SBX/Jax means this much speed improvement?

With the right parameters (see the exact command line argument for the RL Zoo in the OpenRL benchmark organization run on W&B), yes, around 10x faster.

> If you think it is ready for testing I can give a try, just let me know when it is ready to be tested. :)

SBX version is ready to be tested but so far, I didn't manage to see any gain from the PER. I also experienced some explosion in the qf value when using multiple env (so there is probably a bug here).
I'm also wondering if I need to implement double q-learning (easy) too to compare to the original paper.

richardjozsa:
When I tested this PR I also noticed an explosion loss,  in that time I felt that it is because of the tweaking here and there. and I also noticed that it doesn't give me any advantage over a normal buffer(and I used Dobule DQN, even tried duelling), but I tried to tweak an N-step buffer which had a strong effect on the learning, AFAIK N-step(multi step) is also part of Rainbow and giving substantial part of the success.

The key parts are the distributional, PER and N-step parts, as far as I understand the concept. The others are kinda tasks specific parts and can be detrimental to use them.

bilelsgh:
Ok, nice.
The PR for the propritized approzimation loss is pending, let me know if I can help for something.

nf2bi89:
@araffin is the idea to move PER to sbx entirely as it already has multi-env implemented which this version doesn't? or is multi-env support not intended for this version? Also if the prioritized approximation loss implementation is similar enough to the original PER at a reduced computational cost doesn't this solve the exact speed issue you are currently seeking plus it already works with multi-envs? I have been unable to compare them yet though as it appears to ignore the passed buffer class and doesn't have a working example.

PR #1685, opened 2023-09-18
`test_buffers.py::test_device` was not actually checking the device of tensors

rhaps0dy:
## Description
I added a failure in case of unknown type to `test_buffers.py::test_device`, and made it iterate correctly through Tensors and dicts.

## Motivation and Context
In `test_buffers.py::test_device`, the assertions that the device be the correct one would not be executed for `RolloutBuffer` and `DictRolloutBuffer`: the type of `value` is `RolloutBufferSamples` and `DictRolloutBufferSamples` respectively and so falls to the `else` case.

I have not raised an issue on purpose, why bother for a tiny bugfix.

## Types of changes

## Checklist:

### DIFF
--- a/tests/test_buffers.py
+++ b/tests/test_buffers.py
@@ -140,13 +140,16 @@ def test_device_buffer(replay_buffer_cls, device):
     if replay_buffer_cls in [RolloutBuffer, DictRolloutBuffer]:
         data = buffer.get(50)
     elif replay_buffer_cls in [ReplayBuffer, DictReplayBuffer]:
-        data = buffer.sample(50)
+        data = [buffer.sample(50)]
 
     # Check that all data are on the desired device
     desired_device = get_device(device).type
-    for value in list(data):
-        if isinstance(value, dict):
-            for key in value.keys():
-                assert value[key].device.type == desired_device
-        elif isinstance(value, th.Tensor):
-            assert value.device.type == desired_device
+    for minibatch in list(data):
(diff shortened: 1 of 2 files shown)


tests/test_buffers.py
```diff
@@ -140,13 +140,16 @@ def test_device_buffer(replay_buffer_cls, device):
...
 
     # Check that all data are on the desired device
     desired_device = get_device(device).type
-    for value in list(data):
-        if isinstance(value, dict):
-            for key in value.keys():
-                assert value[key].device.type == desired_device
-        elif isinstance(value, th.Tensor):
-            assert value.device.type == desired_device
+    for minibatch in list(data):
```
qgallouedec, line 147: There's only one pass in this loop, isn't there? If so, I think you can simply delete it and drop the brackets from line 143.

rhaps0dy: There are two passes actually, because `.get(50)` returns a Generator for two batches of size 50. There are 2 batches because the buffer is of size 100.

rhaps0dy: The reason line 143 needs the brackets is to make sure both branches (`.get()` and `[.sample()]` ) are iterables with the same type

qgallouedec: makes sens :)

araffin:
I could not edit this PR, so I'll push the updated version in a new PR

PR #1771, opened 2023-12-04
Create optimizer in `OnPolicyAlgorithm` only after the device is set

### LINKED ISSUE #1770: [Bug:] Cannot use the `fused` flag in default optimizer of PPO
The default Adam optimizer has a `fused` flag, which, according to the docs, is significantly faster than the default when used on CUDA. Using it with PPO generates an exception, which complains that the parameters are not of type CUDA.

The `fused` parameter can be specified to PPO using `policy_kwargs = dict(optimizer_kwargs={'fused': True})`.
But, the issue is in the following lines of code:
https://github.com/DLR-RM/stable-baselines3/blob/c8fda060d4bcb283eea6ddd385be5a46a54d3356/stable_baselines3/common/on_policy_algorithm.py#L133-L136

Before line 133 above, the correct device has been initialised in `self.device`. But, the `policy_class` is initialized without it in line 133, so it initialises with the `cpu` device, and that also initialises the optimizer with the `cpu` device. In line 136, the device of the `policy_class` is updated to the correct one, but by then it is too late, because the optimizer had already been initialized, and it thought the device was `cpu`.

This is a problem with the `fused` flag, because the Adam optimiser does check it and then double-checks `self.parameters()` to ensure they are of the correct type, and complains, in my case, that it is not of `cuda` type.

If the `policy_class` in line 133 above was passed the correct device (i.e. `self.device`) in the initialization in the first place, it could set it correctly before `MlpExtractor` gets initialized. `MlpExtractor` gets initialized to the parent class's device in the lines below:
https://github.com/DLR-RM/stable-baselines3/blob/c8fda060d4bcb283eea6ddd385be5a46a54d3356/stable_baselines3/common/policies.py#L568-L581

Here is the traceback I get:

```
  File "x.py", line 350, in <module>
    main(sys.argv)
  File "x.py", line 254, in main
    model = PPO(
            ^^^^
  File "/z/venv/lib64/python3.11/site-packages/stable_baselines3/ppo/ppo.py", line 164, in __init__
    self._setup_model()
  File "/z/venv/lib64/python3.11/site-packages/stable_baselines3/ppo/ppo.py", line 167, in _setup_model
    super()._setup_model()
  File "/z/venv/lib64/python3.11/site-packages/stable_baselines3/common/on_policy_algorithm.py", line 123, in _setup_model
    self.policy = self.policy_class(  # type: ignore[assignment]
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/z/venv/lib64/python3.11/site-packages/stable_baselines3/common/policies.py", line 857, in __init__
    super().__init__(
  File "/z/venv/lib64/python3.11/site-packages/stable_baselines3/common/policies.py", line 507, in __init__
    self._build(lr_schedule)
  File "/z/venv/lib64/python3.11/site-packages/stable_baselines3/common/policies.py", line 610, in _build
    self.optimizer = self.optimizer_class(self.parameters(), lr=lr_schedule(1), **self.optimizer_kwargs)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/z/venv/lib64/python3.11/site-packages/torch/optim/adam.py", line 60, in __init__
    raise RuntimeError("`fused=True` requires all the params to be floating point Tensors of "
RuntimeError: `fused=True` requires all the params to be floating point Tensors of supported devices: ['cuda', 'xpu', 'privateuseone'].
```

cmangla:
Attempt to fix #1770 in a fully backward compatible manner.

## Description
In PPO, the optimizer in the policy is created before the computation-device for the class is correctly set. This is a problem when the optimizer checks the target computation-device on initialization. This is a backward compatible fix.

## Motivation and Context
Fixes #1770 . One can now use the `fused` option in the Adam optimizer on CUDA devices, which, according to the documentation, is faster.
- [x] I have raised an issue to propose this change

## Types of changes
- [x] Bug fix

## Checklist

### DIFF
--- a/stable_baselines3/common/policies.py
+++ b/stable_baselines3/common/policies.py
@@ -441,6 +441,8 @@ class ActorCriticPolicy(BasePolicy):
         ``th.optim.Adam`` by default
     :param optimizer_kwargs: Additional keyword arguments,
         excluding the learning rate, to pass to the optimizer
+    :param _init_optimizer: If the optimizer should be initialised immediately. True by default.
+        When set to False, the optimizer should be explicitly initialized using ``._build_optimizer(...)``.
     """
 
     def __init__(
@@ -462,6 +464,7 @@ class ActorCriticPolicy(BasePolicy):
         normalize_images: bool = True,
         optimizer_class: Type[th.optim.Optimizer] = th.optim.Adam,
         optimizer_kwargs: Optional[Dict[str, Any]] = None,
+        _init_optimizer=True,
     ):
         if optimizer_kwargs is None:
             optimizer_kwargs = {}
@@ -530,7 +533,7 @@ class ActorCriticPolicy(BasePolicy):
         # Action distribution
         self.action_dist = make_proba_distribution(action_space, use_sde=use_sde, dist_kwargs=dist_kwargs)
 
-        self._build(lr_schedule)
+        self._build(lr_schedule, _init_optimizer)
 
     def _get_constructor_parameters(self) -> Dict[str, Any]:
         data = super()._get_constructor_parameters()
@@ -580,7 +583,7 @@ class ActorCriticPolicy(BasePolicy):
             device=self.device,
         )
 
-    def _build(self, lr_schedule: Schedule) -> None:
+    def _build(self, lr_schedule: Schedule, _init_optimizer=True) -> None:
         """
         Create the networks and the optimizer.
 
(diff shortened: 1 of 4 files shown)


cmangla:
@araffin This is ready for the CI tests now and potentially also to merge.

stable_baselines3/common/policies.py
```diff
@@ -885,6 +895,7 @@ def __init__(
         normalize_images: bool = True,
         optimizer_class: Type[th.optim.Optimizer] = th.optim.Adam,
         optimizer_kwargs: Optional[Dict[str, Any]] = None,
+        _init_optimizer=True,  # Currently unused, see PR #1771
```
cmangla, line 898: @araffin I'm currently testing enabling this one too. I will update this PR accordingly, hence switching it back to draft.

cmangla:
@araffin Looks good now

PR #1805, opened 2024-01-11
Fix missing seed / option args in dummy and subproc vec_env resets during stepping

### LINKED ISSUE #1790: [Bug]: Reset options ignored when resetting due to termination / truncation from within wrapper's `step`
### 🐛 Bug

Given a wrapped env, options passed with the [recommended way](https://stable-baselines3.readthedocs.io/en/master/guide/vec_envs.html) (`wrapped_env.set_options`) are ignored when reset is triggered by episode termination / truncation in `step_wait` of the env [wrapper](https://github.com/DLR-RM/stable-baselines3/blob/master/stable_baselines3/common/vec_env/dummy_vec_env.py#L70).

A (myopic) fix could be to pass `self._options[env_idx]` and `self._seeds[env_idx]` to the linked code above, or refactor a single-env resetting function to use both in `DummyVecEnv.step_wait` and `DummyVecEnv.reset`.

Apologies if this expected behavior. If so, what is the recommended way to pass reset options that affect the above resetting scenario?

### To Reproduce

```python
from stable_baselines3.common.vec_env import DummyVecEnv
import gymnasium as gym
import numpy as np


# dummy env
class CustomEnv(gym.Env):
    def __init__(self):
        super().__init__()
        self.action_space = gym.spaces.Discrete(3)
        self.observation_space = gym.spaces.Box(low=0, high=5, shape=(5,), dtype=np.uint8)

    # step terminates the episode
    def step(self, action):
        terminated = 1
        return np.zeros(5), 0, terminated, 0, {}

    # reset prints the options, if provided
    def reset(self, seed=None, options=None):
        if options is not None:
            print(" -- Options supplied:", options)
        return np.zeros(5), {}

# make and wrap the env
env = CustomEnv()
env = DummyVecEnv([lambda: env])

# resetting by invoking the wrapper function -- options are passed
print("Resetting by invoking DummyVecEnv.reset() :")
env.set_options({'opt': 1})
env.reset()

# reset by a terminating environment step:
# the wrapper step function calls self.envs[env_idx].reset(), which ignores self._options
print("Resetting by an episode-terminating invokation to DummyVecEnv.step() :")
env.set_options({'opt': 1})
env.step([0])

print("Done.")
```

```shell
Resetting by invoking DummyVecEnv.reset() :
 -- Options supplied: {'opt': 1}
Resetting by an episode-terminating invokation to DummyVecEnv.step() :
Done.
```

### System Info

- OS: Linux-6.6.7-arch1-1-x86_64-with-glibc2.34 # 1 SMP PREEMPT_DYNAMIC Thu, 14 Dec 2023 03:45:42 +0000
- Python: 3.8.15
- Stable-Baselines3: 2.2.1
- PyTorch: 1.13.1+cu117
- GPU Enabled: True
- Numpy: 1.24.1
- Cloudpickle: 2.2.1
- Gymnasium: 0.28.1

### Checklist

- [X] My issue does not relate to a custom gym environment. (Use the custom gym env template instead)
- [X] I have provided a [minimal and working](https://github.com/DLR-RM/stable-baselines3/issues/982#issuecomment-1197044014) example to reproduce the bug

npit:
## Description

1. Passed the appropriate elements from `self._seeds` and `self._options` to a "done" env that calls its `reset` function, in `DummyVecEnv.step_wait`
2. Added missing `options` argument in the `reset` functions of  `TimeDelayEnv` and `CustomSubClassedSpaceEnv` 
3. Modified `SubprocVecEnv` to send options and seeds arguments as in (1) in `step_async`.
4. Modified the `_worker` function's "step" case, to receive and pass these arguments into the env's `reset`, as in (1).

## Motivation and Context

Closes #1790

## Types of changes

## Checklist

### DIFF
--- a/stable_baselines3/common/vec_env/subproc_vec_env.py
+++ b/stable_baselines3/common/vec_env/subproc_vec_env.py
@@ -123,6 +123,7 @@ class SubprocVecEnv(VecEnv):
 
     def step_async(self, actions: np.ndarray) -> None:
         for remote, action, seed, option in zip(self.remotes, actions, self._seeds, self._options):
+            # seed and option are used if step triggers a reset
             remote.send(("step", (action, seed, option)))
         self.waiting = True
 


araffin:
Hello @npit thanks for the PR and sorry for the delay.

By looking at it, I think it would be better to document the fact that we don't allow to pass options if reset is not called right after.
I'm afraid that dealing with partially reset env is more tricky.

To explain what I mean:
Imagine that you have two envs and you passed `options=[{"a": 1}, {"a": 2}]` (so you want to pass `a=1` for 1st env and `a=2` for second env.

Let's say the first env reset, then you can pass `a=1`, but what shall you discard it from the options list?
Also, what should happen if afterward the user  `options=[{"a": 3}, {"a": 3}]`  before the second env is reset?
In that case, the first env would have received the correct `a=1` but the `a=2` would have been silently discarded by the user passing the new options.

Enforcing this limitation would not limit the user though.
The user can still access setter using the `VevEnv` api (see https://github.com/DLR-RM/stable-baselines3/pull/1789)

What do you think @qgallouedec ?


PS: options should not be required to work with SB3, so you should in theory make it opt-in (see what I did for the `reset()`)

npit:
> By looking at it, I think it would be better to document the fact that we don't allow to pass options if reset is not called right after.
> I'm afraid that dealing with partially reset env is more tricky.

As I see it, providing functionality of parameterizing resets but having it apply on a subset of resets is quite confusing.
What's more, the implicit resets are currently outside the user's control: the `vec_env.set_options()` utility is to be called _before_ the user loses control by invoking `vec_env.step`, `evaluate_policy` or whatever: at that point, the user may as well modify the environment directly by a bunch of `vec_env.set_attr`, since they currently have access to the vec_env. On the other hand, the implicit resets on episode terminations are out of bounds to the user once `vec_env` starts stepping.

What is the alternative if you want to control implicit resets? Should we make custom functionality instead (e.g. constructor args and dedicated logic)? Doesn't that route kind of defeat the purpose of having arguments to `reset` and the `set_options` util?

> Let's say the first env reset, then you can pass a=1, but what shall you discard it from the options list?
> Also, what should happen if afterward the user options=[{"a": 3}, {"a": 3}] before the second env is reset?
> In that case, the first env would have received the correct a=1 but the a=2 would have been silently discarded by the user passing the new options.

Currently, options persist until the user explicitly invokes `vec_env.reset` as per the docs. If #1790 is indeed an issue to be fixed (i.e. if we do want implicit resets to consider passed option/seed parameters), then it would be up to the user to build the reset logic and arguments to `set_options` to fit whatever they want to do, if their use case is complicated enough for the examples listed above to be undesirable. 

E.g. if reset options are a function of what previous options were consumed in the previous step, users can perhaps track / count the number of episodes per env and adjust. If they want different options per implicit reset, maybe pass an options list and cycle elements from it in `reset`. 
Etc.

araffin:
> the user may as well modify the environment directly by a bunch of vec_env.set_attr, since they currently have access to the vec_env. 

yes, this is the recommended way (having setters, see the example I linked which is in the doc now: https://stable-baselines3.readthedocs.io/en/master/guide/vec_envs.html#modifying-vectorized-environments-attributes).

> Doesn't that route kind of defeat the purpose of having arguments to reset and the set_options util?

this API (passing options at reset) was not a SB3 decision but a gymnasium :/ (I was actually against it) and it was designed with single env in mind. We only provided some helper for people that wanted to use it.

What I would do in any way is update the doc, add a warning and link to the current discussion.

I will wait for the opinion of other maintainers to decide if this is something to be fixed or if it is a behavior that is ok but must be documented.

stable_baselines3/common/vec_env/subproc_vec_env.py
```diff
@@ -121,8 +122,8 @@ def __init__(self, env_fns: List[Callable[[], gym.Env]], start_method: Optional[
         super().__init__(len(env_fns), observation_space, action_space)
 
     def step_async(self, actions: np.ndarray) -> None:
-        for remote, action in zip(self.remotes, actions):
-            remote.send(("step", action))
+        for remote, action, seed, option in zip(self.remotes, actions, self._seeds, self._options):
+            remote.send(("step", (action, seed, option)))
```
qgallouedec, line 126: step with option and seed?

qgallouedec: ah ok I get it. I'd add a comment though

qgallouedec:
I have the impression that passing an option to reset is a very little-used feature. I think it's important to support it though.

**However**, supporting very precise usage such as a list of options seems to me very complicated, both in terms of implementation and maintenance. And it seems like a bottomless pit:
What if the user wants to cycle through a common list of options for all envs? If we support both an option in the form of an object and an utterable of objects, what do we do if the option type expected by the environment is an iterable (do we need to fetch the signature, or do a `try;expect`?)?
In short, too complicated for nothing.

I approve of this PR as it stands, and I think we should limit the support to this. 
For more advanced uses of option in reset, I think the user will have to customize their environment.

docs/guide/vec_envs.rst
```diff
@@ -90,7 +90,9 @@ SB3 VecEnv API is actually close to Gym 0.21 API but differs to Gym 0.26+ API:
   Note that if ``render_mode != "rgb_array"``, you can only call ``vec_env.render()`` (without argument or with ``mode=env.render_mode``).
 
 - the ``reset()`` method doesn't take any parameter. If you want to seed the pseudo-random generator or pass options,
-  you should call ``vec_env.seed(seed=seed)``/``vec_env.set_options(options)`` and ``obs = vec_env.reset()`` afterward (seed and options are discarded after each call to ``reset()``).
+  you should call ``vec_env.seed(seed=seed)``/``vec_env.set_options(options)``.
+  Seed and options parameters will be passed to the next call to ``obs = vec_env.reset()`` and any implicit environment reset invoked by episode termination / truncation.
+  The provided seed and options will be discarded after each call to ``vec_env.reset()``.
```
qgallouedec, line 95: outdated right?

qgallouedec (changes requested):

qgallouedec:
If the user absolutely need option to be a cycle, they can still do this: 

```python
import gymnasium as gym

class MyWrapper(gym.Wrapper):
    def __init__(self, env, option_cylce):
        super().__init__(env)
        self.option_cycle = option_cylce

    def reset(self, seed=None, options=None):
        return self.env.reset(seed=seed, options=next(self.option_cycle))
```

stable_baselines3/common/vec_env/dummy_vec_env.py
```diff
@@ -67,7 +67,10 @@ def step_wait(self) -> VecEnvStepReturn:
             if self.buf_dones[env_idx]:
                 # save final observation where user can get it, then reset
                 self.buf_infos[env_idx]["terminal_observation"] = obs
-                obs, self.reset_infos[env_idx] = self.envs[env_idx].reset()
+                # reset the environment, supplying seed and options
+                seed = self._seeds[env_idx]
+                options = self._options[env_idx]
+                obs, self.reset_infos[env_idx] = self.envs[env_idx].reset(seed=seed, options=options)
```
araffin, line 73: as I wrote in my first comment: "PS: options should not be required to work with SB3, so you should in theory make it opt-in (see what I did for the reset())"

PR #1910, opened 2024-04-24
Fix tensorboad video slow numpy->torch conversion

NickLucche:
Hey,

this PR is a small docs fix to the speed issue highlighted here https://github.com/DLR-RM/stable-baselines3/pull/196#issuecomment-714656995, which also shows up as a warning now in more recent torch versions
`:UserWarning: Creating a tensor from a list of numpy.ndarrays is extremely slow. Please  consider converting the list to a single numpy.ndarray with numpy.array() before converting to a tensor. (Triggered internally at ../torch/csrc/utils/tensor_new.cpp:275.)`.

You can also test the speed difference yourself with:
```python3
from time import time
import numpy as np
import torch

rgb = np.ones((3,256,320))
rgbs = [rgb.copy() for _ in range(100)]

s = time()
a = torch.ByteTensor(rgbs)
print(time()-s)

s = time()
a = torch.from_numpy(np.asarray(rgbs))
print(time()-s)
``` 
I get two orders of magnitude of difference in times here.

## Description
I create a numpy array first and then let torch use the same storage. `np.stack` could also be suggested if we prefer being more explicit.

## Motivation and Context
Fixes https://github.com/DLR-RM/stable-baselines3/pull/196#issuecomment-714656995.

## Types of changes

## Checklist


docs/guide/tensorboard.rst
```diff
@@ -239,7 +240,7 @@ Here is an example of how to render an episode and log the resulting video to Te
                 )
                 self.logger.record(
                     "trajectory/video",
-                    Video(th.ByteTensor([screens]), fps=40),
+                    Video(th.from_numpy(np.asarray([screens])), fps=40),
```
araffin, line 243:
```suggestion
                    Video(th.from_numpy(np.asarray([screens], dtype=np.uint8)), fps=40),
```

Thanks for the PR =)
Could you please also update the changelog? (see PR checklist)

NickLucche: Thanks for your review! 
> Could you please also update the changelog? 

Sure, sorry about that, I wasn't sure whether I should still update the changelog for docs changes.

Regarding the explicit dtype, I let numpy maintain the one from the array list, as one may still be (perhaps) interested in writing out float values [0,1] https://pytorch.org/docs/stable/tensorboard.html#torch.utils.tensorboard.writer.SummaryWriter.add_video.
But I don't have a strong opinion here so I would just trust your judgement.

araffin: > in writing out float values [0,1]

I see, fine for me then, but maybe add a comment to explain what types are expected (or link to the doc).

PR #1947, opened 2024-06-13
Fix broken link

ikamensh:
old link stopped working. I've googled the document in question and found a working link, though I don't know if its the same version of the document as before.


araffin:
Hello,
thanks for the PR =)
let me try to contact @joschu first as all the links to pdf on his website seems to be broken: http://joschu.net/presentations.html

araffin:
should be fixed now

PR #2275, opened 2026-08-12
Warn users when `n_steps` is passed with custom replay buffer + doc update

### LINKED ISSUE #2274: [Bug]:  n_steps > 1 is silently ignored when replay_buffer_class is passed explicitly
### 🐛 Bug

When `replay_buffer_class` is passed explicitly to an off-policy algorithm, `n_steps > 1`
is silently ignored and the algorithm trains with 1-step returns.

In `OffPolicyAlgorithm._setup_model()`, both the n-step buffer selection and its validation
live inside the `if self.replay_buffer_class is None:` branch. When the user supplies the
class, neither runs: `self.n_steps` keeps its value, but the buffer is not an
`NStepReplayBuffer`, so `sample()` returns `discounts=None` and `train()` falls back to
`discounts = self.gamma`. No exception and no warning are raised.

### To Reproduce

**1. `n_steps` is accepted but never applied when the buffer class is given explicitly**

```python
from stable_baselines3 import SAC
from stable_baselines3.common.buffers import ReplayBuffer

model = SAC("MlpPolicy", "Pendulum-v1", n_steps=3, buffer_size=200,
            replay_buffer_class=ReplayBuffer)
print(model.n_steps, type(model.replay_buffer).__name__)
```

```
3 ReplayBuffer
```

`n_steps=3` is stored on the model, but the buffer built cannot compute n-step returns.
`sample()` therefore returns `discounts=None`, and `train()` falls back to
`discounts = self.gamma`, i.e. 1-step returns. Nothing is raised or warned.

**2. The same configuration is rejected or accepted depending only on whether the class is named**

```python
from stable_baselines3 import SAC
from stable_baselines3.her import HerReplayBuffer

env_kwargs = dict(n_bits=4, continuous=True)

try:
    SAC("MultiInputPolicy", BitFlippingEnv(**env_kwargs), n_steps=3, buffer_size=200)
except AssertionError as e:
    print("no explicit class ->", type(e).__name__ + ":", e)

m = SAC("MultiInputPolicy", BitFlippingEnv(**env_kwargs), n_steps=3, buffer_size=200,
        replay_buffer_class=HerReplayBuffer, learning_starts=10)
print("explicit HerReplayBuffer -> no error;", type(m.replay_buffer).__name__, "n_steps =", m.n_steps)
```

```
no explicit class -> AssertionError: N-step returns are not supported for Dict observation spaces yet.
explicit HerReplayBuffer -> no error; HerReplayBuffer n_steps = 3
```

```shell
No error and no warning are raised in the silent cases -- that is the bug.

The only exception produced anywhere in the repro is:

    AssertionError: N-step returns are not supported for Dict observation spaces yet.

It comes from the case that already behaves CORRECTLY: a Dict observation space with
n_steps=3 and no explicit replay_buffer_class. It is shown to highlight the inconsistency,
not because it is the problem being reported.
```

### System Info

_No response_

### Checklist

- [x] My issue does not relate to a custom gym environment. (Use the custom gym env template instead)
- [x] I have checked that there is no similar [issue](https://github.com/DLR-RM/stable-baselines3/issues) in the repo
- [x] I have read the [documentation](https://stable-baselines3.readthedocs.io/en/master/)
- [x] I have provided a [minimal and working](https://github.com/DLR-RM/stable-baselines3/issues/982#issuecomment-1197044014) example to reproduce the bug
- [x] I've used the [markdown code blocks](https://help.github.com/en/articles/creating-and-highlighting-code-blocks) for both code and stack traces.

Koustav-github:
## Description

Moves the n-step validation in `OffPolicyAlgorithm._setup_model()` out of the
`if self.replay_buffer_class is None:` branch, so it also applies when the user supplies the
buffer class. Previously, passing `replay_buffer_class` explicitly caused `n_steps > 1` to be
silently ignored: the model kept `n_steps`, but the buffer could not compute n-step returns, so
`sample()` returned `discounts=None` and training fell back to 1-step returns with no error or
warning.

Now, when `n_steps > 1`:
- Dict observation spaces raise (previously raised only on the default path)
- a class that is not an `NStepReplayBuffer` subclass raises, naming the class
- an `NStepReplayBuffer` subclass receives `n_steps`/`gamma`, as the default path already does

The subclass case uses `issubclass` rather than an identity check, so user subclasses keep
working — today such a subclass silently receives the buffer's own defaults (`n_steps=3`,
`gamma=0.99`) instead of the values passed to the algorithm.

## Motivation and Context

Closes #2274.

This mostly affects HER users, since `HerReplayBuffer` is the class the docs recommend passing
explicitly — `SAC(..., replay_buffer_class=HerReplayBuffer, n_steps=3)` silently trained with
1-step returns. It also made the same configuration behave inconsistently: rejected with an
`AssertionError` on the default path, accepted silently when the class was named.

- [x] I have raised an issue to propose this change

## Types of changes

Marked as breaking because the Dict-obs rejection now raises `ValueError` instead of
`AssertionError`. **This is the open question from the issue** — I chose `ValueError` so both
rejections share a type and survive `python -O`, but I'm happy to keep `assert` if you prefer
the smaller diff. Opening as a draft until you've had a chance to weigh in; no rush.

## Checklist
- [x] I have updated the changelog accordingly (`docs/misc/changelog.md`) (**required**).

### Notes on verification

- `black` could not run on my local Python 3.12.5 (it refuses with a CPython memory-safety
  warning), so I ran it via `uvx --python 3.12.4 black` — 95 files unchanged. `ruff` and `mypy`
  pass locally.
- I ran the affected suites rather than the full `make pytest`: `test_n_step_replay.py` (28
  passed), plus `test_buffers.py`, `test_her.py`, `test_run.py`, `test_callbacks.py`,
  `test_save_load.py`. A few failures in `test_save_load.py` and `test_performance_her` are
  pre-existing on Windows and reproduce without this change.
- - Ran the full `make pytest` suite (`pytest -m "not expensive"`): **877 passed, 7 failed,
  29 skipped**. All 7 failures are pre-existing on Windows and unrelated to this change:
  `test_performance_her[10]` (a learning-performance threshold — confirmed it fails
  identically with this patch reverted), two `test_monitor.py` and one `test_vec_monitor.py`
  tests failing with `PermissionError: [WinError 32]` on the monitor CSV file handle, and
  three `test_save_load.py` tests exercising `open_path()` where Windows raises
  `PermissionError` instead of the expected `IsADirectoryError`. CI on Linux should cover
  all of these properly.

Disclosure per CONTRIBUTING.md: I used an AI code assistant (Claude) while investigating this
and drafting the patch. I have reviewed and tested the change and understand it.


stable_baselines3/common/off_policy_algorithm.py
```diff
@@ -175,17 +178,41 @@ def _setup_model(self) -> None:
...
+                f"(`{self.replay_buffer_class.__name__}`) was passed: n-step returns are only set up "
+                "automatically when `replay_buffer_class` is `None`. "
+                "To use n-step returns with a custom replay buffer, configure it explicitly, e.g. "
+                f'`replay_buffer_kwargs={{"n_steps": {self.n_steps}, "gamma": {self.gamma}}}`.'
+            )
+
         if self.replay_buffer_class is None:
             if isinstance(self.observation_space, spaces.Dict):
                 self.replay_buffer_class = DictReplayBuffer
-                assert self.n_steps == 1, "N-step returns are not supported for Dict observation spaces yet."
```
araffin, line 181: this should be reverted no? (well, anything not related to the warning)
And maybe add quick test to check that the warning is correct emitted.

Koustav-github: hey @araffin,

that was a fair call, my files drifted due to some merge conflicts, i changed them in  my latest commit.

and for the quick test u can run this part of the `test_n_step_replay.py`

```
tests/test_n_step_replay.py::test_n_steps_warns_with_custom_replay_buffer_class  -> this will raise the warning
tests/test_n_step_replay.py::test_custom_replay_buffer_class_without_n_steps -> should pass with no warning
```

araffin (approved):

